Issue #8, opened 2017-06-30
As a smith I need two Xpect related files in N4JS repository


jpilgrim:
There are two files in NumberFour/n4js which were not added yet to eclipse/n4js since they contain code originally created by Moritz Eysholdt (@meysholdt), author of Xpect.

After analysis of both files, we (Jens and Moritz) came to the following conclusion:

* ScopeXpectMethod: The originally imported file was called AbstractQnAwareScopingTest. This file was renamed and changed quite often and does contain almost no code of the orignal version (name is different, code is different, some "import statements" remained). We agreed to change copyright notice and remove author tag.
* N4JSCommaSeparatedValuesExpectation is a customised version of CommaSeparatedValuesExpectationImpl. In this case, a fair amount of code written by Moritz can be identified. We therefore decided to add that file with a commit message including Moritz' sign-off (Also-by: Moritz Eysholdt \<his mail address>)

## Acceptance Criteria
- [x] Both files are added to the eclipse/n4js repository with appropriate copyright header and sign-off

yoosiba:
fixed with #9 and #13

PR #9, opened 2017-07-03
GH-8: As a smith I need two Xpect related files in N4JS repository

mmews-n4:
__Task:__ https://github.com/eclipse/n4js/issues/8

Also-by: Moritz Eysholdt <moritz.eysholdt@typefox.io>

### DIFF
--- a/plugins/org.eclipse.n4js.xpect/src/org/eclipse/n4js/xpect/methods/scoping/N4JSCommaSeparatedValuesExpectation.java
+++ b/plugins/org.eclipse.n4js.xpect/src/org/eclipse/n4js/xpect/methods/scoping/N4JSCommaSeparatedValuesExpectation.java
@@ -1,0 +1,249 @@
+/*******************************************************************************
+ * Copyright (c) 2012 itemis AG (http://www.itemis.eu) and others.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *******************************************************************************/
+package eu.numberfour.n4js.xpect.methods.scoping;
+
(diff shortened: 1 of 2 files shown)


jpilgrim:
@mor-n4 Sorry for assigning you again, that was a mouse-glitch... and I do not now how to take it back.

@meysholdt Could you add your ok here as well?

@mmews-n4 Removed Moritz sign-off -- we can only add him as "Also-by"

plugins/org.eclipse.n4js.xpect/src/org/eclipse/n4js/xpect/methods/scoping/N4JSCommaSeparatedValuesExpectation.java
```diff
@@ -0,0 +1,249 @@
...
+				String expDoc = replaceInDocument(expString.toString());
+				String actDoc = replaceInDocument(actString.toString());
+				throw new ComparisonFailure("", expDoc, actDoc);
+			}
+		}
+
+		/**
+		 * Explicitly request all expected items, as they may appear in scope only on explicit request.
+		 */
+		private void adaptActual(Iterable<?> actual, N4JSExpectationCollection exp, ActualCollection act) {
```
meysholdt: I believe this customization is technically not necessary, because it should be possible to wrap the logic from  `adaptActual()` into a `Predicate` and call `expectation.assertEquals(null, predicate)`. It's important that the first parameter is null, otherwise the predicate will only be used on Expectations containing a wildcard.

meysholdt: This way the original CommaSeparatedValuesExpectation from Xpect could be used.

meysholdt:
ok

mmews-n4:
@meysholdt Thank you for your approval! We also think that we should refactor this.

Issue #50, opened 2017-07-11
As a smith I need to update about.html files after transferring initial 3rd-party-related things


jpilgrim:
As a smith I need to update about.html files after transferring 3rd-party-related things and recent changes to the Eclipse N4JS repo.

In order to simplify the FileChecker.java tool, we move all information currently found in `third-party-txt` to comments in the about dialog. The format is as follows:
```
<!--
THIRD-PARTY_PATH: project/relative/path/to.file
THIRD-PARTY_PATH: project/relative/path/to/folder/**
-->
````
# Acceptance Criteria

- about.html of all projects with 3rd party content contain description of this content
- about.html of all projects with content copied from other Eclipse projects contain a comment line

PR #56, opened 2017-07-12
#50: As a smith I need to update about.html files after transferring initial 3rd-party-related things

jpilgrim:
- [x] about.html of all projects with 3rd party content contain description of this content
- [x] about.html of all projects with content copied from other Eclipse projects contain a comment line
- [x] updated about.html_template for new projects
- [x] fixed version number of httpmime in o.e.n.tester.tests.helper


jpilgrim:
used wrong account for some commits (@JensN4 instead of @jpilgrim). Created new branch GH-50a fixing this, see #57

PR #64, opened 2017-07-12
Fixed problems with content assist

szarnekow:
This PR implements a few localized workarounds for bugs in Xtext.

It also contains a new switch in the ParserCompressorFragment2 that allows to squash lines eagerly in the generated Antlr content assist parser to allow debugging. The output is currently very ugly, thus the switch is disabled by default.


szarnekow:
The same contents as https://github.com/eclipse/n4js/pull/65 but easier to review because individual commits. The second PR is squashed properly to have the correct sign-off information.

szarnekow:
Please discard this after the review, @mor-n4 @jpilgrim

Issue #83, opened 2017-07-17
As a smith I want synchronized Maven and local builds: Enable useProjectSettings in Tycho


mmews-n4:
Enable `useProjectSettings` to have a synchronized build on Eclipse and Maven. This is important for external committers, who might not use our Eclipse settings.

PR #88, opened 2017-07-19
#83: As a smith I want synchronized Maven and local builds: Enable useProjectSettings in Tycho

mmews-n4:
__Task #83__


releng/org.eclipse.n4js.parent/pom.xml
```diff
@@ -247,8 +247,11 @@ Contributors:
 							is not considered by Tycho/maven together with our harsh settings for optional warnings/errors;
 							therefore: do not use this at the moment!)
 							Note: by default, the setting is enabled.
-						<useProjectSettings>true</useProjectSettings>
 						-->
+						<useProjectSettings>true</useProjectSettings>
```
mor-n4: add rationale for using this (beyond just "it's more consistent") as comment here

plugins/org.eclipse.n4js.n4jsx.model/pom.xml
```diff
@@ -24,6 +24,13 @@ Contributors:
 		<relativePath>../pom.xml</relativePath>
 	</parent>
 
+	<properties>
+		<skip.mwe2>false</skip.mwe2>
+		<mwe2-file-location>/src/org/eclipse/n4js/model/GenerateN4JSModel.mwe2</mwe2-file-location>
```
mor-n4: don't refer to mwe2 workflow from other bundle (create new workflow in this bundle)

mmews-n4:
Thanks for your review!
I just checked the compiler setting in the parent-pom: Without this setting, the utils project does not compile (with maven). In Eclipse it does, since we disabled some warnings there. I prefer to keep the compiler setting at the same location as the `useProjectSettings` setting, since it could later seem strange, when some other project changes its warning/error-levels but would then cause compiler errors.

Issue #97, opened 2017-07-20
Building N4JS in Oxygen yields many changes in generated code.


yoosiba:
It would be nice to use Oxygen when working with N4JS. Some features like [Launch Groups](https://www.eclipse.org/oxygen/noteworthy/#platform-debug) would really help when working locally.
Currently building N4JS with Oxygen results in many changes to the generated code. This looks like only the EMF comments that have html encoded. 
Committing those changes would allow committers to use Oxygen as development environment.

Note, this does not assume changes to the eclipse platform plugins used to build the product. Changing version of our dependencies (i.e. base platform for the N4JS) is another issue.

PR #98, opened 2017-07-20
#97: add changes generated by the Oxygen

yoosiba:
See #97 
Following are javadoc changes generated by the Oxygen.


mor-n4:
will be done later

Issue #96, opened 2017-07-20
Move N4JSC.jar implementation to the equinox based solution.


yoosiba:
Currently the headless compiler is build as plain java - the `N4JSC.jar`.
This has several drawbacks:
 * All the Eclipse services need to be wired manually (e.g. EMF registrations, etc)
 * We have to do workaround for missing functionality of extension points
 * All dependencies are duplicated between pom.xml files MANIFEST.MF files (which leads to inconsistent set of bundles and versions).

It would be good to migrate `n4js.hlc` to the eclipse-bundle based build that would run with the OSGI / Eclipse platform runtime.

yoosiba:
It is possible to do self contained executable jar with the eclipse application, see proof of concept [fat headless eclipse](https://github.com/yoosiba/fhe)

PR #118, opened 2017-08-02
#96 transition N4JS.jar to platform based jar

yoosiba:
This moves N4JS.jar to be based on the eclipse / osgi platform.

This PR:
- [x] n4js.jar is internally starting and running equinox/eclipse platform
- [x] transformed plain maven jar packaging to eclipse-bundlue type
- [x] removed duplicated dependencies in pom of hlc (all builds rely on MANIFEST.MF)
- [x] switching n4js.jar to use extension points instead of manual wiring


yoosiba:
* added copyright headers, added information about 3rd party part;
* adjusted config to run hlc.tests which were not running before

* `tools` I am keeping for now, but we should discuss if we need this - all projects are `eclipse-bundle` now, and there is no plain maven project anymore, hence the `tools` could be just moved to the `plugins` (to be discussed).

Issue #126, opened 2017-08-08
TModule must be correctly rewired when AST is loaded after TModule


jpilgrim:
When a TModule is loaded from index, and after that the AST is resolved (via astElement),
the module loaded from the index must be correctly rewired. At the moment, a new TModule is created and the old TModule becomes inconsistent.

jpilgrim:
Oliver, on GH-126 I added a first implementation which
- computes an MD5 hash of the source (computing a hash of the AST is very difficult!)
- does not resolve the AST if it has changed -- this is ok I think: The AST has changed thus we cannot resolve the version used by the outdated type model
- in other cases rewires the loaded AST to the existing TModule and keeps that in the resource

This seems to work, unfortunately some tests fail. I have debugged these tests and they are really complicated. I already renamed some stuff, but I haven't figured out a solution. Seems as if the TModule contains some proxies, I don't really know what these proxies are if if they really are a problem (i.e. if they are only a problem for the tests). See branch build for failing tests.

Could you have a look at these tests? Maybe we need to change the tests (it is basically one base test class), or maybe you can fix the problem of the proxies.

jpilgrim:
I added a breakpoint at N4JSResource.demandLoadResource whenever an AST of a TModule loaded from index is to be resolved. This happens often in hover cases, what is ok.

But it also happens during build:
```
Thread [Worker-4] (Suspended (breakpoint at line 396 in N4JSResource))	
	owns: N4JSGenerateImmediatelyBuilderState  (id=18127)	
	N4JSResource.demandLoadResource(EObject) line: 396	
	N4JSResource.access$1(N4JSResource, EObject) line: 395	
	N4JSResource$ModuleAwareContentsList.resolve(int, EObject) line: 115	
	N4JSResource$ModuleAwareContentsList.resolve(int, Object) line: 1	
	N4JSResource$ModuleAwareContentsList(BasicEList<E>).get(int) line: 348	
	N4JSResource.getScriptResolved() line: 1027	
	N4JSResource.resolveLazyCrossReferences(CancelIndicator) line: 1002	
	EcoreUtil2.resolveLazyCrossReferences(Resource, CancelIndicator) line: 498	
	N4JSGenerateImmediatelyBuilderState(ClusteringBuilderState).doUpdate(BuildData, ResourceDescriptionsData, IProgressMonitor) line: 230	
[...]

PR #128, opened 2017-08-08
#126: rewire TModule loaded from index after resolving AST

jpilgrim:
Issue #126

When loading an AST from a TModule (usually retreived from the index) by calling SyntaxRelatedTElement.astElement,  this caused a new TModule to be created again from the AST. That lead to an inconsistent state: Two different TModules which both link to the same AST, and the original TModule not contained in any resource anymore (because it was removed from the resource and replaced with the newly loaded one). Since the original one is detached from a resource, it cannot resolve proxies anymore. This lead to a lot of weird scenarios which are extremely hard to debug because the both TModule instances look similar.

This bugfix adds a new rewire step after the AST has been loaded and a new TModule has been created. It first compares the original TModule with the new one. If both are similar (except for references or derived/transient data), all links from the AST pointing to the new TModule are rewired to point to the old TModule (and also all astElements are set to the new AST).

- [ ] build


mor-n4:
superseded by #153

Issue #133, opened 2017-08-14
As a smith I don't need the Eclipse Setup Tasks to clean-build the whole project


mmews-n4:
Currently, the N4JS.setup file forces Eclipse to clean-build the whole project which is time consuming and often not necessary. Remove that directive.

PR #134, opened 2017-08-14
#133: As a smith I don't need the Eclipse Setup Tasks to clean-build the whole project

mmews-n4:
__Task:__ #133

- [x] Builds


mmews-n4:
Thank you! Unfortunately, I couldn't find an option like you mentioned.

Issue #66, opened 2017-07-12
Cyclic references between files can confuse editor dirty state handling AND incremental builder / Xtext index


szarnekow:
In case of cyclic references between a file A opened in an N4JS editor and some other file B, the editor might show outdated information for the remote file B in case an unsaved change in the editor (for file A) has an effect on the TModule information of file B. In such a case, the dirty state handling does not update the information obtained from the index for file B until the editor contents are saved. In addition, also the incremental builder and Xtext index can become inconsistent until file A is saved repeatedly.

Unfortunately, this is not only a problem of the dirty state handling because it also affects the incremental builder.

To reproduce, create a file A.n4js:

```
import {B} from "B";

export class A {
	static fieldA = 'some string';
}

let str: string = B.fieldB;
```

and a file B.n4js:

```
import {A} from "A";

export class B {
	static fieldB = A.fieldA;
}
```

Now perform a clean build and then

- open file A.n4js in an N4JS editor (do not open file B.n4js).
- change the string literal 'some string' to a number literal 42 (without saving!!).
- now, an error is expected in the editor in the last line of file A.n4js (because the type of B.fieldB has now changed from string to number) but this error is not shown (Problem 1: inconsistent dirty state in editor -- but this may be correct, we do not build other files in dirty state).
- save editor contents (i.e. save file A.n4js).
- now, the editor shows the correct state (i.e. the error mentioned in step 3 shows up) but note that the incremental builder does not show this error, i.e. no error shown in Eclipse Problems View (Problem 2: inconsistent builder / Xtext index state). *This is the real problem!*
- add an empty line to file A.n4js and save again.
- now the error mentioned in step 4 shows up in the Eclipse Problems View as well.

szarnekow:
The problem could be reproduced.

szarnekow:
Same problem happens with Xtend. Basically caused by a loophole in languages with inferred information when resources have bi-directional references.

szarnekow:
Problem is the following:
- A.n4js is changed and saved which triggers the build.
- A.n4js will be built. Builder uses the index information from B.n4js which is outdated at that point.
- A.n4js is compiled and validated against the outdated B.n4js index state.
- A.n4js is detected to have changes signatures.
- B.n4js is detected to be affected by A's change.
- B.n4js is built and validated.
- A.n4js was already built in the same round thus its outdated validation result is used.

jpilgrim:
Simple example w/o classes:

File A:
```javascript
[...]

PR #138, opened 2017-08-15
#66: Improved logic to decide if load from source or index

szarnekow:


szarnekow:
Looks like I cannot invite someone for a review. @mor-n4 please take a look at this PR.

plugins/org.eclipse.n4js.ui/src/org/eclipse/n4js/ui/editor/DirtyStateAwareLoadFromSourceHelper.java
```diff
@@ -0,0 +1,67 @@
...
+			return true;
+		}
+		ResourceSetContext context = ResourceSetContext.get(resourceSet);
+		// check that we are not in the builder scope (which has strict boundaries
+		// wrt project dependencies
+		if (!context.isBuilder()) {
+			ImmutableSet<URI> openEditors = ImmutableSet.copyOf(dirtyStateManager.getDirtyResourceURIs());
+			/* ignore project boundaries */
+			return dependsOnAny(resourceURI, openEditors, getIndex(resourceSet), false);
+		}
```
mor-n4: here we have two invocations of `#dependsOnAny()` (including the one from the super method), which means we traverse & deserialize the dependency graph twice in the worst case.
Would it make sense to somehow combine the URIs from `openEditors` to the set of URIs checked by the super class and have only a single call to `#dependsOnAny()`?

mor-n4:
Thanks, this looks cool!
Great you found a solution that is pretty concise and understandable, considering the complexity of the issue.

plugins/org.eclipse.n4js.ui/src/org/eclipse/n4js/ui/editor/DirtyStateAwareLoadFromSourceHelper.java
szarnekow: Yes, that's something we should consider. I'll look into it.

szarnekow: We won't cross project boundaries in the first case but explicitly do this in the second case. I'd rather like to keep it as is. What do you think?

mor-n4: ok, makes sense

mor-n4:
builds are "green"
(there was only one test failure; should be fixed by cd6a27dde393f7742e7420b571448ae7e76957d9)

merging now...

Issue #132, opened 2017-08-14
Support composed members for non-contained ComposedTypeRef


qtran-n4:
In some cases, a synthetically created composed member is not contained in a resource. Try this example and use AST graph view to see the AST.

```
let a = new Array<String>();
let v = (1==1)? [] : a;

v.join(); 
```

Issue #73, opened 2017-07-13
As a developer I need "find references" to work also in corner cases


qtran-n4:
The following corner cases have to work as well:

- types referenced in polyfills

- references to functions, especially when imported elsewhere
- references to inferenced types
- types referenced in union/intersection types
- methods indirectly referenced (e.g. x.a.bar() should be found if A.bar() has been searched)
- other cases to be discovered.

Note that this is not a pure UI topic, since "find references" is needed for refactorings etc. as well.

## Acceptance Criteria

- detailed list of all cases (working or not) with xpect tests as examples.

- check different kind of elements (class, interface, function, variable, enum, member) from different locations (same file, same project, different project) with different references (type reference, variable reference, import, import with alias)

- if possible, fixes of minor issues
fix Xpect test method if possible, otherwise create follow-up task with detailed problem description

- create follow-up of more complicated cases with description about effort/hints.

jpilgrim:
Some additional remarks:

How does JDT behave? Look at the following snippet:

```
interface I { 	void foo(); }
interface J extends I { void foo(); }
class A { void foo() {} }
class B extends A { void foo() {} }
class CI implements I { public void foo() {} }
interface K { default void foo() {} }
interface L extends K { default void foo() {} }

//=> interfaces: hierarchy up and down
//=> classes: hierarchy only down (only sub classes)
class C {
	<T extends I & K> void f(CI ci, A a, B b, L l, T t, I i, J j) {
		ci.foo();		// CI.foo, I.foo, ==> "super" interfaces 
		a.foo();		// A.foo, B.foo => sub classes
		b.foo();		// B.foo, not A.foo => not super classes
		l.foo();		// L.foo, K.foo ==> "super" interface 
		t.foo();		// I.foo, K.foo ==> intersection
		i.foo();        // I.foo, J.foo ==> sub interface
		j.foo();        // J.foo, I.foo => super interface
	}
}
```

JDT's findReferences(Method m) with C is owner of m, C is class or interface, finds
* C.m itself
* I.m, I is interface, I<:C || C<:I
* B.m, C is class, C <: B   (i.e. B is a super type of C)

But this is the UI only!

For N4JS findReferences(Method m), we define that internally we find:
* C.m itself
nothing else

In order to "emulate" JDT's behaviour, we might add flags (toggles) in the UI to also consider overridden and overriding methods:
* "consider overridden": B.m, iff C<:B (i.e. we also look in supertypes, which is defined by the extend or implements clauses)
[...]

PR #141, opened 2017-08-15
#132: Support composed members for non-contained ComposedTypeRef

mor-n4:
see #132

Tests will be added by @qtran-n4 as part of #73, because noticing the problem fixed by this PR on the surface is easier with functionality from #73.


mor-n4:
@qtran-n4 builds still running, but you can already start looking through the changes.

plugins/org.eclipse.n4js.model/emf-gen/org/eclipse/n4js/n4JS/MemberAccess.java
```diff
@@ -0,0 +1,68 @@
+/**
+ * Copyright (c) 2016 NumberFour AG.
```
qtran-n4: 2017

mor-n4: this is EMF-generated code; we can't bump the copyright year for individual EMF entities

plugins/org.eclipse.n4js.model/emf-gen/org/eclipse/n4js/n4JS/IndexedAccessExpression.java
```diff
@@ -28,7 +28,7 @@
  * @model
  * @generated
  */
-public interface IndexedAccessExpression extends Expression {
+public interface IndexedAccessExpression extends Expression, MemberAccess {
```
mor-n4: excellent find!!

But: we cannot solve this for all cases, because we have a conflict of goals here. We cannot say:
* I want the scoping to return members that are contained in a resource AND
* I want to invoke getAllElements() on the scope AND
* I don't want the composed member cache to explode.

There's no way to solve all of that simultaneously.

But I'll try to improve the client code handling index expressions to avoid calling #getAllElements() + I'll update API doc.

plugins/org.eclipse.n4js/src/org/eclipse/n4js/typesystem/n4js.xsemantics
```diff
@@ -695,9 +698,10 @@ from {
 			val staticAccess = (targetTypeRef instanceof TypeTypeRef)
 			val checkVisibility = false // access modifiers checked in validation
 			val scope = memberScopingHelper.createMemberScope(targetTypeRef, expr, checkVisibility, staticAccess)
-			val member = memberScopingHelper.findUniqueMemberForName(scope, memberName, staticAccess)
-			if(member !== null) {
-				G |- member : var TypeRef memberTypeRef
+			val memberDesc = scope.getSingleElement(qualifiedNameConverter.toQualifiedName(memberName));
```
qtran-n4: The original method `findUniqueMemberForName` filters `IEObjectDescriptionWithError`. In this refactored version, shouldn't we check `! (memberDesc instanceof IEObjectDescriptionWithError)` as well?

tests/org.eclipse.n4js.lang.tests/src/org/eclipse/n4js/tests/scoping/ComposedMemberCachingTest.xtend
```diff
@@ -0,0 +1,109 @@
...
+import org.junit.Test
+
+
+/**
+ * Tests behavior of member scoping related to the caching of composed members. For further information, see
+ * API documentation of methods:
+ * {@link MemberScopingHelper#createMemberScope(TypeRef, MemberAccess, boolean, boolean)} and
+ * {@link MemberScopingHelper#createMemberScopeAllowingNonContainedMembers(TypeRef, EObject, boolean, boolean)}.
+ */
+class ComposedMemberCachingTest extends AbstractN4JSTest {
```
qtran-n4: Really nice tests!!

tests/org.eclipse.n4js.lang.tests/src/org/eclipse/n4js/tests/scoping/ComposedMemberCachingTest.xtend
```diff
@@ -0,0 +1,109 @@
...
+	@Test
+	def void testContainedMemberWhenUsingNonContainedTypeRef() {
+		val script = snippet.parse; // do not validate!
+		val resource = script.eResource as N4JSResource;
+
+		val varDecl = script.eAllContents.filter(VariableDeclaration).head;
+		val propAccess = script.eAllContents.filter(ParameterizedPropertyAccessExpression).head;
+
+		// use a receiver type reference that is *NOT* contained in a resource
+		val receiverTypeRef = TypeUtils.copy(varDecl.declaredTypeRef);
```
qtran-n4: Should we assert

```
assertNull(receiverTypeRef, receiverTypeRef.eResource) 
```

to be absolutely sure?

plugins/org.eclipse.n4js/src/org/eclipse/n4js/typesystem/n4js.xsemantics
mor-n4: done

tests/org.eclipse.n4js.lang.tests/src/org/eclipse/n4js/tests/scoping/ComposedMemberCachingTest.xtend
mor-n4: the copy is never contained, so I don't think we need this

mor-n4: well, it won't hurt and I have to make another commit anyway, so I added that assertion

done.

mor-n4:
builds are green (finally)
merging now ...

qtran-n4:
In the merged solution, we differentiate between two modi:

- The client code of the scoping requires that a composed member must be contained in a resource.
- The client code of the scoping does NOT require that a composed member must be contained in a resource. This is usually the case when the client needs to use the scoping infrastructure for say validation (for example `N4JSExpressionValidator#internalCheckCtorVisibility`).

I am thinking about the question: is the second case a misuse of scoping?

What is the purpose of scoping? Scoping is used by the linking service to wire AST/TModule elements with each other. Validation should happen after the scoping is done and thus can use the information provided by scoping during the validation. That scoping is CONSTRUCTED/CREATED during validation seems weird to me... 

If the creation of scope is not allowed during validation etc., the second mode may become obsolete...

mor-n4:
Mode 2 is used only in cases where no ordinary scoping takes place (otherwise the result from the ordinary scoping could be used, as you said). Examples: typing & validation of new expressions (for looking up the ctor), typing & validation of destructuring patterns.

In addition, mode 2 is also required if, for w/e reason, methods like `#getAllElements()` have to be used and excessive caching of composed members is to be avoided. Afaik, ordinary scoping triggered by Xtext will not use methods like `#getAllElements()` and is thus ok with mode 1.

PR #153, opened 2017-08-18
GH-126: Reconcile TModule on AST resolution

### LINKED ISSUE #126: TModule must be correctly rewired when AST is loaded after TModule

szarnekow:
fixes #126

this PR supersedes PR #128

Please review @jpilgrim , @mor-n4


plugins/org.eclipse.n4js/src/org/eclipse/n4js/resource/N4JSDerivedStateComputer.xtend
```diff
@@ -13,29 +13,50 @@ package org.eclipse.n4js.resource;
...
 	override installDerivedState(DerivedStateAwareResource resource, boolean preLinkingPhase) {
-		if (resource.contents.nullOrEmpty) {
+		val contents = resource.contents;
+		if (contents.nullOrEmpty) {
 			throw new IllegalStateException
-		} else if (! resource.contents.empty) {
+		} else if (contents.size == 1) {
 			typesBuilder.createTModuleFromSource(resource, preLinkingPhase);
+		} else if (contents.size == 2) {
+			/* TODO adjust comments
```
jpilgrim: Can you do that?

plugins/org.eclipse.n4js/src/org/eclipse/n4js/resource/N4JSResource.java
```diff
@@ -834,6 +924,10 @@ protected EObject getEObjectForURIFragmentRootSegment(String uriFragmentRootSegm
 				if (position < contents.size() && position >= 1) {
 					return contents.get(position);
 				}
+				if (position == 0 && contents.size() > 1) {
+					// we have a proxified AST, but a TModule loaded from the index
+
+				}
```
jpilgrim: Nothing is done here?

plugins/org.eclipse.n4js/src/org/eclipse/n4js/typesbuilder/N4JSTypesBuilder.xtend
```diff
@@ -79,6 +85,35 @@ public class N4JSTypesBuilder {
 	@Inject protected JavaScriptVariantHelper jsVariantHelper;
 
 
+	def public void linkTModuleToSource(DerivedStateAwareResource resource, boolean preLinkingPhase) {
+		val parseResult = resource.getParseResult();
+		if (parseResult !== null) {
+
+			val script = parseResult.rootASTElement as Script;
+			
+			// TODO compare hex values
```
jpilgrim: We really should do that ;-)

plugins/org.eclipse.n4js/src/org/eclipse/n4js/typesbuilder/N4JSClassDeclarationTypesBuilder.xtend
```diff
@@ -10,32 +10,33 @@
...
-	@Inject extension N4JSGetterTypesBuilder
-	@Inject extension N4JSSetterTypesBuilder
+	def protected boolean linkTClass(N4ClassDeclaration n4Class, TModule target, boolean preLinkingPhase, int idx) {
+		if (n4Class.name === null) { // may be null due to syntax errors
+			return false;
+		}
+
+		val TClass tclass = target.topLevelTypes.get(idx) as TClass
+		if (tclass.name != n4Class.name) {
+			throw new IllegalStateException('name mismatch: ' + tclass.name + ' vs ' + n4Class.name);
```
mor-n4: here you check the name; but in many other cases no such check is performed (e.g. methods, fpars, exported variables)

What was the general approach?
1. make `N4JSTypesBuilder#linkTypes()` resilient and recognize mismatching old/new AST in types builder (and thus *not* requiring an md5 hash)
2. use an md5 hash and then just assuming in `N4JSTypesBuilder#linkTypes()` that there isn't a mismatch (in that case the above name check is unnecessary)

szarnekow: We should always check the name for equality. That should do the trick. For lists, we could consider to also check the length before we traverse them.

plugins/org.eclipse.n4js/src/org/eclipse/n4js/resource/N4JSResource.java
```diff
@@ -415,25 +425,30 @@ private EObject demandLoadResource(EObject object) {
 			}
 			superLoad(null);
```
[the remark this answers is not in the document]
szarnekow: yes, this assumption is valid. As far as I can tell, contents will not be set to null in this call path.

plugins/org.eclipse.n4js/src/org/eclipse/n4js/resource/N4JSResource.java
```diff
@@ -535,8 +553,10 @@ protected void clearErrorsAndWarnings() {
 	 */
 	@Override
 	protected void doLoad(InputStream inputStream, Map<?, ?> options) throws IOException {
-		if (contents != null && !contents.isEmpty())
-			discardStateFromDescription();
+		if (contents != null && !contents.isEmpty()) {
+			//
```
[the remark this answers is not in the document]
szarnekow: Please remove

plugins/org.eclipse.n4js/src/org/eclipse/n4js/resource/N4JSResource.java
```diff
@@ -810,6 +897,9 @@ public EObject doResolveProxy(InternalEObject proxy, EObject objectContext) {
 	 * @return true, if the referenced object may come from the the index TModule.
 	 */
 	private boolean canResolveFromDescription(URI targetUri) {
+		if (targetUri.fragment().startsWith("/0")) {
+			return false;
+		}
 		return canLoadFromDescriptionHelper.canLoadFromDescription(targetUri.trimFragment(), getResourceSet());
```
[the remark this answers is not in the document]
szarnekow: canLoadFromDescriptionHelper accepts a resourceURI as parameter. We would need to ensure that all callers always have a fragment at hand. Preferably I'd leave this method here. Calling it canLoadFromDescription instead of canResolve.. is fine for me.

plugins/org.eclipse.n4js/src/org/eclipse/n4js/resource/N4JSResource.java
szarnekow: Yes, that slipped through. The condition is save to be removed again.

plugins/org.eclipse.n4js/src/org/eclipse/n4js/typesbuilder/AbstractFunctionDefinitionTypesBuilder.xtend
```diff
@@ -29,6 +29,14 @@ package class AbstractFunctionDefinitionTypesBuilder {
 	@Inject extension N4JSTypesBuilderHelper
 	@Inject extension N4JSFormalParameterTypesBuilder
 
+	def protected linkFormalParameters(TFunction functionType, FunctionDefinition functionDef, boolean preLinkingPhase) {
```
szarnekow: Please use explicit return types

plugins/org.eclipse.n4js/src/org/eclipse/n4js/typesbuilder/N4JSClassDeclarationTypesBuilder.xtend
```diff
@@ -127,49 +128,13 @@ public class N4JSClassDeclarationTypesBuilder {
...
-
-	def private addSetters(TClass tClass, N4ClassDefinition n4Class, boolean preLinkingPhase) {
-
-		// create also getters for all non private fields without explicit getter
-		val n4Setters = n4Class.ownedMembers.filter(N4SetterDeclaration);
-		val setters = n4Setters.map[createSetter(tClass, preLinkingPhase)].filterNull;
-		tClass.ownedMembers.addAll(setters);
-	}
-
 	def private setSuperType(TClass tclass, N4ClassDefinition classDecl, boolean preLinkingPhase) {
```
szarnekow: Please use explicit return types

plugins/org.eclipse.n4js/src/org/eclipse/n4js/typesbuilder/N4JSTypesBuilder.xtend
szarnekow: :+1:

plugins/org.eclipse.n4js/src/org/eclipse/n4js/typesbuilder/N4JSTypesBuilder.xtend
```diff
@@ -192,6 +228,84 @@ public class N4JSTypesBuilder {
...
+
+	def protected dispatch int linkType(N4ClassDeclaration n4Class, TModule target, boolean preLinkingPhase, int idx) {
+		if (n4Class.linkTClass(target, preLinkingPhase, idx)) {
+			return idx + 1;
+		}
+		return idx;
+	}
+
+	def protected dispatch int linkType(N4ClassExpression n4Class, TModule target, boolean preLinkingPhase, int idx) {
+		n4Class.createTClass(target, preLinkingPhase)
```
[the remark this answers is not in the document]
szarnekow: 👍 Yes, we should use explicit return types more often.

plugins/org.eclipse.n4js/src/org/eclipse/n4js/typesbuilder/N4JSTypesBuilder.xtend
```diff
@@ -193,6 +242,84 @@ public class N4JSTypesBuilder {
...
+
+	def protected dispatch int relinkType(N4EnumDeclaration n4Enum, TModule target, boolean preLinkingPhase, int idx) {
+		if (n4Enum.relinkTEnum(target, preLinkingPhase, idx)) {
+			return idx + 1;
+		}
+		return idx;
+	}
+
+	def protected dispatch int relinkType(ObjectLiteral objectLiteral, TModule target, boolean preLinkingPhase, int idx) {
+		objectLiteral.createObjectLiteral(target, preLinkingPhase)
```
mor-n4: @szarnekow Are you sure this works or is this WIP?

`#createObjectLiteral()` simply creates a new type and adds that to the `TModule#internalTypes`. Since we assume the TModule has been loaded from the index, `#internalTypes` will be empty, but for some of the types we create here, there will already be types in `#exposedInternalTypes` that should ideally be relinked.

Also, those new internal types will never be moved to `TModule#exposedInternalTypes`, or have I overlooked something?

(all of the above applies to handling of function expressions in line 315)

szarnekow: I'd think this works since assignability is checked on the structure of the types and not their identity. Thus the exposedInternalTypes will contain the 'same' types as the internal types which should not make a difference. Was there a bug that proved this assumption wrong?

mor-n4: Thanks for the quick reply!
True, in those two cases there will be no identity check, so the duplication should not be an issue.
But the functionality to move some to `#exposedInternalTypes` still seems to be required. Will have a closer look at this tomorrow.

mor-n4:
@jpilgrim  says: :+1:

builds are green

merging now (documentation update and maybe more tests will be provided on separate branch)

szarnekow:
Thank you for taking care of this.

mor-n4:
Thanks for providing these improvements! 😸

Issue #165, opened 2017-08-22
Broken link in Test Support docs


bsmith-n4:
The hyperlink in the Test Support documentation points to an old URL. This should be fixed and other documents should be checked for the same issue.

yoosiba:
fixed with https://github.com/eclipse/n4js/commit/0f338a67d428871813edd75fe17e03f67066b753

PR #166, opened 2017-08-22
#165 - fix dead link in docs

bsmith-n4:
Issue #165


yoosiba:
@jpilgrim I think you are the only one with power to update website...

Issue #169, opened 2017-08-23
Fix smoke test failure


mor-n4:
For a few days now, 2 smoke test cases have been failing in the nightly builds:
```
org.eclipse.n4js.smoke.tests.GeneratedSmokeTestCases4.test_1051 [SkipLastCharacters][DefaultSmokeTester]
org.eclipse.n4js.smoke.tests.GeneratedSmokeTestCases4.test_1052 [SkipLastCharacters][DefaultSmokeTester]
```
Fix it.

mor-n4:
merged

PR #170, opened 2017-08-23
#169: Fix smoke test failures

mor-n4:
see #169

builds are running...


mor-n4:
builds green, merging now...

PR #189, opened 2017-08-29
GHOLD-414: Fix refresh behavior of N4JS Git Working Sets

lbeurerkellner:
Issue: NumberFour/n4js#414

The issue always appeared if some of the workspace projects were contained in Git Repositories that were not listed in the Git Repositories View. So to reproduce, it should suffice to launch a workspace where the projects were imported without adding their repos in the Repository View. After a restart the refresh issues would start. This is fixed now.

For builds see internal PR.


mmews-n4:
@lbeurerkellner Merging now, thank you!

Issue #197, opened 2017-08-31
Cache miss exception when building nested Object Literals containing computed names


qtran-n4:
To reproduce, create a new N4JS project with a `Test.n4js` file containing the following code snippet:

```
export public const myConst = {
    [p1]: {  // this computed property causes problem!
        p2: val,
    }
};
```

Build the project, the following exception is thrown in the console log

```
264336 [Worker-0] ERROR org.eclipse.n4js.utils.UtilN4  - cache miss: no actual type in cache for AST node: org.eclipse.n4js.n4JS.impl.ObjectLiteralImpl@3a7b6623 in resource: platform:/resource/TestMenu/src/Test.n4js
java.lang.IllegalStateException: cache miss: no actual type in cache for AST node: org.eclipse.n4js.n4JS.impl.ObjectLiteralImpl@3a7b6623 in resource: platform:/resource/TestMenu/src/Test.n4js
	at org.eclipse.n4js.postprocessing.ASTMetaInfoCache.getType(ASTMetaInfoCache.java:112)
	at org.eclipse.n4js.postprocessing.TypeProcessor.getType(TypeProcessor.java:299)
	at org.eclipse.n4js.typesystem.N4JSTypeSystem.type(N4JSTypeSystem.java:58)
	at org.eclipse.n4js.validation.validators.N4JSTypeValidator.checkTypeMatchesExpectedType(N4JSTypeValidator.java:484)
	at sun.reflect.GeneratedMethodAccessor53.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.eclipse.xtext.validation.AbstractDeclarativeValidator$MethodWrapper.invoke(AbstractDeclarativeValidator.java:118)
	at org.eclipse.n4js.validation.N4JSValidator$1.invoke(N4JSValidator.java:96)
	at org.eclipse.xtext.validation.AbstractDeclarativeValidator.internalValidate(AbstractDeclarativeValidator.java:312)
	at org.eclipse.xtext.validation.AbstractInjectableValidator.validate(AbstractInjectableValidator.java:71)
	at org.eclipse.xtext.validation.CompositeEValidator.validate(CompositeEValidator.java:151)
	at org.eclipse.emf.ecore.util.Diagnostician.doValidate(Diagnostician.java:171)
[...]

PR #199, opened 2017-08-31
#197: Cache miss exception when building nested Object Literals containing computed names

qtran-n4:
## Task
https://github.com/eclipse/n4js/issues/197

```
export public const myConst = {
    [p1]: {  // this computed property causes problem!
        p2: val,
    }
};
```

`PolyProcessor_ObjectLiteral` needs to access the type model elements of object literals to calculate the type of inner object literals. However, in case of computed names, if a computed name is null the type model element is removed from `TModule` (in `ComputedNameProcessor`). As a result, in case of computed names, types of inner object literals are neither calculated and nor stored in the cache by `PolyProcessor_ObjectLiteral` . 

This PR also prevents a potential NPE.

### DIFF
--- a/plugins/org.eclipse.n4js.generator.common/src/org/eclipse/n4js/generator/common/AbstractSubGenerator.xtend
+++ b/plugins/org.eclipse.n4js.generator.common/src/org/eclipse/n4js/generator/common/AbstractSubGenerator.xtend
@@ -11,6 +11,10 @@
 package org.eclipse.n4js.generator.common
 
 import com.google.inject.Inject
+import java.nio.file.Path
+import java.nio.file.Paths
+import org.eclipse.emf.common.EMFPlugin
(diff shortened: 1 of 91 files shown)


qtran-n4:
@mor-n4 : Please review this PR. I think we need to discuss about how we want to handle `null` computed names correctly and consistently.

mor-n4:
as discussed: types builder does not create TModule elements for declarations with a name of `null`, so to avoid inconsistency with this types builder beahvior, let's first try to find a fix in `PolyProcessor_ObjectLiteral`

also consider this example:
```
let o: ~Object with {prop: ~Object with {propNested: string}} = {
	[missing]: {
		propNested: null
	}
};
```

qtran-n4:
Thanks for your review and pair programming!  Build is green, merging now!

Issue #202, opened 2017-09-02
Additive expression does not handle union{int,number} correctly


jpilgrim:
```
let ni: int|number = 1;

// wrongly inferred to string XPECT FIXME type ni+1 --> number
ni+1;
// wrongly inferred to string XPECT FIXME type of ni+1.5 --> number
ni+1.5;
```

Also the expected type has to allow boolean. E.g.,

```
// XPECT FIXME noerrors -->
true - true
```

(true+true returns 2)

jpilgrim:
Similar problems occur with multiplication and shift operations.

Changing all these cases to accept any types and returning (almost) always number. Only add is treated a bit differently. 

In the future we need to also support Symbol.toPrimitive methods and add more sophisticated analysis for evaluating values at compile time.

jpilgrim:
Fixed this issue and improved inference of "+" in general.

Also adjusted expected type of other math operators (Multiplicative, Shift, Additive expressions) and added some validation instead. Better validations will be available once we have the data flow analysis.

PR #205, opened 2017-09-04
#202: Additive expression does not handle union{int,number} correctly

jpilgrim:
#202

- additive expression with "add" now takes composed types, boolean and null into account to infer return type to number, string, or union{number,string}
- additive, multiplication and shift expressions now accept any type -- further checks need to be done by non-type system analysis
- improved tests (use "xpect type of" now instead of "xpect errors")


jpilgrim:
see #206, used wrong github account...

PR #206, opened 2017-09-04
#202: Additive expression does not handle union{int,number} correctly

jpilgrim:
#202

- additive expression with "add" now takes composed types, boolean and null into account to infer return type to number, string, or union{number,string}
- additive, multiplication and shift expressions now accept any type -- further checks need to be done by non-type system analysis
- improved tests (use "xpect type of" now instead of "xpect errors")


plugins/org.eclipse.n4js/src/org/eclipse/n4js/validation/messages.properties
```diff
@@ -624,8 +624,8 @@ EXP_CAST_FAILED=error;;;Cannot cast from {0} to {1}
 EXP_CAST_UNNECESSARY=warning;;;Unnecessary cast from {0} to {1}
 # no parameter required
 EXP_CAST_INVALID_TARGET=error;;;Can only cast to class, interface, enum, function, primitive, union or intersection types
-# 0: recognized type
-EXP_IS_NOT_A_VALID_NUMBER=error;;;{0} is not a valid number.
+# 0: recognized type, 1: expression type
+EXP_RESULT_IS_NAN=warning;;;Operand of type {0} in {1} will always result in NaN.
```
mor-n4: The actual reason for getting `NaN` is, in most cases, not the type of the operand but the actual value of the operand. Even if those "most cases" are not implemented yet, I'd rephrase the message to:
```
An operand of {0} in {1} will always result in NaN.
```

plugins/org.eclipse.n4js/src/org/eclipse/n4js/validation/validators/N4JSExpressionValidator.xtend
```diff
@@ -874,38 +876,51 @@ class N4JSExpressionValidator extends AbstractN4JSDeclarativeValidator {
...
-				return; // corrupt AST (e.g., while editing)
-			}
-
-			if (trhs.declaredType === bits.nullType || trhs.declaredType === bits.undefinedType)
-				issueNotANumberType(trhs.declaredType.name, ae.rhs);
+		val tlhs = ts.tau(lhs)
+		if (tlhs === null) {
+			return; // corrupt AST (e.g., while editing)
+		}
+		if (tlhs.declaredType === bits.undefinedType)
```
mor-n4: why not already support value `null` as well (i.e. show warning if declaredType if `null` type)?

plugins/org.eclipse.n4js/src/org/eclipse/n4js/typesystem/n4js.xsemantics
```diff
@@ -958,18 +958,25 @@ from {
...
 			} else {
-				if ( (lunknown || runknown) && (lnum || rnum || undef)  ) {
-					T = if (lunknown) l else r
+				if ( (lunknown || runknown) && (lnum || rnum)  ) {
+					// one is numeric, the other unknown --> 
+					T = typeSystemHelper.createUnionType(G, G.numberTypeRef, G.stringTypeRef);
 				} else {
-					T = G.stringTypeRef
+					val lMayNum = lnum || G.containsNumericOperand(l);
+					val rMayNum = rnum || G.containsNumericOperand(r);
```
mor-n4: `lMayNum`/`rMayNum` must be true also if `l`/`r` is a super type of `number` (i.e. `any`).

At the moment, the following code breaks:
```
let a: any = 5;
let str: string = 40 + a;  // does not show error
str.trim();  // exception at runtime
```

plugins/org.eclipse.n4js/src/org/eclipse/n4js/typesystem/n4js.xsemantics
```diff
@@ -2007,33 +2014,21 @@ from {
...
 /* @see [N4JS] 6.1.14 Additive Expression */
 rule expectedTypeInAdditiveExpression
 	G |- AdditiveExpression e |> Expression expression : TypeRef T
 from {
-	if (e.op != AdditiveOperator.ADD && jsVariantHelper.isTypeAware(e)) {
-		T = G.numberTypeRef
-	} else {
-		T = G.anyTypeRef
-	}
+	T = G.anyTypeRef
```
mor-n4: type `symbol` is disallowed, e.g. the following throws an exception at runtime:
```
let sym: symbol = Symbol.iterator;
sym+5;
```

plugins/org.eclipse.n4js/src/org/eclipse/n4js/validation/messages.properties
jpilgrim: Updated the message template, but the message will remain the same since the validation stems from a type check.

We will probably replace these checks with data-analysis related checks anyway.

plugins/org.eclipse.n4js/src/org/eclipse/n4js/validation/validators/N4JSExpressionValidator.xtend
jpilgrim: I added a waring that null is interpreted as 0. Note that

```
1/null != 0/null
```

plugins/org.eclipse.n4js/src/org/eclipse/n4js/typesystem/n4js.xsemantics
jpilgrim: Good catch. Added

```
// XPECT type of a+n --> union{number,string}
a+n
// XPECT type of a+s --> string
a+s
```

plugins/org.eclipse.n4js/src/org/eclipse/n4js/typesystem/n4js.xsemantics
jpilgrim: I added

```
let sym: symbol = Symbol.iterator;
// XPECT errors --> "Operand of type symbol cannot be converted to number." at "sym"
// XPECT type of n+sym --> union{number,string}
n+sym
// XPECT errors --> "Operand of type symbol cannot be converted to number." at "sym"
// XPECT type of sym+n --> union{number,string}
sym+n
// XPECT errors --> "Operand of type symbol cannot be converted to number." at "sym"
// XPECT type of s+sym --> string
s+sym
// XPECT errors --> "Operand of type symbol cannot be converted to number." at "sym"
// XPECT type of sym+s --> string
sym+s
```
(and similar for other operations, in these cases always expecting number).

I also improved the type inference of "+": I will never return "unknown type" but union{number,string} instead.

jpilgrim:
@mor-n4 Fixed issues, please have a quick look.

plugins/org.eclipse.n4js/src/org/eclipse/n4js/typesystem/n4js.xsemantics
mor-n4: > I also improved the type inference of "+": I will never return "unknown type" but union{number,string} instead.

Not sure about this one, because iirc we had the rule if an inference that depends on an unknown type should result to "unknown type" (i.e. don't suppress "unknown type" but pass on though hierarchy). But not sure right now. Would also be ok to keep as is.

mor-n4:
aside from my minor remark about "unknown type", all good!

:+1:

Issue #208, opened 2017-09-05
Post-processing not triggered when resolving proxies without type inference


mor-n4:
Post-processing of an N4JSResource (aka AST traversal) should be invoked when the first proxy is resolved, as stated in the design document Section 7.5.2.). However, this is not working in all cases.

To reproduce:
```
// Xtend code

class TempTest extends AbstractN4JSTest {

	@Test
	def void test() throws Exception {
		val script = '''
			let a = 'hello';
			console.log(a);
		'''.parseN4js;

		val res = script.eResource as N4JSResource;
		val identRef = script.eAllContents.filter(IdentifierRef).head;

		identRef.id; // should trigger post-processing, but doesn't
		assertTrue(res.isFullyProcessed);
	}
}
```

AC:

1. investigate.
2. fix.
3. add tests for all expected ways of triggering post-processing!

PR #216, opened 2017-09-06
#214: ASTStructureValidator must not resolve proxies

### LINKED ISSUE #214: ASTStructureValidator must not resolve proxies
The `ASTStructureValidator` runs even before the types builder, so it is not allowed to resolve EMF proxies, in particular it must not resolve Xtext's lazy-linking proxies.

However, at the moment it is doing that. For example `model.label` in the following line:
https://github.com/eclipse/n4js/blob/master/plugins/org.eclipse.n4js/src/org/eclipse/n4js/validation/ASTStructureValidator.xtend#L1218

AC:
1. change `ASTStructureValidator` to not resolve proxies.
2. consider adding fail-fast assertions to ensure proxy resolution does not take place before types builder has completed.

mor-n4:
fixes #214

AC 2. from the task description will be done as part of #208

### DIFF
--- a/plugins/org.eclipse.n4js/src/org/eclipse/n4js/validation/ASTStructureValidator.xtend
+++ b/plugins/org.eclipse.n4js/src/org/eclipse/n4js/validation/ASTStructureValidator.xtend
@@ -40,6 +40,7 @@ import org.eclipse.n4js.n4JS.IdentifierRef
 import org.eclipse.n4js.n4JS.IfStatement
 import org.eclipse.n4js.n4JS.IndexedAccessExpression
 import org.eclipse.n4js.n4JS.IterationStatement
+import org.eclipse.n4js.n4JS.LabelRef
 import org.eclipse.n4js.n4JS.LabelledStatement
 import org.eclipse.n4js.n4JS.LegacyOctalIntLiteral
 import org.eclipse.n4js.n4JS.LocalArgumentsVariable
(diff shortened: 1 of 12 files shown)


plugins/org.eclipse.n4js/src/org/eclipse/n4js/validation/ASTStructureValidator.xtend
```diff
@@ -1192,7 +1192,7 @@ class ASTStructureValidator {
 				new DiagnosticMessage(IssueCodes.getMessageForAST_INVALID_CONTINUE,
 					IssueCodes.getDefaultSeverity(IssueCodes.AST_INVALID_CONTINUE), IssueCodes.AST_INVALID_CONTINUE))
 		} else {
-			validateLabel(model, model.label, producer, validLabels)
```
qtran-n4: As you also rightly said, since `ASTStructureValidator` supposes to validates things that we do not want to do in the grammar, we should only validate that things that do not require resolving proxies here. In the future, consider implement this `validateLabel` in a "normal" `@Check` validator.

plugins/org.eclipse.n4js/src/org/eclipse/n4js/validation/ASTStructureValidator.xtend
```diff
@@ -1225,9 +1225,11 @@ class ASTStructureValidator {
...
 	}
 
-	def private void validateLabel(Statement model, LabelledStatement label, ASTStructureDiagnosticProducer producer,
-		Set<LabelledStatement> validLabels) {
-		if (label !== null && !label.eIsProxy && !validLabels.contains(label)) {
+	def private void validateLabelRef(LabelRef model, ASTStructureDiagnosticProducer producer,
+		Set<LabelledStatement> validLabels
+	) {
+		val labelAsText = model.labelAsText; // cannot use model.label, because we aren't allowed to resolve proxies in this phase!
+		if (labelAsText !== null && !validLabels.exists[it.name==labelAsText]) {
```
qtran-n4: you can ignore that `it`, e.g.`name == labelAsText`

mor-n4: added that on purpose to highlight that `labelAsText` is not a property of `it`

qtran-n4: ah OK :)

PR #220, opened 2017-09-07
IDE-2828: SWTBot tests are not picked up by Jenkins

qtran-n4:
…cked up on jenkins


qtran-n4:
@yoosiba Please have a look. For some reasons, the surefire report .xml in SWTBot test bundles are ignored by Jenkins. Adding an explicit junit report task to Jenkinsfile solves the problem but this still does not answer the why question :(

yoosiba:
👍

Issue #225, opened 2017-09-10
As a developer I want a warning if an object literal defines a field not defined in structural interface


jpilgrim:
As a developer I want a warning if an object literal defines a field not defined in structural interface which is the type of the variable the object literal is assigned to.

We already have that kind of check but only for spec-constructors.

A warning should only be created in case of object literals and if the type to which the object literal is assigned to is a structural type but not a structural variant of Object. No warning should be created for dynamic types.

h2. Acceptance Criteria

The following tests work

```
interface ~I {
	public x?: int;
}

// XPECT FIXME warnings --> "y is not defined in ~r~I." at "y"
let a: ~I = {y: 5};
a;

function f(p: ~r~I) {}

// XPECT FIXME warnings --> "y is not defined in ~r~I." at "y"
f({y: 5});


function g(p: ~r~Object) {}

// no warings if type is plain object XPECT nowarnings -->
g({y: 5});
```

Issue #224, opened 2017-09-10
Add intl. parameters to Number.toLocaleString


jpilgrim:
Number.toLocaleString is defined in the ECMAScript spec as follows:

```
Number.prototype.toLocaleString ( [ reserved1 [ , reserved2 ] ] )
```

The parameters are defined in the i18n API (ECMA-402):
https://www.ecma-international.org/ecma-402/4.0/index.html#sup-number.prototype.tolocalestring

```
Number.prototype.toLocaleString ( [ locales [ , options ] ] )
```

For details about these parameters, see https://www.ecma-international.org/ecma-402/4.0/index.html#numberformat-objects

We shoud update our signature of that method to
```
toLocaleString(locales=: string, options=: ~Object with {..})
```

(or add a helper type for the options)

yoosiba:
fixed https://github.com/eclipse/n4js/commit/61bae37c8430c7e8a82cdbb4a5794c6a4e3cb8da

PR #227, opened 2017-09-10
#224: defined NumberFormat options type, req. #225 to make sense

jpilgrim:
#224

- [x] add warnings in case of superfluous properties, cf. #225


n4js-libs/packages/n4js-runtime-es2015/src/n4js/intl/NumberFormat.n4jsd
```diff
@@ -10,8 +10,65 @@
  */
 @@Global @@ProvidedByRuntime
 
+/**
+ * Options type of NumberFormat.
+ */
+export external interface ~TEMP_NumberFormatOptions {
```
yoosiba: Please use better name, e.g. `NumberLocaleOptions`

n4js-libs/packages/n4js-runtime-es2015/src/n4js/Number.n4jsd
```diff
@@ -22,6 +22,19 @@
...
+    /**
+	 * Returns a localized string; the optional parameters are defined in ECMA-402 (Internationalization API Specification)
+	 
+	 * @parameter locales locale code, IETF BCP 47
+	 * @parameter options formatting options, also used in NumberFormat, see TEMP_NumberFormatOptions for details.
+	 * @see <a href="https://www.ecma-international.org/ecma-402/4.0/index.html#sup-number.prototype.tolocalestring">ECMA-402, number.prototype.toLocaleString</a>
+	 * @see <a href="https://www.ecma-international.org/ecma-402/4.0/index.html#numberformat-objects">ECMA-402, numberformat</a>
+	 */
+	@Override 
+	public toLocaleString(locales: string|[string]=undefined, options: ~r~TEMP_NumberFormatOptions=undefined): string
```
yoosiba: We should add custom locale options to the `Date`, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleString

Issue #257, opened 2017-10-04
Fix minor runtime library definitions


dbo:
For instance:
- fetch: `RequestInit.referrer`, `Headers` API
- node: `parse` options, optional parameters

yoosiba:
done https://github.com/eclipse/n4js/commit/c7fba8dcd188bdd7b165075f9e38a0f5a24fe757

PR #258, opened 2017-10-04
#257 - Fix minor runtime library definitions

dbo:
@mor-n4 Please have a look.

### DIFF
--- a/n4js-libs/packages/n4js-runtime-fetch/src/n4js/Headers.n4jsd
+++ b/n4js-libs/packages/n4js-runtime-fetch/src/n4js/Headers.n4jsd
@@ -12,7 +12,7 @@
 
 import {IteratorExt} from "Iterator"
 
-export external public class ~Headers implements Iterable<Array<string>> {
+export external public class Headers implements Iterable<Iterable2<string, string>> {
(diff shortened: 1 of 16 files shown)


n4js-libs/packages/n4js-runtime-fetch/src/n4js/Headers.n4jsd
```diff
@@ -12,16 +12,16 @@
 
 import {IteratorExt} from "Iterator"
 
-export external public class ~Headers implements Iterable<Array<string>> {
+export external public class Headers implements Iterable<Iterable2<string, string>> {
```
mor-n4: why change from `Array` to `Iterable2` when then using the same type of the first and later elements?

yoosiba:
regenerated shipped code to be aligned with the changes

n4js-libs/packages/n4js-runtime-fetch/src/n4js/Headers.n4jsd
dbo: Different API:
- `getAll()` returns all values of a particular key
- iterable/entries return tuples of key and single value; keys may appear multiple times

Issue #265, opened 2017-10-10
Avoid locking UI thread due to hover and hyperlink navigation (in editor)


mor-n4:
When the user is hovering (to show a tool tip) or navigating hyperlinks (i.e. hovering over a navigable element while holding down Ctrl / CMD) the default code in

* `DefaultHyperlinkDetector#detectHyperlinks(ITextViewer, IRegion, boolean)`
* `AbstractEObjectHover#getHoverRegion(ITextViewer, int)`

will invoke `XtextDocument#readOnly(IUnitOfWork<T, XtextResource>)` which will wait until a read-lock is available, meaning the UI thread will be blocked during that time.

ACs:

1. avoid locking the UI thread in these cases.
2. consider showing the tooltip / hyperlink later, once the read-lock can be acquired OR at least provide some feedback to the user about why tooltip / hyperlink is unavailable (carefully consider what's the best solution from a usability perspective).

mor-n4:
added issue in Xtext repo: https://github.com/eclipse/xtext-eclipse/issues/408

PR #266, opened 2017-10-11
#265: Avoid locking UI thread due to hover and hyperlink navigation (in editor)

mor-n4:
See #265.


plugins/org.eclipse.n4js.ui/src/org/eclipse/n4js/ui/editor/N4JSDocument.java
```diff
@@ -0,0 +1,100 @@
...
+		private final Lock readLock;
+
+		/** Creates a new instance. */
+		public N4JSDocumentLocker() {
+			super();
+			readLock = (Lock) readField(XtextDocumentLocker.class, "readLock", this);
+		}
+
+		/** See {@link N4JSDocument#tryReadOnly(IUnitOfWork, Object)}. */
+		public <T> T tryReadOnly(IUnitOfWork<T, XtextResource> work, T defaultValue) {
```
szarnekow: Can we somehow use a result value that carries the information whether the lock was obtained or not?

plugins/org.eclipse.n4js.ui/src/org/eclipse/n4js/ui/editor/N4JSDocument.java
```diff
@@ -0,0 +1,100 @@
...
+			}
+			return defaultValue;
+		}
+	}
+
+	/**
+	 * Utility method used to read two private fields from super classes required for the above adjustments.
+	 * <p>
+	 * Note: cannot use {@link ReflectExtensions#get(Object, String)}, because it would read the private field of the
+	 * receiver's class, not of the super class.
```
szarnekow: :) Field name shadowing ftw

plugins/org.eclipse.n4js.ui/src/org/eclipse/n4js/ui/editor/N4JSHover.java
```diff
@@ -0,0 +1,62 @@
...
+ */
+public class N4JSHover extends DispatchingEObjectTextHover {
+
+	/**
+	 * Method copied from super class with only a minor change: call to "readOnly" changed to "tryReadOnly".
+	 */
+	@Override
+	public IRegion getHoverRegion(final ITextViewer textViewer, final int offset) {
+		IXtextDocument xtextDocument = XtextDocumentUtil.get(textViewer);
+		if (xtextDocument == null)
```
szarnekow: maybe use `if (!(xtextDocument instanceof N4JSDocument)) { return super.getHoverRegion(..); }`

plugins/org.eclipse.n4js.ui/src/org/eclipse/n4js/ui/editor/N4JSHyperlinkDetector.java
```diff
@@ -0,0 +1,48 @@
...
+public class N4JSHyperlinkDetector extends DefaultHyperlinkDetector {
+
+	/**
+	 * Method copied from super class with only a minor change: call to "readOnly" changed to "tryReadOnly".
+	 */
+	@Override
+	public IHyperlink[] detectHyperlinks(final ITextViewer textViewer, final IRegion region,
+			final boolean canShowMultipleHyperlinks) {
+		final IHyperlinkHelper helper = getHelper();
+		return ((N4JSDocument) textViewer.getDocument()).tryReadOnly(new IUnitOfWork<IHyperlink[], XtextResource>() {
```
szarnekow: Same here - we should add a guard with instanceof

plugins/org.eclipse.n4js.ui/src/org/eclipse/n4js/ui/editor/N4JSHover.java
```diff
@@ -0,0 +1,62 @@
...
+			return null;
+		// TODO this is being called on change in the UI-thread. Not a good idea to do such expensive stuff.
+		// returning the region on a per token basis would be better.
+		try {
+			return ((N4JSDocument) xtextDocument).tryReadOnly(new IUnitOfWork<IRegion, XtextResource>() {
```
szarnekow: Should we use org.eclipse.xtext.util.concurrent.CancelableUnitOfWork<R, P> ?

plugins/org.eclipse.n4js.ui/src/org/eclipse/n4js/ui/editor/N4JSHyperlinkDetector.java
szarnekow: Should we use org.eclipse.xtext.util.concurrent.CancelableUnitOfWork<R, P> ?

plugins/org.eclipse.n4js.ui/src/org/eclipse/n4js/ui/editor/N4JSHover.java
mor-n4: done

plugins/org.eclipse.n4js.ui/src/org/eclipse/n4js/ui/editor/N4JSHyperlinkDetector.java
mor-n4: done

plugins/org.eclipse.n4js.ui/src/org/eclipse/n4js/ui/editor/N4JSDocument.java
mor-n4: I was considering that, also because it would be more in line with the API of `Lock#tryLock()` which returns a boolean.

However, it would make using this method much more unwieldy and in all our current use cases the pattern with the default value works fine. Did not want to add another tryReadOnly() method that is not used yet (in our code).

But if this is added to Xtext, there should definitely be a method as you propose (probably in addition to the more convenient method with the default value).

plugins/org.eclipse.n4js.ui/src/org/eclipse/n4js/ui/editor/N4JSHover.java
mor-n4: as discussed

PR #276, opened 2017-10-13
IDE-2878: Fixme annotation should allow 2 string parameters

qtran-n4:


qtran-n4:
@yoosiba : I know the change is a joke but according to our _rigorous_ process, please review :))

qtran-n4:
💃

Issue #284, opened 2017-10-17
NPE in new Class/Interface (etc) wizard


yoosiba:
When invoking new class (or interface, etc) wizard in IDE, there are NPEs thrown, but swallowed somewhere in the UI. 
This happens on initial invocation when preview contents is set to empty string and highlighting on the preview is invoked.

yoosiba:
fixed  https://github.com/eclipse/n4js/commit/45ae22f4b89ec36b0b5ae145990ac8f4cf148122

PR #285, opened 2017-10-17
GH-284 avoid NPE in the new class wizard

yoosiba:


eclipsewebmaster:
Issue tracker reference:
https://bugs.eclipse.org/bugs/show_bug.cgi?id=284

Issue #261, opened 2017-10-04
As a smith I want to consider *not* using the Xtext cache for storing the ASTMetaInfoCache


mor-n4:
Instead of storing the ASTMetaInfoCache in the ordinary Xtext resource scope cache, consider storing it simply in a field in `N4JSResource`.

The main difference would be that the ASTMetaInfoCache will no longer be cleared when an Xtext cache clear occurs (e.g. when an EObject in the resource is modified without turning off notifications).

Rationale:
* At first this new approach seems incorrect, because normally you would assume that when Xtext considers the resource scope cache to be invalid (and thus clears that cache), we normally would think also the ASTMetaInfoCache is now invalid and has to be cleared.
* However, Xtext does not always discard the derived state when clearing the cache. So, according to Xtext, there are cases of clearing the cache but the derived state is still valid.
* Since our ASTMetaInfoCache is very closely related to the derived state, we could argue to only invalidate/clear the ASTMetaInfoCache when `#discardDerivedState()` is invoked.

This would simplify a few places in the code. BUT it would not relieve us of turning notifications on/off, because there is still other data stored in the Xtext resource scope cache.

Acceptance Criteria:
1. carefully reconsider the above rationale.
2. if deemed possible and useful, implement the changes.
3. adjust the documentation (in design document) and tests of the resource load states.

mor-n4:
An early implementation of the above idea by @qtran-n4  can be found on branch `GH-261-quang` (but note that this implementation does not follow in all regards what is described above in the issue description).

PR #313, opened 2017-10-23
#261: As a smith I want to consider *not* using the Xtext cache for storing the ASTMetaInfoCache

mor-n4:
see #261.


plugins/org.eclipse.n4js/src/org/eclipse/n4js/resource/N4JSResource.java
```diff
@@ -259,6 +265,32 @@ public N4JSResource() {
 	}
 
 	/**
+	 * Returns the {@link ASTMetaInfoCache} (in states {@link #isFullyProcessed() "Fully Processed"} and during the
+	 * transition from "Fully Initialized" to "Fully Processed", i.e. during post-processing) or throws an exception if
+	 * the cache is unavailable (in all other states).
+	 */
+	public ASTMetaInfoCache getASTMetaInfoCache() {
+		if (astMetaInfoCache == null) {
+			if (!isFullyProcessed() && !isPostProcessing()) {
```
qtran-n4: The inner if check `if (!isFullyProcessed() && !isPostProcessing())`  could be placed at the beginning of the method?

plugins/org.eclipse.n4js/src/org/eclipse/n4js/resource/N4JSResource.java
```diff
@@ -259,6 +265,32 @@ public N4JSResource() {
...
+			} else {
+				// getter invoked in correct state, but cache is still undefined
+				throw new NullPointerException("AST meta-info cache missing");
+			}
+		}
+		return astMetaInfoCache;
+	}
+
+	/**
+	 * Set the receiving resource's {@link ASTMetaInfoCache}. Should only be called at the beginning of post-processing.
```
qtran-n4: The comment is misleading, this method is also called when clearing the cache.

plugins/org.eclipse.n4js/src/org/eclipse/n4js/resource/PostProcessingAwareResource.java
```diff
@@ -125,14 +130,14 @@ public Throwable getPostProcessingThrowable() {
 
 	@Override
 	public void discardDerivedState() {
+		discardPostProcessingResult();
```
qtran-n4: Should `discardPostProcessingResult();` be called after `super.discardDerivedState();`?

plugins/org.eclipse.n4js/src/org/eclipse/n4js/resource/PostProcessingAwareResource.java
```diff
@@ -125,14 +130,14 @@ public Throwable getPostProcessingThrowable() {
...
 		super.discardDerivedState();
-		fullyPostProcessed = false;
 	}
 
 	@Override
 	protected void doUnload() {
+		discardPostProcessingResult();
```
qtran-n4: `discardPostProcessingResult();`after `super.doUnload();`?

plugins/org.eclipse.n4js/src/org/eclipse/n4js/resource/PostProcessingAwareResource.java
mor-n4: Good point. I was thinking about the exact same thing, as well. Esp. because the old code cleared flag `fullyPostProcessed` after the super call.

However, because the creation order is:
* create derived state
* create post-processing result

I think the better deletion order is:

* delete post-processing result
* delete derived state

(though this is a (very minor) modification of the old behavior)

plugins/org.eclipse.n4js/src/org/eclipse/n4js/resource/PostProcessingAwareResource.java
mor-n4: as above

plugins/org.eclipse.n4js/src/org/eclipse/n4js/resource/N4JSResource.java
mor-n4: This check is used only to adjust the error message of the exception thrown immediately afterwards. If `astMetaInfoCache` is non-null, I think it is better to just return it.

plugins/org.eclipse.n4js/src/org/eclipse/n4js/resource/N4JSResource.java
mor-n4: done

mor-n4:
thanks for the feedback, merging ...

PR #306, opened 2017-10-19
IDE-2888: Improved detection of unnecessary editor reconcilation

szarnekow:

PR #316, opened 2017-10-24
IDE-2888: Improved detection of unnecessary editor reconcilation

mor-n4:
This PR supersedes PR #306.

### DIFF
--- a/plugins/org.eclipse.n4js.ui/src/org/eclipse/n4js/ui/N4JSClusteringBuilderConfiguration.java
+++ b/plugins/org.eclipse.n4js.ui/src/org/eclipse/n4js/ui/N4JSClusteringBuilderConfiguration.java
@@ -18,6 +18,7 @@ import org.eclipse.n4js.ui.building.BuilderStateLogger.BuilderState;
 import org.eclipse.n4js.ui.building.N4JSBuildTypeTrackingBuilder;
 import org.eclipse.n4js.ui.building.N4JSTrackedClusteringBuilderState;
 import org.eclipse.n4js.ui.containers.N4JSStorage2UriMapper;
+import org.eclipse.n4js.ui.editor.PrevStateAwareDirtyStateManager;
 import org.eclipse.n4js.ui.internal.ContributingResourceDescriptionPersister;
 import org.eclipse.xtext.builder.builderState.PersistedStateProvider;
 import org.eclipse.xtext.builder.clustering.ClusteringBuilderState;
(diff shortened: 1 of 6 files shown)


mor-n4:
@szarnekow Some comments:

* please create ticket(s) in Xtext repo for the changes in PR #316 that work around Xtext bugs (if any) and add comments in N4JS code to remove work arounds when Xtext ticket is fixed
* the comment `// todo compare user data if module` in `N4JSDirtyStateEditorSupport#equalDescriptions()` is obsolete, right? If yes, please remove.
* I understand the changes in `N4JSDirtyStateEditorSupport`, but not fully in `PrevStateAwareDirtyStateManager`.
  Are the changes in `PrevStateAwareDirtyStateManager` only required to make the previous state available to the new code in `N4JSDirtyStateEditorSupport`?
  If so, please add that explanation to the javadoc of `PrevStateAwareDirtyStateManager`.
* please add tests
* what about the incremental builder? are these changes / similar changes also required there, maybe?

mor-n4:
@szarnekow Some observations from manual testing. Mostly working like a charm, but there are some oddities, see lines marked with (FIXME).

Code changes that should trigger reconciliation in other editors:

* adding/removing an exported variable
* changing name of an exported variable
* changing initializer expression of an exported 'var' to a different value of *different* type
* changing initializer expression of an exported 'const' to a different value of *different* type
* changing initializer expression of an exported 'const' to a different value of *same* type (note the compile time value!)
* changing an exported 'var' to an exported 'const'


Code changes that should **not** trigger reconciliation in other editors:

* (FIXME) adding/removing an empty statement on top-level
* (FIXME) adding/removing a non-exported variable on top-level
* (FIXME) adding/removing a non-exported function on top-level
* adding/removing an empty statement inside a function
* adding/removing a local variable inside a function
* changing initializer expression of an exported 'var' to a different value of *same* type
* changing initializer expression of a non-exported 'var' to a different value of *different* type

To reproduce, set up a workspace as described in IDE-2888 and apply all code changes at the top of file `AccountActions.n4js` right below the import statements.

mor-n4:
Thanks @szarnekow for this great fix!

Builds green; my comments above can be addressed later.
Merging ...

szarnekow:
Hi @mor-n4,

about the three FIXME's:
These empty statements change the proxy URIs that point to the AST elements from the TModule.
We could try to override the URI fragment computation to make it more resilient against empty statements, or we could ignore the AST proxy URIs completely based on the assumption, that internal changes that yield the very same TModule are transparent to the client of the module. It would be worthwhile to try the latter and run a build / some interactive tests with that mode.

Issue #32, opened 2017-07-07
As a smith I want to have a canonical form of composed types so that the associative law applies between two composed types.


bsmith-n4:
Nesting is possible when using composed types. At the moment, these nested composed types are not transformed into a canonical form. Hence, errors occur when checking two composed types with regard to their inheritance. Using a canonical form can solve this issue. Depending on the type constraint solver, a DNF or KNF can be used, which also can be minimised. When minimising, also the inheritance relations might be taken into account to retrieve an even further minimised form. 

### Acceptance criteria
* Composed types have a canonical form, e.g. (minimised?) DNF or KNF
* Associative law applies on composed types 

(moved from internal bug tracker, original reporter: @mmews-n4)

mor-n4:
Example showing the problem:
```javascript
interface A {}
interface B {}
interface C {}

var z1 : A & (B | C);
var z2 : (A & B) | (A & C);

// works fine
// XPECT noerrors --> 
z1 = z2; 

// XPECT FIXME noerrors --> "intersection{A,union{B,C}} is not a subtype of union{intersection{A,B},intersection{A,C}}."
z2 = z1;
```

mor-n4:
Another example provided by @PhilipAB:
```javascript
let x: {function():union{union{string,intersection{string,int}},intersection{int,Iterable2<int,int>}}};
x= x;

let y: ()=>(string | (string & int)) | (int & [int, int]);
y = y;

let z: ()=>(string | string) | (int & string);
z = z;
```
Can be simplified to:
```javascript
let x: Object | (int & string);
// XPECT FIXME noerrors --> "union{Object,intersection{int,string}} is not a subtype of union{Object,intersection{int,string}}." at "x"
x = x;
```

PR #318, opened 2017-10-24
#32: adjust inference rules for subtype checks of ComposedTypeRefs

mor-n4:
See #32.


mor-n4:
hey, you promised to write test cases, not a review 😜 
The code still contains TODO's anyway.

regarding looong expressions:
As a general rule I agree, but in practice I'd say it depends on a bunch of factors: the type of language (in an expression language everything is an expression, anyway; handling of uninitialized variables), the kind of expression (simple, clear structure? free of side effects?), how many of those lines are comments (~25% in this case), and probably more.

mmews-n4:
Did you merge this somewhere?

mor-n4:
This approach seems to work in principle, but leads to performance issues. We have to come up with a better way of doing this.

Moving this to repo `n4js-archive` and closing this PR, for now.

Issue #330, opened 2017-11-02
Switch N4JS to use org.eclipse.xsemantics


yoosiba:
Since XSemantics is now eclipse project, we want to switch from `it.xsemantics` to `org.eclipse.xsemantics`

yoosiba:
done https://github.com/eclipse/n4js/commit/9429990a8d7e6b9863f45a29e4af904c75f79645

PR #333, opened 2017-11-03
GH-330 switch from it.xsemantics to org.eclipse.xsemantics

yoosiba:
#330 
This PR switches n4js from it.xsemantics to org.eclipse.xsemantics:
 - update imports and FQNs in code
 - update dependency in all the `pom.xml` and `MANIFEST.MF`
 - update P2 links in the setup and target platform files


releng/org.eclipse.n4js.targetplatform/N4JS.setup
```diff
@@ -222,10 +222,10 @@
       xsi:type="setup.p2:P2Task"
       label="XSemantics">
     <requirement
-        name="it.xsemantics.sdk.feature.group"
-        versionRange="[1.12.0,1.12.1)"/>
+        name="org.eclipse.xsemantics.sdk.feature.group"
+        versionRange="[1.12.1,1.12.2)"/>
```
qtran-n4: Was the new version Lorenzo's decision?

releng/org.eclipse.n4js.targetplatform/org.eclipse.n4js.targetplatform.target
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <?pde version="3.8"?>
-<target name="Generated target-platform N4JS:N4JS_Neon.2.GH-18" sequenceNumber="125">
+<target name="Generated target-platform N4JS:N4JS_Neon.2.GH-18" sequenceNumber="127">
```
qtran-n4: the next number of 125 is 126? But this does not matter :D

yoosiba: it is auto-generated number. Since I was generating targetplatform more than once, it happens to be incremented by more than one.

releng/org.eclipse.n4js.targetplatform/N4JS.setup
yoosiba: I guess yes. It is the only version published.
It works with our codebase (as far as our tests are concerned).

Issue #152, opened 2017-08-17
N4JS Project wizard can't handle existing project at same location


lbeurerkellner:
Steps to reproduce:  

1. Create a new N4JS project
2. Create a new N4JS Class in the project or modify the greeter file 
3. Delete the project but make sure not to select 'Delete project contents on disk (cannot be undone)' 
4. Create a project with the same name 

The second project creation will yield in multiple exceptions on the console although the project creation seems to be successful.   

What actually happens, is that the IDE fails to create resources (`manifest.n4mf`, greeter N4JS-file) that already exist at the same location.   

JDT, in comparison, handles this more gracefully by informing the user about the collision: 

![image](https://user-images.githubusercontent.com/17903049/29417199-a7e31404-8368-11e7-870c-bc75b981ba1a.png)  

All other options of the wizard are disabled, since everything will be configured based on the existing sources. 

I propose similar behaviour for the N4JS IDE.

yoosiba:
As discussed in #338  Instead of adaptation let user resolve conflict (delete old project, create project with different name, etc.)

The wizard would just show error and not proceed.

yoosiba:
fixed in #364

PR #338, opened 2017-11-06
GH-152: New N4JS project wizard detects existing projects

lbeurerkellner:
Issue: https://github.com/eclipse/n4js/issues/152


yoosiba:
Instead of adaptation let user resolve conflict (delete old project, create project with different name, etc.)

The wizard would just show error and not proceed.

Issue #349, opened 2017-11-07
Headless compiler does not support external libraries in N4JSX files


qtran-n4:
Acceptance criteria:
- Headless compiler supports external libraries in N4JSX files

mor-n4:
A quick look revealed that this is caused by `IN4JSCore` not reporting the project dependencies of the current project when asked for those dependencies by `ProjectImportEnablingScope` (an empty list is returned). The reason for this is a broken dependency injection setup due to the know "multi-language problem", see `IDE-2493`.

PR #352, opened 2017-11-08
#349: Headless compiler does not support external libraries in N4JSX files

mor-n4:
See #349.


mor-n4:
if you don't find issues and builds are green, please merge!

mor-n4:
Obviously, this is just a work around for the problem of IDE-2493. We definitely need to find a better solution.

mor-n4:
thanks! builds green, merging ...

Issue #355, opened 2017-11-14
Clean up Xpect tests in bundle org.eclipse.n4js.n4jsx.xpect.ui.tests


mor-n4:
Execution of many (or all) tests in bundle `org.eclipse.n4js.n4jsx.xpect.ui.tests` seems to be broken. If the test expectation is changed to some non-sense value, the tests remain green. Also, many tests have invalid code and/or manifests and should thus be red anyway.

To reproduce:
Change the expectation of the scoping tests in `SimpleComponentDescriptions.n4jsx.xt` to some non-sense value and run the test. Test will still be green.

1. make sure tests are executed correctly.
2. clean up all tests in this bundle to ensure (1) all code / manifests are valid and (2) all tests and test expectations make sense.

PR #356, opened 2017-11-14
#355: Clean up Xpect tests in bundle org.eclipse.n4js.n4jsx.xpect.ui.tests

mor-n4:
See #355.


mor-n4:
builds running; it's not unlikely that other tests will fail, due to the fixes in the Xpect method "scope"...

mor-n4:
builds green, merging ...

Issue #362, opened 2017-11-20
N4JSResource#loadFromDescription() should not change state of resource in case of failure


mor-n4:
In method `N4JSResource#loadFromDescription()`, the state of the resource is changed even if loading from index failed (for example, because no information was available in the index). In this case, an AST proxy is added to the contents list but no TModule is being added. This can cause issues in builds that run on a broken index (i.e. partially filled index).

Try changing the above method such that when loading from description / the index fails, the receiving resource will remain unchanged.

PR #363, opened 2017-11-20
#362: N4JSResource#loadFromDescription() should not change state of resource in case of failure

mor-n4:
See #362.


mor-n4:
I "fixed" the 3 failed test (see commit message for details).

Builds green, merging now.

Issue #359, opened 2017-11-16
Fix regression in org.eclipse.n4js.xsemantics.caching.PerformanceSubtypeTest.testSubtyping1000


mmews-n4:
**Error Message**
java.lang.NullPointerException

**Stacktrace**
```
java.lang.AssertionError: java.lang.NullPointerException
	at org.eclipse.n4js.flowgraphs.model.ControlFlowEdge.equals(ControlFlowEdge.java:96)
	at java.util.HashMap.getNode(HashMap.java:571)
	at java.util.HashMap.containsKey(HashMap.java:595)
	at java.util.HashSet.contains(HashSet.java:203)
	at org.eclipse.n4js.flowgraphs.analyses.EdgeGuideWorklist.setNext(EdgeGuideWorklist.java:129)
	at org.eclipse.n4js.flowgraphs.analyses.EdgeGuideWorklist.hasNext(EdgeGuideWorklist.java:46)
	at org.eclipse.n4js.flowgraphs.analyses.GraphVisitorGuideInternal.walkthrough(GraphVisitorGuideInternal.java:142)
	at org.eclipse.n4js.flowgraphs.analyses.GraphVisitorGuideInternal.walkthrough(GraphVisitorGuideInternal.java:100)
	at org.eclipse.n4js.flowgraphs.analyses.GraphVisitorGuideInternal.walkthroughForward(GraphVisitorGuideInternal.java:70)
	at org.eclipse.n4js.flowgraphs.analyses.GraphVisitorAnalysis.traverseForwards(GraphVisitorAnalysis.java:68)
	at org.eclipse.n4js.flowgraphs.analyses.GraphVisitorAnalysis.analyseScript(GraphVisitorAnalysis.java:56)
	at org.eclipse.n4js.flowgraphs.N4JSFlowAnalyzer.accept(N4JSFlowAnalyzer.java:196)
	at org.eclipse.n4js.validation.validators.N4JSFunctionValidator.checkFlowGraphs(N4JSFunctionValidator.java:145)
	at sun.reflect.GeneratedMethodAccessor54.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.eclipse.xtext.validation.AbstractDeclarativeValidator$MethodWrapper.invoke(AbstractDeclarativeValidator.java:118)
[...]

PR #365, opened 2017-11-22
#359: Fix regression in org.eclipse.n4js.xsemantics.caching.PerformanceSubtypeTest.testSubtyping1000

mor-n4:
See #359.


mor-n4:
thanks, will merge now ...

Issue #347, opened 2017-11-07
ClassCastException in PrevStateAwareDirtyStateManager


mor-n4:
To reproduce:
1. create a new N4JS project,
2. open the auto-created manifest,
3. add an empty `ProjectDependencies` block to the manifest and save,
4. add a dependency "foo" inside the block.

--> exception thrown even without saving and no matter if the added dependency exists or doesn't exist.

Note: exception also thrown when removing the dependency.

Trace:
```
java.lang.ClassCastException: org.eclipse.n4js.n4mf.ui.internal.N4MFDirtyStateEditorSupport cannot be cast to org.eclipse.n4js.ui.editor.N4JSDirtyStateEditorSupport
	at org.eclipse.n4js.ui.editor.PrevStateAwareDirtyStateManager.reflectiveGetInnerResource(PrevStateAwareDirtyStateManager.java:100)
	at org.eclipse.n4js.ui.editor.PrevStateAwareDirtyStateManager.extractOldDescription(PrevStateAwareDirtyStateManager.java:87)
	at org.eclipse.n4js.ui.editor.PrevStateAwareDirtyStateManager.notifyListeners(PrevStateAwareDirtyStateManager.java:32)
	at org.eclipse.xtext.ui.editor.DirtyStateManager.announceDirtyStateChanged(DirtyStateManager.java:47)
	at org.eclipse.xtext.ui.editor.DirtyStateEditorSupport.announceDirtyState(DirtyStateEditorSupport.java:567)
	at org.eclipse.xtext.ui.editor.reconciler.XtextDocumentReconcileStrategy.postParse(XtextDocumentReconcileStrategy.java:179)
	at org.eclipse.xtext.ui.editor.reconciler.XtextDocumentReconcileStrategy.doReconcile(XtextDocumentReconcileStrategy.java:153)
	at org.eclipse.xtext.ui.editor.reconciler.XtextDocumentReconcileStrategy.reconcile(XtextDocumentReconcileStrategy.java:67)
	at org.eclipse.xtext.ui.editor.reconciler.XtextReconciler.doRun(XtextReconciler.java:449)
	at org.eclipse.xtext.ui.editor.reconciler.XtextReconciler.access$3(XtextReconciler.java:429)
	at org.eclipse.xtext.ui.editor.reconciler.XtextReconciler$1.process(XtextReconciler.java:363)
	at org.eclipse.xtext.ui.editor.reconciler.XtextReconciler$1.process(XtextReconciler.java:1)
[...]

PR #381, opened 2017-11-27
GH-347: Safer access to DirtyStateEditorSupport of dirty resources

lbeurerkellner:
Issue: https://github.com/eclipse/n4js/issues/347

The existing implementation didn't allow for other subclasses of `DirtyStateEditorSupport`.
However, in case of manifest files the actual implementation was an instance of N4MFDirtyStateEditorSupport.

Since the method in use `getDirtyResource()` is part of the common superclass, this should fix the issue.


lbeurerkellner:
@yoosiba maybe you can have a look?

PR #383, opened 2017-11-27
GH-373: Fix random Concurrent Modification exception

mor-n4:
Note: since the exception is nearly impossible to reproduce, we should only merge this if the changes also make sense on their own, without considering the ConcurrentModificationException.

### DIFF
--- a/plugins/org.eclipse.n4js.ui/src/org/eclipse/n4js/ui/external/ExternalLibrariesReloadHelper.java
+++ b/plugins/org.eclipse.n4js.ui/src/org/eclipse/n4js/ui/external/ExternalLibrariesReloadHelper.java
@@ -19,12 +19,12 @@ import java.util.Collection;
 
 import org.apache.log4j.Logger;
 import org.eclipse.core.resources.IProject;
-import org.eclipse.core.resources.IWorkspaceRoot;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.core.runtime.IStatus;
 import org.eclipse.core.runtime.OperationCanceledException;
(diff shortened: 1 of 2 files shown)


qtran-n4:
Reconsider activating the logging disabled by @kduske-n4 when he fixed ConcurrentModificationException.

plugins/org.eclipse.n4js.ui/src/org/eclipse/n4js/ui/external/ExternalLibrariesReloadHelper.java
```diff
@@ -69,17 +69,24 @@
 	public void reloadLibraries(final boolean refreshNpmDefinitions, final IProgressMonitor monitor)
```
qtran-n4: According to this logics, we have to do this for ALL places in which we trigger the build.
The general question is: can multiple threads execute the build code concurrently? Probably yes at the moment.

Isn't it better to make the build code its self thread-safe with e.g. `synchronized`?

qtran-n4: Consider using a workspace root scheduling in `ExternalLibraryBuilderHelper`.

mor-n4: re 1st comment: as discussed (would not be in line with how Eclipse invokes the builder usually)

re 2nd comment: good idea! Will check that.

mor-n4:
@yoosiba @qtran-n4
as discussed: adding a similar lock in `ExternalLibraryBuilderHelper` is a good idea, but it requires some refactoring of that and related classes which is better done in a separate task. Created #389 for that.

Builds green, merging ...

Issue #386, opened 2017-11-28
Introduce dockerized builds


torstenkraemer:
Modify the build system in order to enable builds based on provisioned Docker image. 
The goal is to self-provision the build environment but not to require the image for building.

yoosiba:
superseded by #408

PR #387, opened 2017-11-28
GH-386: Introduce dockerized builds

torstenkraemer:
Please do not merge yet !
#386


torstenkraemer:
As @yoosiba commented:

> siber [10:49] 
> 03:31:12      [echo] Running list-jars-in-lib in /mnt/mesos/sandbox/workspace/n4js-Eclipse_PR-387-SKVAFUY2BHINGGRMFGG6BBPQ4J4WLSOHQO6SDMEIAJLO2ZGEK52Q/tools/org.eclipse.n4js.hlc/target/lib
> 
> this one lists all n4js bundles twice, e.g.
> org.eclipse.n4js.hlc.base_0.0.1.201712060013.jar
> org.eclipse.n4js.hlc.base_0.0.1.201712060209.jar
> org.eclipse.n4js.model_0.0.1.201712060013.jar
> org.eclipse.n4js.model_0.0.1.201712060209.jar
> org.eclipse.n4js.n4mf.model_0.0.1.201712060013.jar
> org.eclipse.n4js.n4mf.model_0.0.1.201712060209.jar
> 
> It looks like it has two jars for each N4JS bundle.
> 
> That would be consistent with error message in the part that fails:
> 03:31:16 [INFO] --- exec-maven-plugin:1.6.0:exec (it-of-n4js-compiler-command) @ org.eclipse.n4js.hlc ---
> 03:31:24   Another singleton bundle selected: osgi.identity; osgi.identity="org.eclipse.n4js.hlc.base"; type="osgi.bundle"; version:Version="0.0.1.201712060209"; singleton:="true"
> 
> if for whatever reason N4JS bundles are added twice to the N4JSC.jar,
> then OSGI will complain when starting them
> 
> 
> I cannot access the workspace of this job run, but my guess is
> that used folder, the /mnt/mesos/sandbox/workspace/n4js-Eclipse_PR-387-SKVAFUY2BHINGGRMFGG6BBPQ4J4WLSOHQO6SDMEIAJLO2ZGEK52Q/tools/org.eclipse.n4js.hlc/target/lib
> contains artifacts from some previous run.
> That is strange, as you do call maven with clean.
> 
> You can either look into org.eclipse.n4js.hlc pom and adjust ant task
> to clean folder before working with it.
> 
> But I would recommend to look into docker config to check why does
> it carries jars between the builds. This can affect other things too.

yoosiba:
this is replaced by #408

Issue #389, opened 2017-11-28
As a smith I want to refactor building of external libraries to adjust locking


mor-n4:
The main goal is to avoid having another Xtext build running in the workspace while external libraries are built (with the Xtext builder). To achieve this, use an `ISchedulingRule` of appropriate scope.

To avoid code duplication, probably some minor refactoring of parts of the library manager is required.

ACs:
1. apply refactoring (see all clients of method `ExternalLibraryBuilderHelper.build(IBuildConfiguration[], IProgressMonitor)`)
2. add calls to `IJobManager.begin/endRule()` where needed.
3. discuss with @kduske-n4 if the logging he temporarily disabled can now be re-enabled again.

mor-n4:
also: discuss with @kduske-n4 if his temporary changes can now be reverted

mor-n4:
For reference:
1. the refactoring proposed in the task description turned out to be too much effort.
2. Instead, the locking was added in multiple places (which is not as bad as it may sound, because `IJobManager.begin/endRule()` is actually supporting such "redundant" locking).
3. logging was reenabled.

PR #391, opened 2017-11-28
GH-373: Fix random Concurrent Modification exception (part 2)

mor-n4:
After the ConcurrentModificationException was (hopefully) fixed in a previous PR, this now re-enables some logging that was commented out by @kduske-n4 because it often produced this exception.


mor-n4:
On second thought, this should not be done before #389 is completed and merged to master. I will re-enable the logging as part of that other issue.

Closing this as INVALID.

PR #393, opened 2017-11-29
GH-389: As a smith I want to refactor building of external libraries to adjust locking

mor-n4:
See #389.


mor-n4:
Oddly, I can't find the build with the failure from last Wednesday anymore... maybe I looked at the build of an incorrect branch.

Anyway, built again and everything is green-ish, so merging now...

Issue #374, opened 2017-11-23
Fix random OutOfMemory when building internal test project.


jpilgrim:
At enfore we have additional, non-public tests using non-public test projects building external libraries. These tests are randomly failing at the moment with OutOfMemory exceptions. These failing tests indicate general problems in N4JS and need to be fixed. If possible, internal tests are to be made public and title of this bug may be adjusted accordingly.

The internal details are described in internal project task IDE-2908.

PR #396, opened 2017-11-30
Fix random OutOfMemory when building internal test project

qtran-n4:
Task: https://github.com/eclipse/n4js/issues/374


qtran-n4:
Thanks. Builds are green so merging :)

Issue #397, opened 2017-11-30
duplicated libraries in library manage cause unresolved references in projects


yoosiba:
Duplicated libraries in library manage cause unresolved references in projects.
For example, assume project with dependency to `n4js.lang`, i.e. project using `N4Injector`
Next add to that project some npm dependency that depends on `n4js.lang`.
Installing main dependency with library manager causes npm to install `n4js.lang`.
Now library manager sees two copies of `n4js.lang`:
 1. first one that comes with ide
 2. second one installed via npm

This will cause unresolved references to the `N4Injector`

yoosiba:
workaround users can apply now:

- Close IDE:
1. remove workspce/.metadata/.plugins/org.eclipse.n4js.external.libraries/.n4npm/package-lock.json
2. remove workspce/.metadata/.plugins/org.eclipse.n4js.external.libraries/.n4npm/node_modules/n4js.lang

- Start IDE and:
3. reload npm libraries from the disk
4. make full clean build

PR #407, opened 2017-12-06
GH-397: duplicated libraries in library manage cause unresolved references in projects

mmews-n4:
see #397 

Fixed in this PR:
- NPM libraries shadow shipped libraries
- library manager triggers clean builds now instead of full builds
- clean build triggers Manifest update
- renamed quick fix from "Get dependency" to "Install npm package into workspace"
- fixed NPE when opening node for libraries in the project explorer that do not contain some default subnodes
- adds support for performing multiple quickfixes for "Install npm package into workspace" at once
- fix for missing entries in 3rd-party libraries (npm packages)


plugins/org.eclipse.n4js.ui/src/org/eclipse/n4js/ui/quickfix/N4JSQuickfixProvider.xtend
```diff
@@ -644,15 +650,36 @@ class N4JSQuickfixProvider extends AbstractN4JSQuickfixProvider {
...
 
-		accept(issue, 'Get dependency', 'Download missing dependency from npm.', null, [ element , context |
+		val modification = new N4Modification() {
+			var boolean multipleInvocations;
 
-				val doc = context.xtextDocument;
-				val documentText = doc.get(issue.offset, issue.length);
+			override Collection<? extends IChange> computeChanges(IModificationContext context, IMarker marker, int offset, int length, EObject element) throws Exception {
+				multipleInvocations = false;
+				invokeNpmManager(element, true);
```
yoosiba: maybe `invokeNpmManager(element, ! multipleInvocations);` ?

plugins/org.eclipse.n4js.ui/src/org/eclipse/n4js/ui/navigator/internal/BuiltInProjectNode.java
```diff
@@ -65,7 +65,14 @@ public String getText() {
 				.transform(p -> p.getLocation())
 				.transform(uri -> uri.toFileString())
 				.transform(uri -> new File(uri))
-				.transform(file -> new ResourceNode(this, file))
+				.transform(file -> {
+					try {
+						return new ResourceNode(this, file);
```
yoosiba: make constructor private, expose factory method, move checks from the constructor to the factory method, instead of throwing exception return null

then filter null here, remove try-catch

mmews-n4:
Thanks!

Issue #358, opened 2017-11-15
As a smith I want to clean up / refactor code related to JSX / React


mor-n4:
Some of the code related to the handling of JSX / React needs some refactoring or at least clean up. For example, in many places "React", "react", and "index" are hard-coded and string comparisons are used more than necessary (ideally, `ReactHelper` should be the only class that uses those strings).

A good entry point is class `JSXBackendHelper` and its use inside the transpiler.

AC:
1. make sure no react-related string literals and string comparisons are used by `JSXBackendHelper` and the transpiler (as far as possible),
2. make sure the implicit import created by the transpiler for JSX elements does not introduce name clashes with existing language constructs, i.e. the following code should not produce duplicate variables in the output code and should not crash at runtime:
```javascript
let React = 0;
console.log(<div></div>);
React;
```
3. make sure the use cases documented IDE-2858 are all supported after the refactoring.
4. fix caching of backends in `JSXBackendHelper` (see comment below).

yoosiba:
@mor-n4 so, can we close this?

mor-n4:
No. The merged PR links here only because this is the follow-up.

qtran-n4:
A small fix was done in https://github.com/eclipse/n4js/issues/426. We still need to improve the JSXBackend class.

mor-n4:
The caching of backends in `JSXBackendHelper` seems incorrect because the cache is never cleared (and thus cache clear is effectively the responsibility of whoever injects the `JSXBackendHelper`, failing for example if it is injected into classes that are singletons, because in that case the cache is never cleared).

Probably caching can be removed altogether because now `ReactHelper` is caching the react module.

PR #430, opened 2017-12-22
GH-358: As a smith I want to clean up / refactor code related to JSX / React

qtran-n4:
Task: https://github.com/eclipse/n4js/issues/358


qtran-n4:
Builds are green (n4js-extended yellow because of randomly failing tests).

plugins/org.eclipse.n4js/src/org/eclipse/n4js/validation/validators/N4JSXValidator.xtend
```diff
@@ -355,30 +357,46 @@ class N4JSXValidator extends AbstractN4JSDeclarativeValidator {
...
+		if (formalParam.name == ReactHelper.REACT_NAMESPACE) {
+			val message = IssueCodes.getMessageForJSX_FORMAL_PARAM_NOT_REACT();
+					addIssue(
+						message,
+						formalParam,
+						findNameFeature(formalParam).value,
+						IssueCodes.JSX_FORMAL_PARAM_NOT_REACT
+					);
+		}
+	}
```
mor-n4: These validations do not cover all special cases. For example, the following snippet is still accepted but fails at runtime:
```
class React {}
console.log(<div></div>);
```

Maybe it would be possible to create a single validation for a common super type of VariableDeclaration, FormalParameter, N4ClassDeclaration, etc. (maybe `NamedElement` or `IdentifiableElement`, I'm not sure) and check that for `name!="React"`.

mor-n4:
ship it! :+1:

qtran-n4:
Thanks!

Issue #433, opened 2018-01-02
Exception in JSXBackendHelper in case project "react" exists in workspace


mor-n4:
If `react` is installed and managed inside the library manager (as an external library) everything works fine; but if in addition to that `react` is imported into the Eclipse workspace, then an exception is thrown in class `JSXBackendHelper`.

To reproduce:
1. create a simple N4JS project with a simple n4jsx file using a JSX tag.
2. create an N4JS project with the below manifest and an `index.n4jsd` in source folder `src`.
3. perform a clean build -> exception should be thrown.

Manifest to use:
```
ProjectId: react
ProjectType: library
ProjectVersion: 16.2.0
VendorId: npm
Output: "."
MainModule: "index"
ModuleLoader: commonjs
Sources {
    source {
        "src"
    }
}
```

PR #434, opened 2018-01-03
GH-433: Exception in JSXBackendHelper in case project "react" exists in workspace

mor-n4:
See #433.

### DIFF
--- a/plugins/org.eclipse.n4js.transpiler/src/org/eclipse/n4js/n4jsx/transpiler/utils/JSXBackendHelper.java
+++ b/plugins/org.eclipse.n4js.transpiler/src/org/eclipse/n4js/n4jsx/transpiler/utils/JSXBackendHelper.java
@@ -10,34 +10,21 @@
  */
 package org.eclipse.n4js.n4jsx.transpiler.utils;
 
-import java.io.File;
-import java.util.Collections;
 import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
 import java.util.Map;
 import java.util.NoSuchElementException;
 import java.util.Objects;
-import java.util.Set;
-import java.util.function.Function;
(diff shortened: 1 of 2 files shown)


mor-n4:
builds are green

plugins/org.eclipse.n4js.transpiler/src/org/eclipse/n4js/n4jsx/transpiler/utils/JSXBackendHelper.java
```diff
@@ -168,74 +139,33 @@ public String jsxBackendModuleQualifiedName(Resource resource) {
...
-	 */
-	private static URI stubMerger(URI first, URI second) {
-		if (first.lastSegment().compareToIgnoreCase(second.lastSegment()) > 0)
-			return first;
-		return second;
+	private void addReactToBackendsCache(Resource resource) {
+		TModule reactModule = reactHelper.lookUpReactTModule(resource);
+		if (reactModule != null) {
+			URI reactModuleURI = reactModule.eResource().getURI();
+			jsxBackends.put(ReactHelper.REACT_PROJECT_ID, reactModuleURI);
```
yoosiba: If we delegate to react helper, which is caching lookup result, do we need to cache here?

mor-n4: The caching in JSXBackendHelper is problematic, anyway, and probably has to be removed; but this has to be investigated further as part of #358.

But I'll add a comment there to make sure it won't be forgotten!

mor-n4:
thanks, merging ...

mor-n4:
see https://github.com/eclipse/n4js/issues/358#issuecomment-354984772

PR #436, opened 2018-01-03
GH-435: Wrong directory structure after NPM package adaptation

### LINKED ISSUE #435: Wrong directory structure after NPM package adaptation
Some NPMs have a directory structure such as

```
/package.json
/dist/this.js
/dist/that.js
...
```

Note that the .js files are inside a directory called `dist` here. During the adaptation process, the N4JSD files should be put alongside the .js files into the dist directory, which is currently accomplished correctly if the following two conditions are met:

- The main module is set in the NPM package.json (see `NpmPackageToProjectAdapter#computeMainModule).
- The directory structure of the N4JSD project is

```
/<version>/.project
/<version>/manifest.n4mf
/<version>/manifest.fragment
/src/this.n4jsd
/src/that.n4jsd
```

Note that the `src` folder does not reflect the structure of the NPM package, that is, there is no `dist` folder inside it. But to facilitate the shadowing necessary to edit the N4JSD project inside the workspace, and for reasons of coherence, the directory structure inside the `src` folder should reflect the directory structure inside the NPM package, i.e., it should be

```
/<version>/.project
/<version>/manifest.n4mf
/<version>/manifest.fragment
/src/dist/this.n4jsd
/src/dist/that.n4jsd
```

But in that case, the n4jsd files are put into the wrong place after adaptation has taken place. This is because the adaptation process takes the main module directive into account when computing the target path for the n4jsd files.

To allow the correct directory structure in the N4JSD project, the adaptation process should not evaluate the main module directive and just merge the contents of the `src` folder into the NPM package.

kduske-n4:
Closes #435. Additionally, these related PRs change the N4JSD project structures so that they mirror the structure of the corresponding NPM packages:

*Related PRs*
- N4JSD: https://github.com/NumberFour/n4jsd/pull/10
- N4JSD Sandbox: https://github.com/NumberFour/n4jsd-sandbox/pull/4
  
**Attention**
Note the TODOs in `TypeDefinitionGitLocationProvider.java`: Before merging, reset the type definition branches to master!

### DIFF
--- a/plugins/org.eclipse.n4js/src/org/eclipse/n4js/external/NpmPackageToProjectAdapter.java
+++ b/plugins/org.eclipse.n4js/src/org/eclipse/n4js/external/NpmPackageToProjectAdapter.java
@@ -20,7 +20,6 @@ import java.io.FileFilter;
 import java.io.FileWriter;
 import java.io.IOException;
 import java.nio.file.Path;
-import java.nio.file.Paths;
 import java.util.Collection;
 import java.util.HashSet;
 import java.util.List;
@@ -33,6 +32,7 @@ import org.eclipse.emf.common.util.URI;
 import org.eclipse.emf.ecore.EObject;
 import org.eclipse.emf.ecore.resource.Resource;
 import org.eclipse.emf.ecore.resource.ResourceSet;
+import org.eclipse.n4js.N4JSGlobals;
 import org.eclipse.n4js.binaries.BinaryCommandFactory;
 import org.eclipse.n4js.external.libraries.PackageJson;
 import org.eclipse.n4js.external.libraries.TargetPlatformFactory;
@@ -93,22 +93,11 @@ public class NpmPackageToProjectAdapter {
 
 	/** Default filter for copying N4JSD project contents during adaptation */
 	private final static Predicate<Path> COPY_N4JSD_PREDICATE = new Predicate<Path>() {
-		private final Path[] EXCLUDE = { Paths.get(".project"), Paths.get("manifest.n4mf"),
-				Paths.get("manifest.fragment") };
+		private final static String SUFFIX = "." + N4JSGlobals.N4JSD_FILE_EXTENSION;
 
 		@Override
 		public boolean apply(Path path) {
-			Path filename = path.getFileName();
-			if (filename == null)
-				return false;
-
-			for (Path exclude : EXCLUDE) {
-				if (filename.equals(exclude)) {
-					return false;
-				}
-			}
-
-			return true;
+			return path.endsWith(SUFFIX);
 		}
 	};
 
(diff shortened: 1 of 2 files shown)


plugins/org.eclipse.n4js/src/org/eclipse/n4js/external/NpmPackageToProjectAdapter.java
```diff
@@ -90,8 +91,26 @@ public boolean accept(File pathname) {
...
+				Paths.get("manifest.fragment") };
+
+		@Override
+		public boolean apply(Path path) {
+			Path filename = path.getFileName();
+			if (filename == null)
+				return false;
+
+			for (Path exclude : EXCLUDE) {
+				if (filename.equals(exclude)) {
```
yoosiba: should we have filter, to allow only `n4jsd` files only? Otherwise someone can inject code in `.js` file, or sneak in other content with side effects (scripts, symlinks, binaries ?)

plugins/org.eclipse.n4js/src/org/eclipse/n4js/external/NpmPackageToProjectAdapter.java
```diff
@@ -311,26 +314,11 @@ IStatus addTypeDefinitions(File packageRoot, PackageJson packageJson, String mai
...
-			 *
-			 * The .n4jsd files are assumed to be in a folder called "src" within the N4JSD project root folder.
-			 *
-			 * Their target path is computed from the package root path and the main module path. The given main module
-			 * path contains the file name (without extension) of the main module of the NPM, potentially prepended by a
-			 * path that represents folders. An example would be "lib/index". The target path for the N4JSD files is
-			 * then assumed to be in a folder called "lib" within the package root path.
-			 */
-			Path sourcePath = packageVersionedN4JSDSrcFolder.toPath();
+			Path sourcePath = packageVersionedN4JSDProjectRoot.toPath();
```
yoosiba: how about different approach discussed in IDE-2429:
 - the N4JSD repo contains API projects with definitions
 - installed npm packages could have just `manifest.mf` generated with IMPL project type.
 - API-IMPL mechanisms in IDE take care of shadowing / runtime

this way we could avoid polluting npm with definition files (still need manifest though.)
Also this makes it easier to maintain n4jsds, i.e. runtime always works against IMPL, while IDE links against API. We avoid hammering npm package with n4jsd files.

plugins/org.eclipse.n4js/src/org/eclipse/n4js/external/NpmPackageToProjectAdapter.java
kduske-n4: Already changed.

plugins/org.eclipse.n4js/src/org/eclipse/n4js/external/NpmPackageToProjectAdapter.java
kduske-n4: Yes, I agree in principle that it's nicer. As discussed, I do remember that it didn't work when I tested it for IDE-2429, but I don't remember the reason. I don't think we should introduce this change as part of the PR in any case.

kduske-n4:
I'll rerun the tests and merge. Thanks!

Issue #442, opened 2018-01-05
As a developer I want the layout of generated Requirement ADOC files to be adjusted


mmews-n4:
Generated adoc code of requirements should look like below:

```
. *++receiver type++* (link:++stdlib_api:packages:eu.numberfour.stdlib.model.base.api-tests:test/n4js/n4/util/PSCollectionSelectionExecTest:PSWildcardSelectionTest#_02_receiver_type___1_Iterable++[++Iterable++], link:++stdlib_api:packages:eu.numberfour.stdlib.model.base.api-tests:test/n4js/n4/util/PSCollectionSelectionExecTest:PSWildcardSelectionTest#_02_receiver_type___2_not_a_string++[++not a string++]) 
+
======
$R \subtype{Iterable} \wedge R \neq \type{string} \wedge R \neq \type{String}$
======

. *++result type++* link:++stdlib_api:packages:eu.numberfour.stdlib.model.base.api-tests:test/n4js/n4/util/PSCollectionSelectionExecTest:PSWildcardSelectionTest#_02_receiver_type___1_Iterable++[++1 Iterable++], link:++stdlib_api:packages:eu.numberfour.stdlib.model.base.api-tests:test/n4js/n4/util/PSCollectionSelectionExecTest:PSWildcardSelectionTest#_02_receiver_type___2_not_a_string++[++2 not a string++] 
+
====
The result is a subtype of IDataList:
$\tee result: T, T \subtype \type{IDataList}$
====
```

PR #446, opened 2018-01-08
GH-442: As a developer I want the layout of generated Requirement ADOC files to be adjusted

mmews-n4:
see: #442 

follow-up of PR: #443


mmews-n4:
🤣  Thanks!

PR #451, opened 2018-01-11
GH-448: Subtype check fails for React components defined in an external library when using the headless compiler

### LINKED ISSUE #448: Subtype check fails for React components defined in an external library when using the headless compiler
Assum an N4JS project with the following project dependencies in its manifest:

```
...

ProjectDependencies {
	nuka-carousel,
	react
}
```

Further assume an N4JSX source file with the following contents:

```
import NukaCarousel from "nuka-carousel";

<NukaCarousel />
```

Assume that the external libraries are set up correctly, i.e., the npms are installed and the package adaptation process was successful.

Then, the N4JSX file is validated without problems in the IDE. But when using the headless compiler, validation fails with the following error:

```
@issue: ERROR:The referred class is not a subtype of React.Component (file:/private/var/folders/ty/4_s26gmx69ndqb4jcyqdnpth0000gp/T/probands_external_project_dependencies_4525916572405877045/wsp/P1/src/test.n4jsx line : 4 column : 2)
```

kduske-n4:
Closes #448.


kduske-n4:
Builds are green!

tests/org.eclipse.n4js.hlc.tests/probands/external_project_dependencies/P2/manifest.n4mf
```diff
@@ -0,0 +1,26 @@
...
+Output: "src-gen"
+Sources {
+	source {
+		"src"
+	}
+}
+
+ProjectDependencies {
+	react
+}
```
yoosiba: why no new line?

kduske-n4: fixed, the other one too

Issue #454, opened 2018-01-12
Layout problem in generated adoc files


mmews-n4:
Nested blocks in AsciiDoc cause problems when the generated adoc sources are included inside a block. That's why the syntax for the nested blocks in the generated adoc source needs to be replaced to be nestable inside common AsciiDoc blocks (`--`).

PR #462, opened 2018-01-16
GH-454b: Layout problem in generated adoc files (Refactoring)

mmews-n4:
see #454


mmews-n4:
@qtran-n4 Please have a second look here. I just renamed the class of the generated adoc element since the original name was misleading. Thank you!

mmews-n4:
Thanks!

Issue #466, opened 2018-01-18
Exported asciidoc output lacks an empty line in descriptions of class-jsdoc


mmews-n4:
As a result, nothing is included in the final adoc document.

PR #468, opened 2018-01-18
GH-466: Exported asciidoc output lacks an empty line in descriptions of class-jsdoc

mmews-n4:
see #466


mmews-n4:
Thanks again!

Issue #464, opened 2018-01-17
As a developer I want to have guards in dataflow analysis


mmews-n4:
Make flow analysis aware of conditions of branches (e.g. if statements).

Again, this has to be "demonstrated" with the null-analysis.

The guards do not have to be complete (gaps need to be documented though), but they must be correct.

**Acceptance Criteria**

```javascript
foo() {
  val a;
  // XPECT warnings
  a.tString();
  // ...
  if (a) { // and other variations of null-check
   // XPECT no warnings
    a.toString();
  }
```

PR #423, opened 2017-12-15
GH-331: As a smith I need a Data-Flow analyses engine (with warnings if a ref is used before initialization)

mmews-n4:
see #311

PR #469, opened 2018-01-18
GH-464: As a developer I want to have guards in dataflow analysis

mmews-n4:
see #464 

This PR consists of two parts: (1) is of PR #423 and (2) is of Issue #464. Only the part (2) should be reviewed here. Part (2) starts at commit af39beb.


mor-n4:
Not really part of this review, but I experimented a bit with the null/undef analysis.
```javascript
function foo(p: string) {
	p.length; // why no warning here? (p may be null or undefined)
	if(!p) {
		p.length; // why no warning here? (p IS null or undefined)
	}

	p = bar();
	p.length; // why no warning here? (p may be null or undefined)

	p = undefined;
	p.length; // ok (warning shown)
	if(p) {
		p.length; // ok (no warning shown)
	}
}

function bar(): string {
	// body unknown (we only know bar's signature)
}
```

mor-n4:
Just for reference one example that we already discussed:
```javascript
interface I {}
interface J {}
let value;

if(value instanceof I || value instanceof J) {
	// XPECT FIXME instanceofguard of 'value' 'MayHolds' --> value<:I, value<:J
	// XPECT FIXME instanceofguard of 'value' 'AlwaysHolds' --> value<:(I || J)
	value;
}
```

1. the first XPECT fails, because `value<:I` appears twice in the list in the actual result, which seems odd.
2. the second XPECT is to show that just having the two MayHolds is too weak, because we lose the information that we can be sure that value is at least a subtype of either I or J (or even both). But maybe this can be adjusted later when type guards are finalized.

plugins/org.eclipse.n4js.flowgraphs/src/org/eclipse/n4js/flowgraphs/analysis/EdgeGuide.java
```diff
@@ -8,7 +8,7 @@
  * Contributors:
  *   Marcus Mews - Initial API and implementation
  */
-package org.eclipse.n4js.flowgraphs.analyses;
+package org.eclipse.n4js.flowgraphs.analysis;
```
mor-n4: why rename "org.eclipse.n4js.flowgraphs.analyses" to "org.eclipse.n4js.flowgraphs.analysis"?
For packages "org.eclipse.n4js.flowgraphs.dataflow.guards" and "org.eclipse.n4js.flowgraphs.dataflow.symbols" you also use the plural form.

plugins/org.eclipse.n4js.flowgraphs/src/org/eclipse/n4js/flowgraphs/dataflow/guards/GuardType.java
```diff
@@ -0,0 +1,54 @@
...
+	/**
+	 * For {@link Expression}s that check for 0:<br/>
+	 * {@code if (a == 0)}
+	 */
+	IsZero,
+	/**
+	 * For {@link Expression}s that check for a type:<br/>
+	 * {@code if (a instanceof AType)}
+	 */
+	InstanceOf,
```
mor-n4: as discussed: we should consider moving things like ``isNull``, ``isUndefined``, and ``instanceOf`` and the related logic from the engine to client code (possible draw back: performance?). But this refactoring can be considered when we have more experience with concrete analyses (e.g. type guards).

plugins/org.eclipse.n4js.flowgraphs/src/org/eclipse/n4js/flowgraphs/N4JSFlowAnalyser.java
```diff
@@ -32,13 +36,11 @@
...
 import org.eclipse.n4js.smith.Measurement;
 
-import com.google.common.collect.Lists;
-
 /**
  * Facade for all control and data flow related methods.
  */
-public class N4JSFlowAnalyzer {
-	static private final Logger LOGGER = Logger.getLogger(N4JSFlowAnalyzer.class);
+public class N4JSFlowAnalyser {
```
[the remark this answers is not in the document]
mmews-n4: There are many classes, variables, etc. that contain either `Analy{z|s}er` or `Analysis` or `Analyses` in their names. It was much easier to keep them consistent by using an `s` at all the places, but I can change that.

plugins/org.eclipse.n4js.flowgraphs/src/org/eclipse/n4js/flowgraphs/analysis/EdgeGuide.java
mmews-n4: Actually, I am not sure. I think I renamed this because `analysis` is easer to distinguish from `analysers`, compared to distinguishing `analyses` from `analysers`.
I also see it analog to the package `validation` (would map to `analysis`) which provides the framework, and `validators` (would map to `analysers`) which uses the framework by implementing specific analyses.

plugins/org.eclipse.n4js.flowgraphs/src/org/eclipse/n4js/flowgraphs/dataflow/guards/GuardType.java
mmews-n4: True, this could be refactored to be extensible for client analyses later.

plugins/org.eclipse.n4js.model/src/org/eclipse/n4js/n4JS/DestructureUtils.java
```diff
@@ -0,0 +1,316 @@
...
+
+	/**
+	 * Tells if the given assignment expression is a destructuring assignment, i.e. has an array or object destructuring
+	 * pattern on its LHS.
+	 * <p>
+	 * NOTE: does <b>not</b> check the entire supplementary syntax for destructuring assignments from Section 12.14.5
+	 * "Destructuring Assignment" of the ECMA Script 6 specification. This is done by dedicated validations in
+	 * ASTStructureValidator.
+	 */
+	public static boolean isTopOfAssignment(EObject eobj) {
```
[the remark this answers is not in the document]
mmews-n4: The class name of the call `DestructureUtils.isTop(astElem)` shows the context. Actually, I moved the destructure related methods into the new class `DestructureUtils`; before, they were part of the class `N4JSAST`.

Static methods names can always be too generic when omitting the class name.

mmews-n4:
@mor-n4: Regarding your first two comments.
* We talked about the second comment (about the instanceofguard Xpect tests) and concluded that the implementation can be refined when type guards are implemented).
* I elaborate on the first example below:

```javascript
function foo(p: string) {
L1:	p.length; // why no warning here? (p may be null or undefined)
	if(!p) {
L2:		p.length; // why no warning here? (p IS null or undefined)
	}

	p = bar();
L3:	p.length; // why no warning here? (p may be null or undefined)

	p = undefined;
	p.length; // ok (warning shown)
	if(p) {
		p.length; // ok (no warning shown)
	}
}
```

**Regarding L1**
The analysis is optimistic. Only when there are indications (such as a variable declaration without initializer, or a guard that ensures that a variable must be undefined), a warning is shown. In case of L1, there is no indication that `p` might be null or undefined.

**Regarding L2**
We agreed that a guard that ensures `falsy` does not suffice to issue a warning. If the method `foo` gets called with the empty string `foo("")`, then `!p` is true and the return value of `p.length` is 0.

**Regarding L3**
(see L1)

mor-n4:
> Regarding L2
> We agreed that a guard that ensures falsy does not suffice to issue a warning. If the method foo gets called with the empty string foo(""), then !p is true and the return value of p.length is 0.

Ah, right, had forgotten about that.

mmews-n4:
Thanks a lot for the review! I wait for a last Jenkins build before I squash-merge.

PR #463, opened 2018-01-17
GH-460: As a smith I want to add support for definition site variance in constraint solver

qtran-n4:
Task: https://github.com/eclipse/n4js/issues/460

Issue #460, opened 2018-01-15
As a smith I want to add support for definition site variance in constraint solver


mor-n4:
Currently, the constraint solver is ignoring definition site variance when reducing constraints with generic types. For example, given
```javascript
class B {}
class G<out T> {}
```
the constraint `⟨ G<α> <: G<A> ⟩` is incorrectly reduced to `⟨ α = A⟩`, instead of `⟨ α <: A⟩`.

Full example:
```
class A {}
class B extends A {}
class C extends B {}
class G<out T> {}
class H<T> {}

function <T> foo(p:H<T>):G<T> {
	return null;
}

let a: A;
let b: B;
let c: C;

let hb: H<B>;

// XPECT FIXME noerrors -->
let ga: G<A> = foo(hb);
```

Acceptance Criteria:
* adjust constraint solver accordingly.
* add test (more cases than just the above example).

qtran-n4:
Fixed

PR #473, opened 2018-01-22
GH-460: As a smith I want to add support for definition site variance in constraint solver (PART 2)

mor-n4:
See #460.

1. Return to original solution of PR #463.
2. Improve a heuristic in `InferenceContext` to obtain "nicer" results.


mor-n4:
thanks! builds green, merging ....

Issue #479, opened 2018-01-23
Regression in Type inference on destructuring literals


mmews-n4:
I just noted that one of my tests failed after merging master.

```javascript
function f1() {
	let a : int = null;
	let [b] = [a];
	// XPECT type of "b" --> "int"
	b.toString();
}
```

```javascript
function f2() {
	let a : int = null;
	let {x:b} = {x:a};
	// XPECT type of "b" --> "int"
	b.toString();
}
```

mmews-n4:
```javascript
function f3() {
	let arr : Array<int> = [undefined]; // works
	// XPECT noerrors -->
	for (var a: int of [undefined]) {
			    ^^^^^^^^^
		a.toString();
	}  
}
```

qtran-n4:
Moved the example in the previous to a new ticket: https://github.com/eclipse/n4js/issues/493

mor-n4:
@qtran-n4 shouldn't this be closed? The PR is merged.

PR #480, opened 2018-01-23
GH-479: Regression in Type inference on destructuring literals

qtran-n4:
Task: https://github.com/eclipse/n4js/issues/479


qtran-n4:
@mor-n4 Thanks! Guilds are green, OPR check is fine. Merging...

Issue #467, opened 2018-01-18
Spec ctor accepts enums


dbo:
##### specctor.n4js.xt
```typescript
enum E {
    bar
}
class C {
    constructor(@Spec spec: ~i~this) {}
    public foo?: string;
}
// XPECT errors --> ""
new C(E.bar)

```

dbo:
This also works for `@StringBased` enums.

jpilgrim:
IMHO it's a bug.

mor-n4:
after discussion with @dbo and @jpilgrim:

We decided to keep the old behavior, i.e. let enum literals be subtypes of `Object`.
Rationale:
1. enum literals *are* objects at runtime (and inherit, for example, `hasOwnProperty()`).
2. we intend to change enums in the future to be even more similar to classes (e.g. enums can have methods, like in Java).

However, while looking into this, some other inconsistencies were found:
```
class C {}

let o: Object;
let so: ~Object;


enum E {
    bar
}

// XPECT FIXME noerrors -->
o = E.bar; // currently an error (inconsistent)
// XPECT noerrors -->
so = E.bar; // currently not an error (ok)


@StringBased
enum ES {
    bar
}

// XPECT errors --> "ES is not a subtype of Object." at "ES.bar"
o = ES.bar;  // currently an error (ok)
// XPECT FIXME errors --> "ES is not a subtype of ~Object." at "ES.bar"
so = ES.bar; // currently not an error (inconsistent)
```
After resolving these inconsistencies, the original example from the task description will show an error at least for string-based enums.

PR #516, opened 2018-01-29
GH-467: Spec ctor accepts enums

mor-n4:
See #467.


mor-n4:
manually tested with OPR code

mor-n4:
thanks! Builds green, merging ...

Issue #487, opened 2018-01-24
String based enum with literal `default` does not validate.


yoosiba:
```typescript
@StringBased enum E {
    // XPECT FIXME noerrors --> "no viable alternative at input 'default'"
    default
}
``` 

moved from https://github.com/NumberFour/n4js/issues/373

mor-n4:
This change seems to make sense. Rationale: consistency with names of members in classes (which are allowed to be called "default").

PR #517, opened 2018-01-30
GH-487: String based enum with literal `default` does not validate.

mor-n4:
See #487.

N.b.: requires changes in both repos.


mor-n4:
Builds green.

plugins/org.eclipse.n4js/src/org/eclipse/n4js/N4JS.xtext
```diff
@@ -1307,7 +1307,7 @@ N4EnumDeclaration <Yield>:
 /*
  * Only upper case literals are allows, this is to be checked by the validator
```
yoosiba: `are allowed` ?

plugins/org.eclipse.n4js/src/org/eclipse/n4js/N4JS.xtext
```diff
@@ -1307,7 +1307,7 @@ N4EnumDeclaration <Yield>:
...
  */
-N4EnumLiteral: name=IdentifierOrThis (':' value=STRING)?;
+N4EnumLiteral: name=IdentifierName (':' value=STRING)?;
```
yoosiba: good catch

PR #519, opened 2018-01-30
GH-518 handle null in MemberScopingHelper

### LINKED ISSUE #518: Exception thrown in MemberScopingHelper
In case of broken ASTs `N4JSScopeProvider` can call `MemberScopingHelper.createMemberScope` with `null` for `TypeRef`. That leads to exceptions thrown from Xtend dispatch method, and messes up state of the editor.

yoosiba:
resolved #518


plugins/org.eclipse.n4js/src/org/eclipse/n4js/scoping/members/MemberScopingHelper.xtend
```diff
@@ -113,6 +113,8 @@ class MemberScopingHelper {
 	 * Creates member scope via #members and decorates it via #decorate.
 	 */
 	private def IScope decoratedMemberScopeFor(TypeRef typeRef, MemberScopeRequest memberScopeRequest) {
+		if(typeRef === null)
+			return IScope.NULLSCOPE
```
mor-n4: maybe add comment "broken AST" (to highlight why this fail-safe behavior was added here)

Issue #490, opened 2018-01-24
Return types of FunctionExpressions are incorrect when based on a nested function call's return type


yoosiba:
Hover over the variable to see the incorrect type.
Hover over the fat arrow to see the correct type.

```typescript
function bar():void {}
let v04 = a=>bar();
// XPECT type of 'v04' --> {function(any)}
v04; // v04 is currently {function(any):any}

```
moved from https://github.com/NumberFour/n4js/issues/344

PR #522, opened 2018-01-31
GH-490: Return types of FunctionExpressions are incorrect when based on a nested function call's return type

mor-n4:
See #490.


mor-n4:
test with OPR before merging!

mor-n4:
tested with OPR (working fine)

mor-n4:
thanks, merging!

Issue #533, opened 2018-02-05
Automatically start multibranch builds only for branch "master"


mor-n4:
Multibranch builds should be started automatically only for branch `master`, not for topic branches. However, the Jenkins jobs should still be created for the topic branches, just not started automatically.

This requires a change in the `Jenkinsfile`.

PR #534, opened 2018-02-05
GH-533: Automatically start multibranch builds only for branch "master"

mor-n4:
See #533.


mor-n4:
We decided to do this in two steps. This PR provides the change only for the old Jenkins and provides only a simple solution which will still lead to builds on every push (but these builds are extremely short and are marked as "aborted"/gray).

@gbastkowski will create a task for the CD team about a full-blown solution (probably requires a custom pipeline step) that will be applied (only) to `Jenkinsfile.new`.

Issue #539, opened 2018-02-06
Suspicious argument of a private method in class N4JSResourceDescriptionStrategy


mor-n4:
Method `#createClassUserData()` in class `N4JSResourceDescriptionStrategy` has a parameter `immutableUserData` which is not used in this method, which seems to violate the method's API documentation.

Probably this happened when method `#createClassUserData()` was factored out from method `#internalCreateEObjectDescription()` (compare with old code of that method).

Probably the solution is to change
```java
	private Map<String, String> createClassUserData(final Map<String, String> immutableUserData,
			final TClass tClass) {
		Map<String, String> userData = newHashMap();
```
to
```java
	private Map<String, String> createClassUserData(final Map<String, String> immutableUserData,
			final TClass tClass) {
		Map<String, String> userData = newHashMap(immutableUserData); // <-- sic!
```
but this has to be tested thoroughly.

PR #540, opened 2018-02-06
GH-539: Suspicious argument of a private method in class N4JSResourceDescriptionStrategy

mor-n4:
See #539.


mor-n4:
Tests are green.

Issue #541, opened 2018-02-06
Add 2 internal performance testing utilities to N4JS code base


mor-n4:
During last year's performance investigations (~ Oct 2017) we created a number of performance testing utilities. Most of that was already merged to master (esp. the performance monitoring view) but the following smaller ones still reside on branches and should be merged to master:

1. menu action "mass open" (to open many files at once)
2. resource loading statistics

PR #544, opened 2018-02-06
GH-541: Add 2 internal performance testing utilities to N4JS code base

mor-n4:
See #541.


mor-n4:
This adds two actions to the performance view's drop down menu: "Mass Open" and "Resource Loadings Statistics" (see tool tips for more info).

Issue #595, opened 2018-02-08
Implementing interfaces before extending base class causes warning with odd message


mmews-n4:
For example in this case:

```ts
class A {}
interface I {}

class B implements I extends A  {   
}
```

you get a warning:
Use comma (",") instead of keyword "extends". This warning will become an error soon!

Perhaps an error instead of a warning and a quickfix to rearrange?

copied from https://github.com/NumberFour/n4js/issues/138

PR #622, opened 2018-02-13
GH-595: Implementing interfaces before extending base class causes warning with odd message

mor-n4:
See #595.

I removed the warning altogether (instead of turning it into an error + implementing a quick fix).


mor-n4:
note to self: test with OPR before merging.

mor-n4:
builds green

mor-n4:
builds green

mor-n4:
Thanks for the review!

Works with OPR; merging ...

Issue #450, opened 2018-01-11
As a smith I want to port return-/throw-analysis to use flowgraph engine


mmews-n4:
solves: #449 

Currently, an old implementation is used to identify functions that lack a return or throw statement. This should be replaced to use the flowgraph engine. Doing so will make it easy to solve #449.

PR #627, opened 2018-02-14
GH-450: As a smith I want to port return-/throw-analysis to use flowgraph engine

mmews-n4:
see #450


mmews-n4:
continued here https://github.com/eclipse/n4js/pull/628

Issue #644, opened 2018-02-21
Type of parameter in a spec constructor must not be ~this


mor-n4:
The old `~this` is still accepted as type of a spec parameter, which is incorrect:
```ts
class A {
	// XPECT errors --> "Annotation @Spec may only be used with formal parameters of type ~i~this." at "Spec"
	constructor(@Spec spec: Object) {}
}
class B {
	// XPECT FIXME errors --> "Annotation @Spec may only be used with formal parameters of type ~i~this." at "Spec"
	constructor(@Spec spec: ~this) {}
}
class C {
	// XPECT noerrors -->
	constructor(@Spec spec: ~i~this) {}
}
```

PR #648, opened 2018-02-21
GH-644: Type of parameter in a spec constructor must not be ~this

mor-n4:
See #644.

Tested with OPR (works fine).


mor-n4:
@mmews-n4 this fixes a bug I noticed while working on the chapter for spec constructors in the specification.

mor-n4:
thanks, merging ...

Issue #640, opened 2018-02-20
srclnk macro wrongly creates hyperlinks to generated files


mmews-n4:
In case of polyfilled classes, the srclnk macro always links to the StaticPolyfillAware class, which is ok in general. Still, in case this StaticPolyfillAware class is itself generated, the srclnk macro should link to the StaticPolyfill class instead.

mmews-n4:
Investigation revealed another bug:
Two modules with the same name in different projects are not supported when exporting both of these projects at once.

PR #652, opened 2018-02-26
GH-640: srclnk macro wrongly creates hyperlinks to generated files

mmews-n4:
see: #640


plugins/org.eclipse.n4js.jsdoc2spec/src/org/eclipse/n4js/jsdoc2spec/KeyUtils.java
```diff
@@ -0,0 +1,135 @@
+/**
+ * Copyright (c) 2016 NumberFour AG.
```
qtran-n4: 2018

qtran-n4:
Change the year to 2018. Also, in the future, it is possible better to create a new PR for refactoring changes that are unrelated to the Asciidoc task. Aside from  👍

Issue #659, opened 2018-02-28
Disallow extending a class with a non-accessible constructor


mor-n4:
A.n4js
```javascript
export public class A {
	private constructor() {}
}
```
B.n4js
```javascript
import {A} from "A";

// XPECT FIXME errors --> "Cannot extend class A because its constructor is not accessible." at "A"
class B extends A {}
```

Also look for similar cases (e.g. interfaces?).

dbo:
Why not if the sub class just defines static methods?

mor-n4:
Every class in N4JS has a constructor, at least an implicit one. And that is required to invoke the super class's constructor. But if the super constructor is not accessible, this is impossible.

To disallow instantiation while allowing subclassing, a protected constructor should be sufficient. Don't you think so?

dbo:
No, since it has project visibility.

mor-n4:
Not within the same project, yes. If that is important, you could make the class abstract.

In any case, the inconsistency I mentioned above remains and needs to be fixed. If something is missing we need support for that after the fix.

dbo:
We used to define a private ctor in a common trait/base class to prevent instantiation (like in C++). I think we should allow to define `final abstract` classes to achieve the same effect, although we can't encapsulate this into a trait then.

mor-n4:
Yes, we should think about allowing that. Or we could reconsider the fact that `protected` implies `project` accessibility.

PR #660, opened 2018-02-28
GH-659: Disallow extending a class with a non-accessible constructor

mor-n4:
See #659.


mor-n4:
works fine with OPR

mor-n4:
thanks!

Issue #507, opened 2018-01-24
N4 meta-properties should be read-only


yoosiba:
Currently one can modify for instance the FQN of a type. This should not be possible. Other meta properties should be checked as well.

```typescript
class A { }
console.log(A.n4type.fqn)
// XPECT errors --> "The accessor fqn is read-only." at "fqn"
A.n4type.fqn = 'asd';
console.log(A.n4type.fqn)
```

moved from https://github.com/NumberFour/n4js/issues/36

PR #671, opened 2018-03-07
GH-507: N4 meta-properties should be read-only

mor-n4:
See #507.


mor-n4:
n.b. I chose to not implement "deep immutability" as part of this bug. If we want that, we should create a user story for improving the reflection in general and then add it there.

FYI: all interesting changes are in file `builtin_n4.n4ts`. The rest is just follow-up.

mmews-n4:
Should I still review this?

mor-n4:
tests green
PENDING: test with OPR

mor-n4:
ok with OPR

Issue #653, opened 2018-02-27
As a user I need better error reporting for external libraries


mmews-n4:
follow-up of #713

investigate current state of the library manager and improve it as far as 5 story points allow. try to achieve the following:
- detect changes of index and external workspaces
- add error markers and quickfix when both are out of sync
- revisit install/uninstall actions of library manager

PR #675, opened 2018-03-07
GH-651 As a developer I want the targetplatform file (n4tp) file to be removed

### LINKED ISSUE #651: As a developer I want the targetplatform file (n4tp) file to be removed
Target platform files (`*.n4tp`) are used to define workspace wide dependencies, in particular:

1. used by the headless compiler to instal given set of dependencies priori to compilation of projects
2. used by the IDE to install set of dependencies to make them available to all projects in the workspace.

The `*.n4tp` files are obsolete, as all the information can be derived from `manifest.n4mf` files.
In fact that is already done in IDE with feature added in #512.
Once #654 lands on master, targetplatform file is obsolete from user point of view.

It is expected that targetplatform file will be removed from
* IDE - the maintenance actions in the Library manager
* N4JSC - the command line option for specifying targetplatfrom file

Optionally it has to be checked how targetplatform file is used internally and if it makes sense to remove it completely.

yoosiba:
resolves #651

due to ongoing parallel work on library manager (in #653) this PR only removes support for `n4tp` on user level. Appropriate parts are removed from the N4JSIDE UI and from N4JSC.jar command line options. What is not covered here is refactoring of `NpmManager`, `ExternalLibraryWorkspace` and other deep parts of support for external libraries.


plugins/org.eclipse.n4js.generator.headless/src/org/eclipse/n4js/generator/headless/HeadlessHelper.java
```diff
@@ -80,27 +91,70 @@ public static void registerProjects(List<File> projectLocations, FileBasedWorksp
...
+				.map(HeadlessHelper::fileToURI)
+				.collect(Collectors.toList());
+	}
+
+	private static Stream<File> getProjectStream(List<File> absProjectRoots) {
+		return absProjectRoots.stream()
+				// find all contained folders
+				.flatMap(root -> Arrays.asList(root.listFiles(File::isDirectory)).stream())
+				// only those with manifest
+				.filter(dir -> new File(dir, IN4JSProject.N4MF_MANIFEST).isFile());
```
mmews-n4: streams with comments 👍

Issue #674, opened 2018-03-07
Validations for member redefinition handle @Final for fields incorrectly


mor-n4:
It seems the different semantics of `@Final` for fields compared to those for methods, is not taken into account by the member redefinition validations:
```typescript
class A {
	field: string;
}
interface I {
	field2: number;
}
class B extends A implements I {
	// XPECT FIXME errors --> "must not make an overridden field final" at "@Final"
	@Final @Override field: string = 'hello';
	
	// XPECT FIXME errors --> "must not make an overridden field final" at "@Final"
	@Final @Override field2: number = 42;
}

class C extends B {
	
	// XPECT FIXME noerrors --> "The field C.field cannot override final field B.field."
	@Override field: string = 'hello';
	
	// XPECT FIXME noerrors --> "The field C.field2 cannot override final field B.field2."
	@Override field2: number = 42;
}
```

AC:
1. check spec.
2. fix validation(s) / add new validations if necessary.

PR #684, opened 2018-03-12
GH-674: Validations for member redefinition handle @Final for fields incorrectly

mor-n4:
See #674.


tests/org.eclipse.n4js.spec.tests/xpect-tests/Ch05_04__Redefinition_of_Members/Req70_OverrideFieldFinal.n4js.xt
```diff
@@ -0,0 +1,27 @@
+/*
```
qtran-n4: This is not a review comment but general observation. It is interesting that as opposed to N4JS, if we redeclare a field in the subclass, a new field is created and this new field _shadows_ that field in the superclass.

```java
// JAVA CODE!
class A {
	public String field;

	public void foo() {
		field = "Something else";
	}
}

interface I {
	int field2 = 10;
}

class B extends A implements I {
	public String field = "hello";
	final int field2 = 42;
}

class C extends B {

	String field = "hello";
	int field2 = 42;
}

public class Test {
	public static void main(String[] args) {
		B b = new B();
		b.foo();
		System.out.println(b.field);               // Output: hello
		System.out.println(((A) b).field);     //  Output: Something else
	}
}
```

tests/org.eclipse.n4js.spec.tests/xpect-tests/Ch05_04__Redefinition_of_Members/Req70_OverrideFieldFinal.n4js.xt
```diff
@@ -0,0 +1,27 @@
...
+
+class A {
+	field: string;
+}
+interface I {
+	field2: number;
+}
+class B extends A implements I {
+	// XPECT errors --> "Cannot override non-final field A.field with a final field." at "field"
+	@Final @Override field: string = 'hello';
```
qtran-n4: At the moment, the @Final annotation for a field has two meaning:
- The field is read-only
- The field cannot be overridden in a subclass.

As discussed, we could also consider an alternative semantics of @Final:
- The field is read-only

But it is not part of this PR.

mor-n4:
builds green, OPR code was adapted ...
merging now!

Issue #222, opened 2017-09-07
FQN are supported in type references


yoosiba:
Currently in type references it is possible to use FQN to refer to non global type without any imports.

Assuming module Mod1.n4js with content
```typescript
export public var A :string = "Mod1.A";
```
in module Mod2.n4js it is possible to refer to `A` with FQN like:
```typescript
let a : Mod1.A;
```

Type references to `A` should be allowed only with import:
```typescript
import {A} from "Mod1"
let a : Mod1.A;
```
or with the namespace import
```typescript
import * as namespace from "Mod1"
let a : namespace.A;
```

Notice that using namespace with name as the module only looks like the original example:
```typescript
import * as namespace from "Mod1"
let a : Mod1.A;
```
but this is actually different in a way that import is required and the type ref is not fqn but namespace access.

lbeurerkellner:
Note that this only seems to be an Issue for TypeRefs:

```
// variable declaration
var a : Module.Class

// formal parameter declaration
function f1(m : Module.Class) {}

// return value declaration
function f2() : Module.Class { return null}

// type variable bound declaration
function <T extends Module.Class>generic() {}

// type argument
<Module.Class> generic();
```
(the IDEs validates all of the snippet as valid N4JS, no additional imports were used)

For identifier references however, this seems to be validated correctly:

```
function f(p : any) {}

var a = new Module.Class(); // Couldn't resolve reference to IdentifiableElement 'Module'
f(new Module.Class()); // Couldn't resolve reference to IdentifiableElement 'Module'
```

yoosiba:
Yes, this is only in type refs.
There is a clash, were between namespace and "FQN". In the end without having a context it is hard to say if `foo.Bar` is a FQN access or namespace access. 
Maybe there was some FQN leftover in the handling of the type refs, that is used now to handle the namespace access. 
If I remember correctly `foo.bar.Baz` won't work - namespace imports can have only 2 "segments" so it won't be resolved correctly.

Anyway, there is issue for type refs only. And yes, the handling in scopes is different.

lbeurerkellner:
I am currently looking into this.

PR #686, opened 2018-03-13
GH-222: FQN are supported in type references

mor-n4:
See #222.


mor-n4:
The support of (fully-)qualified names is an Xtext default behavior, which we might need to disable for N4JS to fix this bug. I'm not sure I have found the best way to do this.

@qtran-n4 please check if you can find a simpler/more elegant way of disabling the default FQN support.

mor-n4:
just saw there are test failures on this branch; need to investigate this further, so better postpone the review for now.

mor-n4:
superseded by #739; closing now.

Issue #662, opened 2018-03-01
As a smith I want to expose N4IDL migrations to the runtime


lbeurerkellner:
Currently, developers may declared migrations in N4IDL as follows:

```
@Migration
function migrateA1toA2(a1 : A#1) : A#2 {
      // migrate
}
```

For now, it is still required to explicitly call migrations by their name (e.g. `migrateA1toA2(a)`). However, according to the initial design of N4IDL, we want to make the execution of migrations implicit. That is, the developer only specifies a generic call such as `migrate(a)`. Based on such a call a migration runtime computes the set of migrations to execute.

To allow for runtime code which accesses the set of available migrations, they need to be exposed more generically. More specifically, it is not sufficient to expose them by name, as the runtime does not have the ability to collect them on its own. Therefore, additional code needs to be generated, which registers migrations with their source and target types in such that the migration runtime can access them.

yoosiba:
@lbeurerkellner can we close this? we have merged https://github.com/eclipse/n4js/pull/696

lbeurerkellner:
Yes, this can now be merged.

PR #696, opened 2018-03-16
GH-662: Language support for N4IDL migrations and implicit migrate calls

lbeurerkellner:
This PR introduces the following changes to N4IDL:

* Elevates **N4IDL migrations to first-class citizens in our type model** in the form of the new model type `TMigration`. The types builder now creates TMigration instances from FunctionDeclarations that are annotated as `@Migration`.

* **Support for calling migrations implicitly** via migration calls (function `migrate` in migrations):
```
@Migration function migrateA(a : A#1) : A#2 {
	(...)
	const migratedA : A#2  = migrate(a.a);
	(...)
}
```
These so-called migration calls are statically bound using a strategy based on type distance computation that is very similar to method overloading as it is implemented in Java. At runtime, migration calls are dispatched via the N4IDL migration runtime. The implementation of compile-time migration call binding required changes in the `ASTProcessor`.

* **Transpiler support for migrations** N4IDL migrations are now available at runtime via the hidden static field `migrations__n4` of a versioned type. Via that field, generated type switching functions are exposed. Those return matching migrations for a given list of arguments.

* In the scope of migrations, it is now possible to access an object of new **built-in type `MigrationContext` via the `context` identifier**. The transpiler translates references to the migration context to references to `this.context` which is assumed to be provided by the N4IDL migration runtime.

```
@Migration function migrateA(a : A#1) : A#2 {
	(...)
	const previousRevision = context.getTrace(a)[0];
	(...)
}
```

* Overall, this PR also includes **fixes for minor N4IDL bugs** I noticed during the work on my thesis.


mor-n4:
weird error:
```javascript
class A#1 {
	a:A;
}

@Migration
function migrateA(a : A#1) : A#2 {
	let migratedA = migrate(a.a);
	let x = migratedA; // error: "cannot type IdentifierRef migratedA."
	return null;
}
```

lbeurerkellner:
I addressed your feedback in the following points:

* Changed ASTProcessing to avoid postponing migration calls (avoid cannot type follow up errors)
* Compile-Time migration call binding now still binds an `UnresolvableObjectDescription` but in the type system we now type unresolved migration calls as `any+`
* Added more tests that check for the correct binding of migration calls wrt. forward references 
* Added built-in type implementation to N4BuiltInClasses so that the `MigrationController` and `MigrationContext` interfaces now are defined at runtime and a meta-representations can be obtained (e.g. `MigrationController.n4type`)

lbeurerkellner:
I have merged master and triggered a last CI build. Once those are green, we can merge this as far as I am concerned.

plugins/org.eclipse.n4js/src/org/eclipse/n4js/postprocessing/ASTProcessor.xtend
```diff
@@ -436,6 +451,17 @@ public class ASTProcessor extends AbstractProcessor {
 				// process expression before varDeclOrBindings
 				obj.eContents.bringToFront(obj.expression)
 			}
+			ParameterizedCallExpression case MigrationUtils.isMigrateCall(obj): {
+				// link and type migration arguments first 
+				obj.eContents.bringToFront(obj.arguments)
+			}
+			Script case variantHelper.allowVersionedTypes(obj): {
+				// For variants that support versioned types and therefore migrations,
+				// process the migrations first.
```
mor-n4: please add "process the migrations first, because information about signatures of all migrations will be required when processing migration calls".

lbeurerkellner:
Builds are green and compatibility with existing code has been checked. This can be merged.

PR #734, opened 2018-04-13
GH-730: Smoke test fail on loops

mmews-n4:
see #730


plugins/org.eclipse.n4js.flowgraphs/src/org/eclipse/n4js/flowgraphs/factories/DoWhileFactory.java
```diff
@@ -31,7 +31,7 @@ static ComplexNode buildComplexNode(ReentrantASTIterator astpp, DoStatement doSt
 		ComplexNode cNode = new ComplexNode(astpp.container(), doStmt);
 
 		Node entryNode = new HelperNode(NodeNames.ENTRY, astpp.pos(), doStmt);
-		Node bodyNode = DelegatingNodeFactory.create(astpp, NodeNames.BODY, doStmt, doStmt.getStatement());
+		Node bodyNode = DelegatingNodeFactory.createOrHelper(astpp, NodeNames.BODY, doStmt, doStmt.getStatement());
```
yoosiba: It seems to be unclear when to call `createOrHelper` or `create`. I am happy that it fixes the test, but in the long run, how do we maintain it? And by *we* I mean everyone not only you?

mmews-n4: I can answer the concrete question in three steps:
1) Since the flow graph is created from the AST, nodes are only created iff an AST node exists (e.g. if there is no else-node in the AST, the flow graph does not have a corresponding node). In other words, node are not created if the corresponding AST element is missing.
2) After the nodes in the flow graph have been created, they are connected by edges. These connection procedures assume a flow graph that is based on a valid AST.
3) However, to simplify this edge creation, it is convenient to have some pseudo nodes are created even when the corresponding AST element is missing. This is in particular true for broken ASTs where some arbitrary AST elements are missing notoriously.

I add this as a comment somewhere to the sources.

yoosiba: So... you are saying that it could be done in a way that `create` is always creating node, either real one, or virtual one that allows to connect graphs with edges? Thus, there would be no need for `create` **and** `createOrHelper` ?
I assume analysis of the graphs could check if nodes are not sub-type of this special error case Node (to avoid analyzing broken AST) ?

mmews-n4: It's true that the code should be simplified if possible. It might be so that the `create` method can be replaced by the `createOrHelper` method at many places. This might have a negative impact on runtime performance, but should be evaluated.

PR #739, opened 2018-04-17
GH-222: FQN are supported in type references

mor-n4:
cf. #222


mor-n4:
thanks! builds green, merging ...

PR #749, opened 2018-04-20
adjust test stability

### LINKED ISSUE #748: HLC test with external dependency is unstable
in hotfix 3661aa732fee90aa0af08c1906e9187a1d06a70a test that was failing in some build runs was adjusted to pass in most cases.
Still in some setup test fails, i.e. it behaves like before hotfix. 
Issue seems to be with external library that reexports APIs of its transitive dependencies. That creates impression of non-deterministic behavior since it is hard to control all contributing factors.

Test in question should have its assertions relaxed:
 - we want to assert that installation of  external libraries succeeds,
 - we want to assert that we can execute code from external library,
 - we are **not** interested in asserting details of the external library API

yoosiba:
resolves #748


lbeurerkellner:
Looks good 👍

Issue #552, opened 2018-02-07
Initializer field structural typing is broken


mmews-n4:
The following example reproduces the error.

```javascript
class C {
	public set s(value: int) {}
}
let c: ~i~C;
c.s; // PROBLEM: we receive an error message "The accessor s is write-only." here even though c should have a getter s !
```

The bug also affects the error messages for structural type check:

```javascript
class C {
	public set s(value: int) {}
}

class D {
	//public get s(): int { return null; }
	public set s(value: int) {}
}

let d: D;

//let c1: ~w~C;
//c1.s = 1;

let c: ~i~C = d; // WRONG ERROR MESSAGE: D is not a structural subtype of ~i~C: missing setter s; s failed: non-optional writable field requires a readable field or a getter in subtype.      
```

missing setter s is WRONG. It must be "missing getter s".

copied from https://github.com/NumberFour/n4js/issues/379

mor-n4:
See also #140 (duplicate?).

Issue #140, opened 2017-08-15
Correctly handle scoping members of structural field initializer type ~i~C


qtran-n4:
Currently, in case a `ParameterizedPropertyAccess` is bound to a member of the structural initializer type i.e. `~i~C`, no synthetic member is created (as opposed to composed member)
which is may be OK because the binding is unique. However, we receive `wrong write access` error messages. This example illustrates the issue:

```
class C {
	public set accWrite(v: int) { }
}
let ci: ~i~C;
// XPECT FIXME noerrors
ci.accWrite;
```

Also chains of structural field initializer types are not typed correctly.

```
class C {
	public set accWrite(v: int) {}
}

class G<T> {
	f: ~i~T;
}

let g: G<~i~C>;
// XPECT FIXME type of 'g.f' --> ~i~(~i~C)
g.f.accWrite;
```
## Acceptance Criteria
- The scoping of structural field initializer works correctly, also in the chaining case
- Consider if we may want to create a synthetic member in this case similar to the case with composed members. In the yes case, make sure the handling in both cases is based on the same mechanism.

qtran-n4:
Related to: https://github.com/NumberFour/n4js/issues/12

mor-n4:
It was decided to not create synthetic members in the above cases, for now. We will only reconsider this if we will run into problems with this approach.

PR #750, opened 2018-04-23
GH-140/552: Initializer field structural typing is broken

mor-n4:
See #140 and #552.

Also covers parts of IDE-2965 (will update that task after merging this PR).


mor-n4:
I'll create some more tests, later today, but I think you can go ahead and already have a look.

mor-n4:
Tests green; merging ...

Issue #724, opened 2018-04-11
Editing a location file removes it from search results in find references


yoosiba:
To reproduce

1. Create class `A` in `A.n4js`
2. Create class `B` in `B.n4js`, make `B` extend `A`
3. Search for references of `A`
4. Assert when `B.n4js` is in the search results 
5. Edit `B.n4js`, e.g. add/edit a comment and save.
6. `B.n4js` is **removed** from search results 

Expected behaviour (either):
a) search results are updated with correct information (e.g. `B.n4js` is removed if user removes all references, file `C.n4js` is added to the result when user adds reference)
b) search results do not get updated with edits, and user has to re-run search to get new results.

Note that in JDT behaviour `a` is implemented (but does not have this buggy behaviour).

![screencast 2018-04-11 16-36-26](https://user-images.githubusercontent.com/944287/38624225-e29cfa52-3da7-11e8-983c-f28b30448a63.gif)

qtran-n4:
In the class `ReferenceSearchResultContentProvider`, changed resource description is listened and when that happens the tree node in the search result is removed (see method `descriptionsChanged` in this class). We have a problem here: the `ReferenceSearchResultContentProvider` is bound in a private module is not supposed to be overridden via DI.

qtran-n4:
Fixed.

PR #752, opened 2018-04-23
GH-724: Editing a location file removes it from search results in find references

qtran-n4:
Task: https://github.com/eclipse/n4js/issues/724

Use Guice overriding module to provide a customized ReferenceSearchResultContentProvider


qtran-n4:
@mor-n4 Please review this PR. Thanks!

plugins/org.eclipse.n4js.ui/src/org/eclipse/n4js/ui/MyReferenceSearchResultContentProviderCustomModule.java
```diff
@@ -0,0 +1,44 @@
...
+ * that aims to fix GH-724.
+ */
+@SuppressWarnings("restriction")
+public class MyReferenceSearchResultContentProviderCustomModule extends AbstractModule {
+
+	@Override
+	protected void configure() {
+		binder().install(new PrivateModule() {
+			@Override
+			protected void configure() {
```
mor-n4: Is it possible to somehow define the binding to `MyReferenceSearchResultContentProvider` without copying the other bindings (e.g. `ReferenceSearchViewPage`).

If not (which I guess is the case), then at least add a comment that you here copy the defaults defined in `org.eclipse.xtext.ui.shared.internal.SharedModule` (at the bottom) and that the following has to be kept in line with those defaults.

plugins/org.eclipse.n4js.ui/src/org/eclipse/n4js/ui/search/MyReferenceSearchResultContentProvider.java
```diff
@@ -0,0 +1,37 @@
...
+	 * Constructor
+	 */
+	@Inject
+	public MyReferenceSearchResultContentProvider(IResourceDescriptions resourceDescriptions) {
+		super(resourceDescriptions);
+	}
+
+	@Override
+	public void descriptionsChanged(final Event event) {
+		// Do nothing. Workaround to fix GH-724
```
mor-n4: So it seems you have implemented "behavior b" from the task description, right?
Have you also considered "behavior a"? Would that have been possible to implement?

(doesn't matter much, because I think "behavior b" is fine too, esp. because the search view has a button to conveniently re-run a search)

plugins/org.eclipse.n4js.ui/src/org/eclipse/n4js/ui/MyReferenceSearchResultContentProviderCustomModule.java
qtran-n4: Done. Comments added.

plugins/org.eclipse.n4js.ui/src/org/eclipse/n4js/ui/search/MyReferenceSearchResultContentProvider.java
qtran-n4: "Behavior a" would requires much more effort. We decided to go with behavior b for now and will go for a if it is really needed.

mor-n4: ok!

qtran-n4:
Thanks for your review! Merging now.

PR #782, opened 2018-05-04
GH-780 N4JSC should handle projects with the same project ID

### LINKED ISSUE #780: Duplicate projects are not handled in N4JSC
When user invokes N4JSC with multiple project locations, it may happen that those locations contain multiple projects with the same `ProjectID`. Depending on the order in which locations will be traversed, this might cause builds to fail, even if two projects with the same ID are essentially to identical copies of the same project. 
Additionally if those two projects are different (but have same ID) only one of them will be processed, which again will result in build failure.

Same as in IDE case, the N4JSC does not support compiling multiple projects with the same ID. 
N4JSC needs to handle invalid user input.

(Note, this was discovered after switching  Eclipse Oxygen and Xtext 2.13 but seems unrelated to those two directly. Quite possibly internally order in which some build info has changed, which exposed issue in  the N4JSC).

yoosiba:
resolved #780 

Projects with duplicated `ProjectID` are not correctly handled in IDE in general. This PR adjust behaviour of N4JSC to detect this issue early and abort build.


yoosiba:
@mmews-n4 small update:
* same project (from the same location) passed twice is just ignored on subsequent registrations
* HeadlessHelper takes state of the workspace as initial state for comparisons
* added new test

yoosiba:
@mmews-n4 changes, please take a look.

plugins/org.eclipse.n4js.generator.headless/src/org/eclipse/n4js/generator/headless/HeadlessHelper.java
```diff
@@ -61,50 +62,93 @@ public static void registerProjectsToFileBasedWorkspace(Iterable<URI> projectURI
...
+	 *            of the new project to be considered for registering
+	 * @param projectLocation
+	 *            of the new project to be considered for registering
+	 * @param registeredProjects
+	 *            local cache of already known projects
+	 * @return {@code false} if projects needs to be registered
+	 * @throws N4JSCompileException
+	 *             if project conflicts with project in different location
+	 */
+	private static boolean skipRegistering(File manifest, URI projectLocation, Map<String, URI> registeredProjects)
```
mmews-n4: 👍

plugins/org.eclipse.n4js.utils/src/org/eclipse/n4js/utils/URIUtils.java
```diff
@@ -85,19 +85,27 @@ static public int hashCode(org.eclipse.emf.common.util.URI uri) {
...
 			try {
-				Path realPath = path.toRealPath();
-				return realPath.toString();
+				String newResult = path.toRealPath().toFile().toURI().toString();
+				if (newResult.endsWith("/"))
+					newResult = newResult.substring(0, newResult.length() - 1);
+				result = newResult;
 			} catch (IOException e) {
-				return fileString;
+				// conversion unsuccessful, return original
```
mmews-n4: good

Issue #764, opened 2018-04-25
Adjust scoping of composed types for structural field initializer types


mor-n4:
For example:
```javascript
class C {
	public set acc(value: int) {}
}

let i: ~i~C & Object;
// XPECT FIXME noerrors -->
i.acc;
// XPECT errors --> "setter acc is not available for structurally referenced type ~i~C." at "acc"
i.acc = 42;
```
Make sure to write corresponding tests (and fix them) for all kinds of members and other cases of composed types.

Note: closely related to #552.

PR #786, opened 2018-05-07
GH-764: Adjust scoping of composed types for structural field initializer types

mor-n4:
See #764.


mor-n4:
builds green, merging ...

Issue #759, opened 2018-04-24
Type modifiers dropped / not merged when capturing wildcards with upper bounds


mor-n4:
When capturing `Wildcard`s / `ExistentialTypeRef`s with upper bounds, type modifiers should be properly merged:
```javascript
class C {}

// typing strategy:

class G1<T extends ~w~C> {
	public value: ~i~T;
}
let g1: G1<?>;
let result1 = g1.value;
// XPECT FIXME type of 'result1' --> ~i~C
result1;


class G2<T extends ~i~C> {
	public value: ~w~T;
}
let g2: G2<?>;
let result2 = g2.value;
// XPECT FIXME type of 'result2' --> ~0~C
result2;


class G3<T> {
	public value: ~i~T;
}
let g3: G3<? extends ~w~C>;
let result3 = g3.value;
// XPECT FIXME type of 'result3' --> ~i~C
result3;


class G4<T> {
	public value: ~w~T;
}
let g4: G4<? extends ~i~C>;
let result4 = g4.value;
// XPECT FIXME type of 'result4' --> ~0~C
result4;


// dynamic flag:

class G11<T> {
	public value: T+;
}
let g11: G11<? extends C>;
let result11 = g11.value;
// XPECT FIXME type of 'result11' --> C+
result11;


class G12<T extends C> {
	public value: T+;
}
let g12: G12<?>;
let result12 = g12.value;
// XPECT FIXME type of 'result12' --> C+
result12;
```

PR #788, opened 2018-05-07
GH-759: Type modifiers dropped / not merged when capturing wildcards with upper bounds

mor-n4:
See #759.

Not that many changes, actually, but I had to move things around a bit in `TypeUtils`, so the diff is a bit messy.


mor-n4:
builds green, merging ...

Issue #712, opened 2018-03-29
Initializer field type must be extensible with fields not computed from base type


jpilgrim:
The following tests demonstrates the problem in the type system. Additionally we must ensure that the code would work at runtime. E.g. I'm not quite sure if the compiled code tries to copy the "field accessor" field automatically when using it in the initializer parameter.

```ts
/*
 * In the following test, 
 * the field accessor is correctly NOT added to the initializer field type (see class C).
 * However, it must be possible to add it via a with clause, as shown in class D.
 */

/**
 * Correct.
 */
class C {
    private internal: string;
    public data: string;
    public get getter(): string { return this.internal; }
    
    constructor(@Spec spec: ~i~this) {
        // XPECT errors --> "getter getter is not available for structurally referenced type ~i~C." at getter
        this.internal = spec.getter;
    }
}

// XPECT warnings --> "getter is not defined in ~i~this[C]." at getter
new C({data: "ok", getter: "correct"});

/**
 * Bug
 */
class D {
    private internal: string;
    public data: string;
    public get getter(): string { return this.internal; }
    
    // XPECT noerrors --> The structural field getter (line 33) duplicates getter getter (line 30).
    constructor(@Spec spec: ~i~this with { getter: string}) {
        // XPECT noerrors --> "getter getter is not available for structurally referenced type ~i~C." at getter
        this.internal = spec.getter;
    }
}

// XPECT noiwarnings --> "getter is not defined in ~i~this[C]." at getter
new D({data: "ok", getter: "correct"});
```

mor-n4:
Doesn't look like an issue in the type system, probably just a bug in a validation.

mor-n4:
After looking into this a bit:
* The problem reported here seems to be caused mainly by validation `N4JSMemberValidator#internalCheckDuplicateFieldsIn()`.
* A similar validation for "with"-clauses in general (not in ctor and not with "this"-type) seems to be missing.
* All of this is also closely related to #552, which is a scoping problem, and a proper fix of the above validation(s) might require fixing the scoping issue first.

mor-n4:
For "similar validation for "with"-clauses in general", mentioned above, I have created IDE-2994.

PR #792, opened 2018-05-08
GH-712: Initializer field type must be extensible with fields not computed from base type

mor-n4:
See #712.


mor-n4:
builds green (finally!), merging ...

Issue #771, opened 2018-04-27
N4JS IDE launch configuration reports a missing dependency when run from a smith's IDE


lbeurerkellner:
When running the launch configuration `N4JS__IDE` from a smith's IDE, the following dialog pops up:

![image](https://user-images.githubusercontent.com/17903049/39353258-694cb018-4a07-11e8-8cee-ad0eec26728e.png)

This may indicate that the launch configuration does not correctly configure all dependencies of our N4JS IDE product.

As it seems, this dialog can be dismissed and the N4JS IDE launches as usual, but the cause of this warning still needs to be investigated.

PR #798, opened 2018-05-11
GH-771: Fix N4JS_IDE launch configuration

lbeurerkellner:
Fixes https://github.com/eclipse/n4js/issues/771.

A previous misconfiguration of the included features lead to a validation issue of a dependency with regard to slf4j. This PR fixes this issue.

### DIFF
--- a/builds/org.eclipse.n4js.product.build/N4JS__IDE.launch
+++ b/builds/org.eclipse.n4js.product.build/N4JS__IDE.launch
@@ -28,6 +28,7 @@
 <stringAttribute key="pde.version" value="3.3"/>
 <stringAttribute key="product" value="org.eclipse.n4js.product.product"/>
 <setAttribute key="selected_features">
+<setEntry value="org.eclipse.n4js.sdk:default"/>
 <setEntry value="org.eclipse.e4.rcp:default"/>
 <setEntry value="org.eclipse.ecf.core.feature:default"/>
 <setEntry value="org.eclipse.ecf.core.ssl.feature:default"/>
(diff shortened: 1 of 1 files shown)


builds/org.eclipse.n4js.product.build/N4JS__IDE.launch
```diff
@@ -66,6 +59,8 @@
 <setEntry value="org.eclipse.wst.xml_core.feature:default"/>
 <setEntry value="org.eclipse.wst.xml_ui.feature:default"/>
 <setEntry value="org.eclipse.wst.xml_userdoc.feature:default"/>
+<setEntry value="org.eclipse.epp.mpc:default"/>
```
qtran-n4: Why did you remove `org.eclipse.epp.mpc` bundle from the run configuration? This bundle allows market place. So theoretically, we can try the market place in the IDE started from within the development IDE. Practically, due to some other issues, this didn't work for me but still.

builds/org.eclipse.n4js.product.build/N4JS__IDE.launch
```diff
@@ -66,6 +59,8 @@
...
+<setEntry value="org.eclipse.userstorage:default"/>
```
qtran-n4: This was required by `org.eclipse.epp.mpc` as far as I remember correctly.

qtran-n4:
Can we list the required features (i.e. sdk) in the launch configuration instead of plugins ?

qtran-n4:
We may consider to always start an N4JS IDE using the production definition (as a single source of truth) as discussed. See for instance https://www.modumind.com/2012/11/08/rcp-best-practices-always-run-code-using-a-product-configuration/

builds/org.eclipse.n4js.product.build/N4JS__IDE.launch
lbeurerkellner: According to my tests, this configuration allows for the use of the marketplace in a smith-launched IDE.

qtran-n4: Exactly, that's why it may be a good idea to keep it there?

qtran-n4: Ah it is kept. The diff view is a bit confusing.

Issue #800, opened 2018-05-13
As a smith I need a source map visualizer


jpilgrim:
As a smith I need a source map visualizer so that I can validate generated source maps.

In #799 we added support for Base64 VLQ strings. The transpiler is already prepared for source maps. However it is very complicated to validate whether the correct source maps are generated. In order to allow for validating source maps, a source map view inspired by [source-map-visualization](http://sokra.github.io/source-map-visualization/) would be very helpful.

# Acceptance Criteria

- [x] a source map view which shows the original code, the generated code, and the source map
- [x] the view highlights the mappings in all three views

Issue #799, opened 2018-05-13
As a smith I need base 64 VLQ codec support


jpilgrim:
As a smith I need base 64 VLQ support so that I can implement the source map format used by Chrome and others.

Source maps as described in the [Source Map Revision Proposal 3](https://docs.google.com/document/d/1U1RGAehQwRypUTovF1KRlpiOFze0b-_2gc6fAH0KY0k) require the mappings to be stored as Base 64 [VLQ](https://en.wikipedia.org/wiki/Variable-length_quantity) strings. This is described in detail in the blog entry [Source Maps under the hood – VLQ, Base64 and Yoda](https://blogs.msdn.microsoft.com/davidni/2016/03/14/source-maps-under-the-hood-vlq-base64-and-yoda/).

Note: We know, there exist Java libraries already doing that, but these libraries cannot be used due to IP issues (or it would be more difficult to get them approved than implementing it).

## Acceptance Criteria

- [x] a Base64 VLQ can be decoded into an array of integers
- [x] an array of integers can be encoded into a Base64 VLQ string

PR #801, opened 2018-05-13
As a smith I need base 64 VLQ codec support

jpilgrim:
#799 As a smith I need base 64 VLQ codec support

Added this as a first step toward source maps towards debugging.
Basically one class (Base64VLQ) with tests. The tests are contained in a new plugin project (which includes preference), this is why there are so many files.

This branch/code is needed for #800 .


plugins/org.eclipse.n4js.transpiler/src/org/eclipse/n4js/transpiler/sourcemap/Base64VLQ.java
```diff
@@ -0,0 +1,141 @@
...
+		}
+
+		int[] integers = new int[numCount];
+		int charIndex = 0;
+		int numIndex = 0;
+		while (charIndex < stringLength) {
+			char c = s.charAt(charIndex++);
+			byte b = base64ToByte(c);
+			boolean sign = (byte) (b & 1) > 0;
+			long n = (b & 0b11110) >> 1;
```
mmews-n4: I would always state the complete 8-bit-mask and not omit leading zeros, especially when performing an and-operation (mind other occurrences).

plugins/org.eclipse.n4js.transpiler/src/org/eclipse/n4js/transpiler/sourcemap/Base64VLQ.java
```diff
@@ -0,0 +1,141 @@
...
+			byte shift = 4; // first segment has 4 bits for number
+			while ((b & 0b100000) > 0) {
```
mmews-n4: When checking for specific bits to be set or unset, I would compare against zero, i.e. `== 0` or `!= 0`. The comparison greater `> 0` looks a little like algebra for me (mind one other occurrence).

plugins/org.eclipse.n4js.transpiler/src/org/eclipse/n4js/transpiler/sourcemap/Base64VLQ.java
```diff
@@ -0,0 +1,141 @@
+/**
+ * Copyright (c) 2018 Jens von Pilgrim.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   Jens von Pilgrim - Initial API and implementation
```
mmews-n4: I understand that this will lead to other contributions with authors mentioned in the file header, also from the IDE team. Until now I thought we only add authors when existing code was copied from somewhere else. The new rule would be that author information can be added by anybody iff the file is newly created on the weekend day, right?

PR #803, opened 2018-05-14
Disable automatic trigger for branch builds

gbastkowski:


gbastkowski:
Wrong compare. Closed

Issue #766, opened 2018-04-25
As a developer I want to have a generic JSON editor


lbeurerkellner:
Write a Xtext grammar for JSON. Make validation and some UI parts extensible in order to support special handling of package json.

Maybe there exist ANTLR grammars – but for sure there exist the N4JS grammar for object literals.

Note: The package.json parse should be as robust as possible.

Check if there exists a test suite – in this case we might need to create a CQ!

ANTLR 4 grammar for JSON
https://github.com/antlr/grammars-v4/blob/master/json/JSON.g4

*Acceptance Criteria*
A robust JSON parser with a decent test suite.

PR #811, opened 2018-05-16
GH-766: Third Party JSON Test Suite

lbeurerkellner:
This PR is meant to be merged as part of a CQ.


tests/org.eclipse.n4js.json.testsuite/pom.xml
```diff
@@ -0,0 +1,53 @@
...
+	<build>
+		<plugins>
+		</plugins>
+	</build>
+	
+	<profiles>
+		<profile>
+			<!-- JSONTestSuite -->
+			<id>execute-json-testsuite</id>
+			<activation><activeByDefault>true</activeByDefault></activation>
```
torstenkraemer: Note: "activeByDefault" activation might be misleading here. 
Basically as we use other profiles a lot, this profile will *effectively deactivated*. 

C.f.: https://maven.apache.org/guides/introduction/introduction-to-profiles.html  and the section about `activeByDefault`
This profile will automatically be active for all builds *unless another profile in the same POM is activated using one of the previously described methods*. All profiles that are active by default are automatically deactivated when a profile in the POM is activated on the command line or through its activation config.

Issue #745, opened 2018-04-19
Quick Fix for "node_modules folder and N4JS index are out of sync" does not handle cancellation correctly (InterruptedException)


lbeurerkellner:
When the application of the quick-fix for issue "node_modules folder and N4JS index are out of sync" is cancelled using the 'Cancel' button of its progress monitor, exceptions are thrown on the console.

**Suspected Reason**
The implementation of the Quick Fix does not consider its cancellation via the progress monitor.

**Reproduce**

1. Provoke inconsistent N4JS index state so that the issue appears as a marker in the Problems View
2. In the Problems View, select "Quick Fix" from the context menu of such a marker
3. In the appearing dialog select the corresponding Quick Fix "Synchronize N4JS index to node_modules folder" at press "Finish"
4. Immediately cancel the appearing Progress Dialog using 'Cancel'

On the console output in the hosting developer IDE, the following exceptions can be observed:

```
java.lang.reflect.InvocationTargetException
	at org.eclipse.jface.operation.ModalContext.runInCurrentThread(ModalContext.java:448)
	at org.eclipse.jface.operation.ModalContext.run(ModalContext.java:353)
	at org.eclipse.jface.wizard.WizardDialog.run(WizardDialog.java:993)
	at org.eclipse.ui.internal.views.markers.QuickFixPage.performFinish(QuickFixPage.java:464)
	at org.eclipse.ui.internal.views.markers.QuickFixWizard.lambda$0(QuickFixWizard.java:80)
	at org.eclipse.jface.operation.ModalContext.runInCurrentThread(ModalContext.java:437)
	at org.eclipse.jface.operation.ModalContext.run(ModalContext.java:353)
	at org.eclipse.jface.wizard.WizardDialog.run(WizardDialog.java:993)
	at org.eclipse.ui.internal.views.markers.QuickFixWizard.performFinish(QuickFixWizard.java:85)
	at org.eclipse.jface.wizard.WizardDialog.finishPressed(WizardDialog.java:790)
	at org.eclipse.jface.wizard.WizardDialog.buttonPressed(WizardDialog.java:423)
	at org.eclipse.jface.dialogs.Dialog$2.widgetSelected(Dialog.java:618)
	at org.eclipse.swt.widgets.TypedListener.handleEvent(TypedListener.java:249)
	at org.eclipse.swt.widgets.EventTable.sendEvent(EventTable.java:84)
	at org.eclipse.swt.widgets.Display.sendEvent(Display.java:4256)
[...]

PR #830, opened 2018-05-18
GH-745: Quick Fix for "node_modules folder and N4JS index are out of sync" does not handle cancellation correctly (InterruptedException)

mmews-n4:
see #745

### DIFF
--- a/plugins/org.eclipse.n4js.ui/src/org/eclipse/n4js/ui/quickfix/N4JSQuickfixProvider.xtend
+++ b/plugins/org.eclipse.n4js.ui/src/org/eclipse/n4js/ui/quickfix/N4JSQuickfixProvider.xtend
@@ -83,6 +83,7 @@ import org.eclipse.n4js.n4mf.ProjectType
 import org.eclipse.xtext.EcoreUtil2
 import org.eclipse.n4js.n4mf.ProjectDescription
 import org.eclipse.n4js.ui.changes.ManifestChangeProvider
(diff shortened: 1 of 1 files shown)


plugins/org.eclipse.n4js.ui/src/org/eclipse/n4js/ui/quickfix/N4JSQuickfixProvider.xtend
```diff
@@ -778,6 +779,8 @@ class N4JSQuickfixProvider extends AbstractN4JSQuickfixProvider {
 				new ProgressMonitorDialog(UIUtils.shell).run(true, true, [monitor |
 					try {
 						libraryManager.synchronizeNpms(monitor);
+					} catch (OperationCanceledException e) {
```
yoosiba: ignore ? not even log the exception?

yoosiba: also, why not multi-catch?

mmews-n4: a) at other places, we do nothing either, so I thought that okay. Anyway, I aligned the code more to other places, i.g. `MaintenanceActionsButtonListener:114`
b) xtend does not support multi-catch

Issue #809, opened 2018-05-16
As a developer I want to replace N4MF files by package JSON files


mor-n4:
Instead of defining dependencies etc. in the manifest, we need to use package.json. There must be no redundancies!

Some information may not be found in the package.json by default. We need to decide whether we want a separate file (in json format, similar to typescript) or a section in the package json (similar to linters) containing that information. We would prefer using only package.json -- investigate why TypeScript uses a different file.

We need to be able to read all information from package.json that are important for n4js.

### Acceptance Criteria

- [x] Dependencies, Name, Version etc. are read from package.json
  - this includes supporting semver syntax as defined in package json
- [x] Missing information is available either in package.json or n4js.json depending on decision above
- [x] n4mf related plugins are removed (to be clarified how we do that)
- [x] spec updated
- [x] When no type information is additionally provided, the library manager must NOT change the downloaded NPMs at all. At the moment, the library manager automatically adds the manifest and may perform other changes. This must not happen anymore.

PR #856, opened 2018-05-29
GH-809: copy information from manifest.n4mf to package.json n4js section

mor-n4:
This prepares the `package.json` files in a backward compatible way, by copying all information from each `manifest.n4mf` to a new top-level property `n4js` in the `package.json` in the same folder.

See #809.

### DIFF
--- a/n4js-libs/packages/n4js-es5/package.json
+++ b/n4js-libs/packages/n4js-es5/package.json
@@ -16,5 +16,25 @@
   "files": [
     "manifest.n4mf",
     "src-gen"
-  ]
+  ],
+  "n4js": {
+    "projectType": "runtimeEnvironment",
+    "vendorId": "org.eclipse.n4js",
+    "vendorName": "Eclipse N4JS Project",
+    "output": "src-gen",
+    "sources": {
+      "external": [
+        "src/js"
+      ]
+    },
+    "moduleFilters": {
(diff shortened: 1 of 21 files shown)


n4js-libs/packages/n4js-es5/.npmignore
```diff
@@ -0,0 +1 @@
+/.project
```
yoosiba: if it is generated, add new line in all those file

mor-n4:
@yoosiba @lbeurerkellner 

Thanks for the review!

1. the files with the misssing NL at EOF were created by a script from Daniel. With the added NL chars, the `n4js-libs` is now inconsistent with stdlib, opr, etc. But I guess it does not make a difference, so we can keep it.
2. the dependency to `n4js-es5` was added by a script from Daniel. I will double check with him if it is truly required/desired.

mor-n4:
After discussion with @lbeurerkellner:

we need to remove the dependency to `n4js-es5` from `n4js-node`'s package json, for now, because otherwise during `npm install` in UpdateShippedCode workflow the very out-dated version of `n4js-es5` on npm will be used, instead of the one in the folder under top-level folder `n4js-libs`.

mor-n4:
builds green (except a single test that also failed on master)
merging now ...

Issue #862, opened 2018-05-30
Library Manager does not install types for transitive dependencies


mmews-n4:
Assuming the N4JS project N in the workspace and N has a project dependency to A. A is an external project (i.e. an npm package) that has type definitions in the repository `n4jsd`. Regarding A, the npm itself is installed using `npm -install` and the types are installed by the library manager.

A problem arises when project A is depending on B which also has type definitions. B will be installed by npm, but its type definitions will not be installed by the library manager.

PR #863, opened 2018-05-30
GH-862: Library Manager does not install types for transitive dependencies

mmews-n4:
see #862

### DIFF
--- a/plugins/org.eclipse.n4js/src/org/eclipse/n4js/external/LibraryManager.java
+++ b/plugins/org.eclipse.n4js/src/org/eclipse/n4js/external/LibraryManager.java
@@ -210,10 +210,13 @@ public class LibraryManager {
 	}
 
 	/**
+	 * This method will install all dependencies of the requested NPMs. Since it calls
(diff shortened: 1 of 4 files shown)


plugins/org.eclipse.n4js/src/org/eclipse/n4js/external/LibraryManager.java
```diff
@@ -194,15 +197,50 @@ private IStatus installNPMsInternal(Map<String, String> versionedNPMs, IProgress
 		try (ClosableMeasurement mes = dcLibMngr.getClosableMeasurement("installDependenciesInternal");) {
 
 			List<LibraryChange> actualChanges = installUninstallNPMs(monitor, status, versionedNPMs, emptyList());
+
 			indexSynchronizer.synchronizeNpms(monitor, actualChanges);
 
+			installDependenciesOfNPMs(monitor, actualChanges);
```
yoosiba: wouldn't this need another synchronize?

mmews-n4: 👍

Issue #874, opened 2018-06-01
Library Manager unnecessarily uninstalls and then installs npms


mmews-n4:
This happens when the user requests to install an npm package that is already installed AND the requested version is empty / not given.

PR #875, opened 2018-06-01
GH-874: Library Manager unnecessarily uninstalls and then installs npms

mmews-n4:
see #874

### DIFF
--- a/plugins/org.eclipse.n4js/src/org/eclipse/n4js/external/LibraryManager.java
+++ b/plugins/org.eclipse.n4js/src/org/eclipse/n4js/external/LibraryManager.java
@@ -278,8 +278,8 @@ public class LibraryManager {
 			String versionRequested = Strings.emptyIfNull(reqestedNpm.getValue());
 			if (installedNpms.containsKey(name)) {
 				org.eclipse.emf.common.util.URI location = installedNpms.get(name).getKey();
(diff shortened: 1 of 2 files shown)


plugins/org.eclipse.n4js/src/org/eclipse/n4js/external/LibraryManager.java
```diff
@@ -278,8 +278,8 @@ private void installDependenciesOfNPMs(IProgressMonitor monitor, List<LibraryCha
 			String versionRequested = Strings.emptyIfNull(reqestedNpm.getValue());
 			if (installedNpms.containsKey(name)) {
 				org.eclipse.emf.common.util.URI location = installedNpms.get(name).getKey();
-				String versionInstalled = installedNpms.get(name).getValue();
-				if (versionRequested.equals(Strings.emptyIfNull(versionInstalled))) {
+				String versionInstalled = Strings.emptyIfNull(installedNpms.get(name).getValue());
+				if (versionRequested.isEmpty() || versionRequested.equals(versionInstalled)) {
```
yoosiba: That is improvement, but it would be even better to use `org.eclipse.n4js.utils.Version.compareTo(Version)`
While our implementation does not cover all SemVer cases, and npm extensions, it is better than string comparison.

mmews-n4:
@yoosiba Please have a second look -- Thank you!

Issue #907, opened 2018-06-12
Library manager should install transitive dependencies without building in between


mmews-n4:
Currently, the library manager builds the code and installs transitive dependencies afterwards. This leads to problems in the headless case.

PR #908, opened 2018-06-12
GH-907: Library manager should install transitive dependencies without building in between

mmews-n4:
see #907

This PR fixes:
- bug of #907 
- re-build during NPM-install in case of transitive dependencies

### DIFF
--- a/plugins/org.eclipse.n4js/src/org/eclipse/n4js/external/NpmCLI.java
+++ b/plugins/org.eclipse.n4js/src/org/eclipse/n4js/external/NpmCLI.java
@@ -130,6 +130,9 @@ public class NpmCLI {
 					actualChanges.add(actChg);
 				}
 				subMonitor.worked(1);
+				if (!batchStatus.isOK()) {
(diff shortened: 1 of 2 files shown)


plugins/org.eclipse.n4js/src/org/eclipse/n4js/external/NpmCLI.java
```diff
@@ -172,7 +172,7 @@ private LibraryChange installUninstall(MultiStatus batchStatus, File installPath
 			actualChangeType = LibraryChangeType.Removed;
 		}
 
-		if (packageProcessingStatus != null && packageProcessingStatus.isOK() && batchStatus.isOK()) {
+		if (packageProcessingStatus != null && packageProcessingStatus.isOK()) {
```
yoosiba: Maybe keep the check here, but also add the check in the calling loop,
we want to fail fast. Otherwise user waits an hour to end up in a broken state anyway.

mmews-n4:
@yoosiba Please have another look!

Issue #917, opened 2018-06-14
IllegalStateException: Workbench has not been created yet.


mmews-n4:
**Stacktrace**
```
Caused by: java.lang.IllegalStateException: Workbench has not been created yet.
02:33:06 	at org.eclipse.ui.PlatformUI.getWorkbench(PlatformUI.java:93)
02:33:06 	at org.eclipse.n4js.ui.workingsets.ManualAssociationAwareWorkingSetManager.<init>(ManualAssociationAwareWorkingSetManager.java:74)
02:33:06 	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
02:33:06 	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
02:33:06 	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
02:33:06 	at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
02:33:06 	at com.google.inject.internal.DefaultConstructionProxyFactory$1.newInstance(DefaultConstructionProxyFactory.java:85)
02:33:06 	at com.google.inject.internal.ConstructorInjector.construct(ConstructorInjector.java:85)
02:33:06 	at com.google.inject.internal.ConstructorBindingImpl$Factory.get(ConstructorBindingImpl.java:254)
02:33:06 	at com.google.inject.internal.InjectorImpl$4$1.call(InjectorImpl.java:978)
02:33:06 	at com.google.inject.internal.InjectorImpl.callInContext(InjectorImpl.java:1031)
02:33:06 	at com.google.inject.internal.InjectorImpl$4.get(InjectorImpl.java:974)
02:33:06 	... 83 more
```

PR #918, opened 2018-06-14
GH-917: IllegalStateException: Workbench has not been created yet.

mmews-n4:
see: #917

### DIFF
--- a/plugins/org.eclipse.n4js.ui/src/org/eclipse/n4js/ui/containers/N4JSAllContainersState.java
+++ b/plugins/org.eclipse.n4js.ui/src/org/eclipse/n4js/ui/containers/N4JSAllContainersState.java
@@ -11,12 +11,9 @@
 package org.eclipse.n4js.ui.containers;
 
 import static com.google.common.collect.FluentIterable.from;
-import static org.eclipse.core.runtime.Status.OK_STATUS;
-import static org.eclipse.core.runtime.jobs.Job.INTERACTIVE;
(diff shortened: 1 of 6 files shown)


plugins/org.eclipse.n4js.ui/src/org/eclipse/n4js/ui/containers/N4JSAllContainersState.java
```diff
@@ -155,18 +160,25 @@ private void tryValidateManifest(final IResourceDelta delta) {
 				final ResourceSet resourceSet = core.createResourceSet(Optional.of(project));
 				final Resource resource = resourceSet.getResource(manifestLocation, true);
 				final Job job = Job.create("Update validation markers for " + resource.getURI(), monitor -> {
-					validatorExtension.updateValidationMarkers(manifest, resource, ALL, monitor);
-					return OK_STATUS;
+					ISchedulingRule rule = builder.getRule();
```
yoosiba: @szarnekow we have noticed another race condition triggered from the builder. This seems to fix it, without affecting overall behaviour. Could you take a look too?

mmews-n4: (needed to merge already since the sprint ends now)

szarnekow: I recommend to set the rule on the job itself

PR #941, opened 2018-06-27
 Prepare build-image to container docker-host for DinD

qtran-n4:
In the build log, search for `docker ps` to verify that a call to docker, which requires docker daemon, works.


qtran-n4:
Close before we need a DinD solution.

Issue #968, opened 2018-07-10
NPE when running code in IDE with runtime code installed from npm


yoosiba:
Running code in IDE works only with bundled runtime code. Using runtime code in custom version provided by `npm` yields NPE exception (see stacktrace below).

Steps to reproduce

1. use wizard to create `New N4JS Project` with default greeter file
2. in generated greeter file use runtime code explicitly, e.g.  
```typescript
import {join} from "path";
console.log(join);
``` 
3. right click on a greeter file and run it -> should be no errors.
4. modify `package.json` to use runtime code from npm, e.g.: 
```typescript
{
	"name": "test",
	"version": "0.0.1",
	"n4js": {
		"projectType": "library",
		"vendorId": "org.eclipse.n4js",
		"sources": {
			"source": [
				"src"
			]
		},
		"output": "src-gen",
		"requiredRuntimeLibraries": [
			"n4js-runtime-node"
		]
	},
	"dependencies": {
		"n4js-runtime-node": "0.10.0-alpha.7aa7bbc0"
	}
}
```
5. run library manager to setup dependencies
6. right click greeter file to run it -> code is not executed, exception is logged internally

#### stacktrace
```
!ENTRY org.eclipse.ui 4 0 2018-07-10 18:41:35.018
!MESSAGE Unhandled event loop exception
!STACK 0
java.lang.NullPointerException
	at com.google.common.base.Preconditions.checkNotNull(Preconditions.java:210)
	at com.google.common.base.Optional.of(Optional.java:85)
	at org.eclipse.n4js.runner.RunnerHelper.getExecModuleAsURI(RunnerHelper.java:252)
	at org.eclipse.n4js.runner.RunnerHelper.lambda$8(RunnerHelper.java:146)
	at java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:193)
	at java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:175)
	at java.util.Spliterators$ArraySpliterator.forEachRemaining(Spliterators.java:948)
	at java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:481)
	at java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:471)
	at java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:708)
	at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
[...]

PR #969, opened 2018-07-11
GH-968: NPE when running code in IDE with runtime code installed from npm

lbeurerkellner:
Issue: https://github.com/eclipse/n4js/issues/968

This PR adjusts the package.json files of the shipped n4js runtime environment projects to include the "src/" directories in the published version. Furthermore, the IDE runner was adjusted to not rely on the existence of source containers anymore, when computing the path of the execModule to be executed.


mor-n4:
Builds green, test will be provided in a second PR.
Merging!

Issue #975, opened 2018-07-12
N4JSC.jar injection problem (Jetty port not injected, cannot run tests)


yoosiba:
Currently it is not possible to run tests with N4JSC.jar :
```
42593 INFO  TesterEventBus     - Registered: org.eclipse.n4js.hlc.base.testing.LoggingTestListener@200437e8
42608 WARN  JettyManager       - Port is already in use: 9415. Trying to use next available port: 60559.
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
44682 INFO  JettyManager       - Jetty instance has successfully started on '60559'.
{ FetchError: request to http://localhost:-1/n4js/testing/sessions/3dd163af-704e-411f-be49-1ebe17faa9f9/start failed, reason: connect ECONNREFUSED 127.0.0.1:80
    at ClientRequest.<anonymous> (/Users/me/myproject/node_modules/node-fetch/index.js:133:11)
    at emitOne (events.js:116:13)
    at ClientRequest.emit (events.js:211:7)
    at Socket.socketErrorListener (_http_client.js:387:9)
    at emitOne (events.js:116:13)
    at Socket.emit (events.js:211:7)
    at emitErrorNT (internal/streams/destroy.js:64:8)
    at _combinedTickCallback (internal/process/next_tick.js:138:11)
    at process._tickCallback (internal/process/next_tick.js:180:9)
  name: 'FetchError',
  message: 'request to http://localhost:-1/n4js/testing/sessions/3dd163af-704e-411f-be49-1ebe17faa9f9/start failed, reason: connect ECONNREFUSED 127.0.0.1:80',
  type: 'system',
  errno: 'ECONNREFUSED',
  code: 'ECONNREFUSED' }
```

most important parts are
```
Jetty instance has successfully started on '60559'.
request to http://localhost:-1/
```

Which suggests that there is an issue with setting the port.
Value used suggests the place holder value: https://github.com/eclipse/n4js/blob/master/plugins/org.eclipse.n4js.tester/src/org/eclipse/n4js/tester/internal/DefaultTestTreeTransformer.java#L155
[...]

PR #978, opened 2018-07-13
GH-975: N4JSC.jar injection problem (Jetty port not injected, cannot run tests)

### LINKED ISSUE #611: n4jsc.jar doesn't provide core runtime libraries and core libraries
In contrast to the IDE, the `N4JSC.jar` doesn't provide essential libraries when compiling. In effect developers have to provide these manually.

This is usually about:
- n4js.lang (DI Injector)
- org.eclipse.mangelhaft.assert.*

But also RuntimeLibs and RuntimeEnvironments.

However, in case the developer wants to compile with custom versions of these libraries, this should be possible, e.g.
``` bash
$ java -jar n4jsc.jar -t projects ./my-custom-n4js-runtime-es2015 ...
```
should take precedence over the the implicit `n4js-runtime-es2015` hosted by `n4jsc.jar`.

copied from https://github.com/NumberFour/n4js/issues/167

lbeurerkellner:
Ticket: https://github.com/eclipse/n4js/issues/975

With the changes in this PR, it is assured that the TesterActivator is initialized correctly, when using the headless compiler via the `n4jsc.jar`. This is done by explicitly calling `TesterActivator#startupWithInjector` when the headless injector is created.

Fixes #611 (sort of, shipped code is no longer meant to be shipped with the headless compiler)


tests/org.eclipse.n4js.hlc.integrationtests/src/org/eclipse/n4js/hlc/integrationtests/CompileAndRunTestsTest.java
```diff
@@ -0,0 +1,52 @@
...
+		super("probands/GH-975-tester-ws", true);
+	}
+
+	/**
+	 * Compiles and runs all tests in a simple N4JS test project.
+	 *
+	 * The test project is expected to contain
+	 */
+	@Test
+	public void testRunTests() throws IOException, InterruptedException {
```
yoosiba: maybe add one test where test execution is successful

PR #987, opened 2018-07-17
GH-986: IDE does not allow to launch modules via 'Run with Node.js' if n4js-cli is a declared dependency of the project

### LINKED ISSUE #986: IDE does not allow to launch modules via 'Run with Node.js' if n4js-cli is a declared dependency of the project
In case `n4js-cli` is installed as external library and declared as project dependency, the IDE does no longer offer the option 'Run As > Run with Node.js` in the Run menu.

This is a bug in the computation of the required runtime libraries when determining whether a given runner can be activated or not. In case of `n4js-cli` the IDE detects a dependency cycle (via `eslint` in the devDependencies of the package) and as a consequence does not finish computing the required runtime libraries. 

**Acceptance Criteria**:
Projects that depend on `n4js-cli` can be run via the 'Run As' menu

lbeurerkellner:
Fixes #986 

Due to our switch to `package.json` files, the IDE deals with a larger set of dependencies. This can lead to the detection of dependency cycles within the external library workspace (in this case eslint -> mocha -> eslint). In this issue, this would prevent our runtime library computation in the runner from computing the set of required runtime libraries for projects that referenced a dependency cycle with their project dependencies. In this PR, I have added a recursion guard, so that required runtime libraries can also transitively be computed if dependency cycles are involved.


plugins/org.eclipse.n4js.runner/src/org/eclipse/n4js/runner/RuntimeEnvironmentsHelper.java
```diff
@@ -170,15 +169,23 @@ private void getEnvironemntWithAncestorsRecursive(IN4JSProject project, Set<IN4J
...
+			RecursionGuard<IN4JSProject> recursionGuard) {
 
 		IN4JSProject project = (extractProject(sourceContainer));
 
+		// if recursion is detected, skip this project (already processed)
+		if (!recursionGuard.tryNext(project)) {
+			return;
+		}
+
+		System.out.println("Collecting required runtime libraries from " + project.getProjectId());
```
mor-n4: probably a left-over from debugging

plugins/org.eclipse.n4js.runner/src/org/eclipse/n4js/runner/RuntimeEnvironmentsHelper.java
```diff
@@ -145,11 +146,9 @@ private void getEnvironemntWithAncestorsRecursive(IN4JSProject project, Set<IN4J
 	 * @return list of transitive dependencies of type runtime library, no duplicates
 	 */
 	private List<IN4JSProject> collectRequiredRuntimeLibraries(IN4JSProject project) {
-		if (new SourceContainerAwareDependencyTraverser(project).getResult().hasCycle()) {
-			throw new DependencyCycleDetectedException(project);
-		}
 		Set<IN4JSProject> depsRuntimeLibraries = new HashSet<>();
-		recursiveDependencyCollector(project, depsRuntimeLibraries, p -> isRuntimeLibrary(p));
+		recursiveDependencyCollector(project, depsRuntimeLibraries, p -> isRuntimeLibrary(p),
+				new RecursionGuard<>((new IN4JSProjectEquivalence())));
```
mor-n4: Maybe I don't get it, but it seems the old code used the dependency traverser class only to check for a cycle and if the isn't a cylce, it implemented its own dependency collection. This seems redundant (and odd). Don't we have a utility for this? (maybe the `SourceContainerAwareDependencyTraverser` itself)

tests/org.eclipse.n4js.ui.tests/src/org/eclipse/n4js/tests/externalPackages/GH_986_IntroduceExternalCircularDependencyAndRunTest.java
```diff
@@ -0,0 +1,140 @@
...
+import org.junit.Before;
+import org.junit.Test;
+
+import com.google.common.base.Optional;
+import com.google.inject.Inject;
+
+/**
+ * Installs {@code mocha} and {@code eslint} as project dependency, which will introduce a dependency cycle in the
+ * dependency tree of the test project ({@code eslint} and {@code mocha} depend on each other via
+ * {@code devDependencies}).
```
mor-n4: good to have a test for this!
But could we add a second test with a dependency cycle in the main dependencies (not devDependencies).
Rationale: we might change how we handle `devDependencies` and then this test might become ineffective.

tests/org.eclipse.n4js.ui.tests/src/org/eclipse/n4js/tests/externalPackages/GH_986_IntroduceExternalCircularDependencyAndRunTest.java
```diff
@@ -0,0 +1,140 @@
...
+	 * See JavaDoc of class.
+	 */
+	@Test
+	public void testRunWithCycleInDependencyTree() throws CoreException {
+		final IProject project = createJSProject(CLIENT, "src", "src-gen",
+				b -> b
+						// test project of type 'library'
+						.withType(ProjectType.LIBRARY)
+						.withDependency("mocha", "*")
+						.withDependency("eslint", "*"));
```
mor-n4: Since the test relies on the internals of mocha and eslint (i.e. having a dependency cycle between them) it's important to use fixed versions of these projects. Otherwise the mocha/eslint devs might render our tests ineffective at any time by releasing a new version.

lbeurerkellner:
Thanks for the comments. The last builds have shown some test failures as well, so I would consider this PR a work-in-progress for now.

plugins/org.eclipse.n4js.runner/src/org/eclipse/n4js/runner/RuntimeEnvironmentsHelper.java
lbeurerkellner: As far as I am familiar with our `DependencyTraverser` and `SourceContainerAwareDependencyTraverser`, the sole purpose of these traversers is to detect dependency cycles (e.g. JavaDoc of `DependencyTraverser` say "TRUE as switch result means: found a cycle and should stop.").

lbeurerkellner:
Builds are green, this can be merged.

plugins/org.eclipse.n4js.utils/src/org/eclipse/n4js/utils/DependencyTraverser.java
```diff
@@ -10,28 +10,35 @@
...
-
-import org.eclipse.xtext.xbase.lib.Functions.Function1;
 
 import com.google.common.base.Equivalence;
-import com.google.common.base.Function;
-import com.google.common.base.Optional;
 
 /**
- * TRUE as switch result means: found a cycle and should stop.
+ * A generic dependency traverser for traversing a dependency graph as given by a {@link DependencyVisitor}.
```
mor-n4: as discussed

plugins/org.eclipse.n4js/src/org/eclipse/n4js/validation/helper/SourceContainerAwareDependencyVisitor.xtend
```diff
@@ -0,0 +1,59 @@
...
+import org.eclipse.n4js.n4mf.ProjectType
+import org.eclipse.n4js.projectModel.IN4JSProject
+import org.eclipse.n4js.projectModel.IN4JSSourceContainerAware
+import org.eclipse.n4js.utils.DependencyTraverser.DependencyVisitor
+
+/**
+ * Class for traversing {@link IN4JSSourceContainerAware source container aware} dependencies and
+ * finding cycles in the dependency graph.
+ */
+class SourceContainerAwareDependencyVisitor implements DependencyVisitor<IN4JSSourceContainerAware> {
```
mor-n4: will implement both function interfaces

lbeurerkellner:
Please excuse the last commit message, as discussed, we should correct this when squash merging.

mor-n4:
builds green, merging ...

PR #999, opened 2018-07-22
GH-800: implemented source map support with visualizer

jpilgrim:
#800 

- [x] source maps are created according to standard
- [x] source map visualizer shows generated, original code and mappings
- [x] source map visualizer highlights mappings and regions in all three areas

Restrictions:
- the generation of the mapping information itself by the transpiler has not been changed. The transpiler may should create more fine-grained mapping information.
- indexed mappings (a mapping containing references to other mappings) are not yet supported.


jpilgrim:
non merged due to messed up commits with wrong users. will create a new one.

PR #1000, opened 2018-07-22
GH-800: implemented source map support with visualizer

jpilgrim:
#800 

- [x] source maps are created according to standard
- [x] source map visualizer shows generated, original code and mappings
- [x] source map visualizer highlights mappings and regions in all three areas

Restrictions:
- the generation of the mapping information itself by the transpiler has not been changed. The transpiler may should create more fine-grained mapping information.
- indexed mappings (a mapping containing references to other mappings) are not yet supported.

The IP check failed since I used the wrong user for some commits. I was the only one working on this topic.


plugins/org.eclipse.n4js.ui/src/org/eclipse/n4js/ui/internal/EclipseBasedN4JSWorkspace.java
```diff
@@ -83,6 +83,16 @@ public URI findProjectWith(URI nestedLocation) {
 				&& nestedLocation.segmentCount() >= DIRECT_RESOURCE_IN_PROJECT_SEGMENTCOUNT) {
 			return URI.createPlatformResourceURI(nestedLocation.segment(1), true);
 		}
+		if (nestedLocation.isFile()) {
```
yoosiba: Why would file in eclipse project not be a platform resource? 
In which cases is this fallback needed?

plugins/org.eclipse.n4js.ui/src/org/eclipse/n4js/ui/internal/N4JSEclipseModel.java
```diff
@@ -80,15 +80,15 @@ public N4JSEclipseProject getN4JSProject(URI location) {
 		}
 
 		final String projectName = location.lastSegment();
-		final IProject project;
-		if (location.isFile()) {
+		IProject project;
+		project = workspace.getProject(projectName);
+		if (project == null && location.isFile()) {
 			project = externalLibraryWorkspace.getProject(projectName);
 			checkNotNull(project, "Project does not exist in external workspace. URI: " + location);
```
yoosiba: this check is obsolete now

plugins/org.eclipse.n4js.ui/src/org/eclipse/n4js/ui/internal/N4JSEclipseModel.java
```diff
@@ -128,12 +128,28 @@ public N4JSEclipseProject findProjectWith(URI nestedLocation) {
 
 				if (location.isFile()) {
 					final IN4JSEclipseProject eclipseProject = findProjectWith(location);
-					if (null != eclipseProject && eclipseProject.exists()) {
+					if (eclipseProject != null && eclipseProject.exists()) {
```
yoosiba: whole `org.eclipse.n4js.ui.internal.N4JSEclipseModel.findN4JSSourceContainer(URI)` needs refactoring.

plugins/org.eclipse.n4js.ui/src/org/eclipse/n4js/ui/internal/EclipseBasedN4JSWorkspace.java
jpilgrim: In case of a source map, the file is retrieved via the absolute or relative path in the source map.

plugins/org.eclipse.n4js.ui/src/org/eclipse/n4js/ui/internal/N4JSEclipseModel.java
jpilgrim: That checked caused the "Run As" to not show the possible runners.

plugins/org.eclipse.n4js.ui/src/org/eclipse/n4js/ui/internal/N4JSEclipseModel.java
jpilgrim: Yes. I just changed some minor issues when I read the code and tried to understand it.

jpilgrim:
see #1005

Issue #984, opened 2018-07-16
yarn workspace imported to IDE causes issues


yoosiba:
When importing projects with from git repository IDE is importing also top level project, in particular the yarn workspace. It is useful to do this, root location usually contains a lot of important scripts (e.g. all releng scripts) and top level dependencies. Having this in the IDE allows to edit and mange those from one place.

Unfortunately this breaks IDE behaviour in terms of validations, installing dependencies and other making IDE unusable. Breakage happens due to projects appearing as nested in other projects.

Note that this is N4JS IDE specific issue, as in JDT this does not happen. For example working in Eclipse it is possible to import top level `n4js` as well as nested `n4js/plugins/org.eclipse.n4js` and IDE still works as expected.

Supporting importing level folder, a.k.a. yarn workspace, would be great connivence for yarn based projects.

mor-n4:
> Note that this is N4JS IDE specific issue, as in JDT this does not happen. For example working in Eclipse it is possible to import top level n4js as well as nested n4js/plugins/org.eclipse.n4js and IDE still works as expected.

There is a difference: the top-level "n4js" project is not a Java project, just a plain Eclipse project (with Maven nature, maybe). If having full-blown Java projects nested inside a higher-level Java project, the JDT might have issues too.

Still, it would be nice to support this for N4JS (and it seems almost mandatory once we have support for yarn workspaces). At the very least we should detect this and show an error message (current behavior is very confusing).

yoosiba:
@mor-n4 yes you are right, project is different. Case above was worst mixture of both:
1.  top level project does not have `XtextNature` and it is not processed by the xtext builder
2. top level project has `package.json` 

Due to `1.` it is not real N4JS Project but plain Eclipse Project, but due to `2.` it is considered as N4JS Project by some internals of IDE (but not all of them, e.g. builder).

Note that adding proper nature to the top level to make it real N4JS project, just makes things worse.

Again, supporting importing level folder, a.k.a. yarn workspace, would be great connivence for yarn based projects.

lbeurerkellner:
Some investigation and experimentation has lead me to the following results:

[...]

PR #1004, opened 2018-07-23
GH-984: Separate project description cache invalidation from isAffectingContainerState computation

lbeurerkellner:
Addresses parts of #984 as described in https://github.com/eclipse/n4js/issues/984#issuecomment-406289497.

For performance reasons our `EclipseBasedN4JSWorkspace` caches the contents of project description files (`package.json`). In order to avoid inconsistencies, this cache must be invalidated whenever the underlying project description file changes. Until now, we overrode the method `AbstractAllContainersState#isAffectingContainerState` and invalidated the cache whenever we encountered a `package.json` (formely `manifest.n4mf`) file. In doing that we assumed that all `IResourceChangeEvent`s and resulting `IResourceDelta`s would be passed to our custom implementation of this method. However, as it turns out this assumption does not always hold true. 

When looking at the implementation of `AbstractAllContainersState#resourceChanged` (cf. https://github.com/eclipse/xtext-eclipse/blob/master/org.eclipse.xtext.ui/src/org/eclipse/xtext/ui/containers/AbstractAllContainersState.java#L200), it becomes clear that `isAffectingContainerState` is only invoked until it returns `true`. Once it returned `true`, the traversal of the remaining delta tree is effectively canceled (see variable `clear` in `resourceChanged`). Therefore, we may not be notified of all changed `package.json` files. This can cause our cache to become outdatet which may result in very curious behaviour of the IDE. This behaviour could especially be observed when using the Search And Replace functionality of the IDE or when performing workspace refreshes after IDE-external edits.

In this PR, this issue is addressed by separating the invalidation of the project description cache from the `isAffectingContainerState` method. As an alternative, we now override the `resourceChanged` method as well, and first update the project state (cache) when necessary. After a brief discussion with @mor-n4, we decided not to implement this as a separate `IResourceChangeListener` to guarantee, that we always update the project state before the container state is potentially invalidated and a new build is triggered.


lbeurerkellner:
@szarnekow Could you please have a brief look at this?

plugins/org.eclipse.n4js.ui/src/org/eclipse/n4js/ui/containers/N4JSAllContainersState.java
```diff
@@ -81,7 +87,7 @@
 	private ResourceUIValidatorExtension validatorExtension;
 
 	@Inject
-	private ExternalLibraryBuilder builder;
+	private ExternalLibraryBuilder externalLibraryBuilder;
```
mor-n4: +1 for rename!

plugins/org.eclipse.n4js/src/org/eclipse/n4js/validation/validators/packagejson/PackageJsonValidatorExtension.java
```diff
@@ -158,8 +158,8 @@ public void checkName(JSONValue projectNameValue) {
 			}
 
 			if (platformProjectContainer == null) {
-				throw new IllegalStateException("Failed to determine project name "
-						+ "for resource " + packageJsonUri.toString());
+				// container project cannot be determined, fail gracefully (possibly a nested package.json file)
```
mor-n4: add additional info that we assume that this will only occur in the editor, as nested package.json files are ignored by the builder

mor-n4:
@lbeurerkellner @yoosiba very good catch!

Issue #1012, opened 2018-07-25
NPE and AIOOB exception when using performance view


mmews-n4:
After enabling the performance view and comping O, an exception is thrown when selecting "Build" in the performance view and clicking the camera symbol.

PR #1013, opened 2018-07-25
GH-1012: NPE and AIOOB exception when using performance view

mmews-n4:
see #1012


plugins/org.eclipse.n4js.smith/src/org/eclipse/n4js/smith/TimedDataCollector.java
```diff
@@ -60,7 +60,7 @@ public ClosableMeasurement getClosableMeasurement(String name) {
 		return (TimedMeasurement) getMeasurement(name);
 	}
 
-	private void consume(TimedMeasurement measurement) {
+	synchronized private void consume(TimedMeasurement measurement) {
```
yoosiba: Please add javadoc explaining the issue.

PR #1014, opened 2018-07-25
GH-809: As a developer I want to replace N4MF files by package JSON files

mor-n4:
#809

This contains some clean up and refactorings.


mor-n4:
tested with OPR, works fine!

Issue #821, opened 2018-05-17
As a user I want type definitions of external projects separated from original project


jpilgrim:
As a user I want type definitions of external projects separated from original project so that we do not modify the node-module folder managed by yarn and not modify the original project at all.

Instead we need a concept of separated projects which are related to the original projects. We should use concepts similar to TypeScript (separated and included, the latter is to be supported by original project)

Sub-tasks:
- [x] [sub1] add property to package.json of npm A to specify which npm B in which version range is defined by A
- [x] [sub2] change structure of all projects in the repo n4jsd so that they are standalone npms
- [x] [sub3] change scoping (in particular project imports) and transpiler (all imports) to support import of type definitions using the new property
- [x] [sub4] change library manager to use a different target location for definition projects (separation of projects)
- [x] [sub5] new project type "definition": only n4jsd files, no output folder, no transpile
- [x] [sub6] add implicit project dependency for implementation projects to type project (for client projects with n4js nature); introduce typeDeps section in package.json and its validations; implement ordering of dependencies to support module shadowing
- [x] [sub7] Discovery: How can npm scopes (e.g. '@n4jsd') be supported? (requirements and proposal)

Note regarding the library manager install routine:
 - assume that dependencies of type definition projects are available and do not need to be installed

Note regarding implicit project dependencies:
- this mechanism applies to all projects including those of external libraries
- in case plain JS projects come with some n4jsd files (as a courtesy of the author), the mechanism would also be required although these projects may not have an n4js nature
- The following way to deal with this is proposed:
 - create new project type "plainjs" which also will be the new default for projects that don't specify an n4js section
 - all n4js files in plainjs projects are treated like js files, i.e. their TModules are empty
 - a warning should be shown for imports on n4js files of plainjs projects

Note regarding (transitive) (dev-) dependencies:
- transitive devDependencies are not installed by npm/yarn
- assume that a client project C depends on an n4js project N which depends on npm P
 - there exist a type definition project Pt of P which is necessary to validate N
 - hence, the workspace (its external projects) only compiles when Pt is installed/available
 - remember that Pt would not be installed in the workspace of C
-> Pt must be a (runtime) dependency of N --- type definition projects must not be devDependencies

Restrictions:
- type definition projects cannot have a name of an already existing project (esp. the one of its implementation project)
- one type definition project can define exactly one implementation project

## Acceptance Criteria
- [ ] new concept created and specified
- [ ] new concept is implemented by library manager or other components affected by that in a minimal was
[...]

PR #1031, opened 2018-08-03
GH-821-sub3: As a user I want type definitions of external projects separated from original project

mmews-n4:
see #821


plugins/org.eclipse.n4js/src/org/eclipse/n4js/scoping/utils/ProjectImportEnablingScope.java
```diff
@@ -251,7 +256,12 @@ private IN4JSProject findProject(String projectId, IN4JSProject project) {
 		if (Objects.equals(project.getProjectId(), projectId)) {
 			return project;
 		}
-		for (IN4JSProject p : project.getDependencies()) {
+
+		Iterable<IN4JSProject> dependencies = TypeDefinitionsAwareDependenciesSupplier.get(project);
```
lbeurerkellner: Maybe we can make use of `IResourceScopeCache` so this comparatively expensive operation is only performed once per resource (at least in the scoping).

plugins/org.eclipse.n4js.transpiler.es/src/org/eclipse/n4js/transpiler/es/transform/ModuleWrappingTransformation.xtend
```diff
@@ -330,7 +331,11 @@ class ModuleWrappingTransformation extends Transformation {
 			return moduleSpecifierAdjustment.prefix + '/' + module.moduleSpecifier
 
 		var specifier = completeModuleSpecifier
-		val depProject = n4jsCore.findProject(module.eResource.URI).orNull
+		var depProject = n4jsCore.findProject(module.eResource.URI).orNull
+		if (depProject !== null && depProject.projectType === ProjectType.DEFINITION && depProject.definesPackage !== null) {
+			depProject = n4jsCore.findAllProjectMappings.get(depProject.definesPackage);
```
lbeurerkellner: The results of `n4jsCore.findAllProjectMappings` could also be cached as part of the `TranspilerState`. This would assure that we only compute the map once per transpiled resource.

plugins/org.eclipse.n4js/src/org/eclipse/n4js/scoping/utils/ProjectImportEnablingScope.java
mmews-n4: Thanks! Yes, I introduced the `MultiCleartriggerCache` to take care of this.

plugins/org.eclipse.n4js.transpiler.es/src/org/eclipse/n4js/transpiler/es/transform/ModuleWrappingTransformation.xtend
mmews-n4: Thanks, I also think we should do this in a follow-up sub-task or task!

plugins/org.eclipse.n4js.utils/src/org/eclipse/n4js/utils/MultiCleartriggerCache.java
```diff
@@ -0,0 +1,119 @@
...
+
+import com.google.common.collect.HashMultimap;
+import com.google.common.collect.Multimap;
+import com.google.inject.Singleton;
+
+/**
+ * The {@link MultiCleartriggerCache} is similar to {@link IResourceScopeCache} but provides the means to add more than
+ * one clear-trigger. A clear-trigger is an {@link URI} that will cause the cached value to be deleted.
+ */
+@Singleton
```
mor-n4: as discussed: because it's a singleton, better move to god bundle

plugins/org.eclipse.n4js.utils/src/org/eclipse/n4js/utils/MultiCleartriggerCache.java
```diff
@@ -0,0 +1,119 @@
...
+				Multimap<URI, URI> triggerMap = triggerCache.get(key);
+				for (URI trigger : ratProvider.getCleartriggers()) {
+					triggerMap.put(trigger, reference);
+				}
+
+				ratProvider.postSupply();
+			}
+		}
+		@SuppressWarnings("unchecked")
+		Entry entry = (Entry) entryMap.get(reference);
```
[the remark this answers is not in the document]
mmews-n4: The problem was that it needs to be bound in the ContributingModule which does not support to bind type arguments.

plugins/org.eclipse.n4js/src/org/eclipse/n4js/scoping/utils/ProjectImportEnablingScope.java
```diff
@@ -87,8 +88,10 @@
 	 *            {@link IEObjectDescriptionWithError} will be returned instead of <code>null</code> in case of
 	 *            unresolvable references).
 	 */
-	public static IScope create(IN4JSCore n4jsCore, Resource resource, Optional<ImportDeclaration> importDecl,
+	public static IScope create(IN4JSCore n4jsCore, N4JSModel n4jsModel, Resource resource,
+			Optional<ImportDeclaration> importDecl,
```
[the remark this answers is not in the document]
mmews-n4: I like the idea of a clear structure. However, currently the model is injected at 8 places in our codebase. Maybe we should perform a refactoring so that the model is encapsulated by the core? In this case, the model could even be of package local accessibility and/or not use injection.

mmews-n4:
regarding "IN4JSProject#getDefinesPackage() should return an IN4JSProject"

I started to refactor this and needed to move the findAllProjectMapping (maps name to project) to the N4JSModel which was not easily possible. Hence I reverted the refactoring. I hope this can be done later. Anyhow, I renamed the method to getDefinesPackageName in anticipation of getDefinesPackage which would then return a project.

Issue #1016, opened 2018-07-25
Test Results View does not reflect a process crash/early termination


lbeurerkellner:
Consider the following test class: 

```
import { Assert } from "org/eclipse/n4js/mangelhaft/assert/Assert"

export public class Test_test {

	@Test
	testPass() {
		console.log("A");
		process.exit();
	}

}
```

This test will terminate the process before it gets a chance to report back to the test results view that it will be terminated. As a consequence the red terminate button in the test view does not reflect when the process terminates early.

When working on this issue, cherry-pick the fix from GH-767.

Issue #767, opened 2018-04-25
As a user I want to run/test things from the Yarn workspace folder


jpilgrim:
Instead of running things from a temporarily created folder, I want to run things from the yarn workspace folder and re-use the yarn installed node_module folder.

mmews-n4:
Is this ticket (and related branch) still valid with respect to our efforts on switching to package.json and yarn workspace support?

mor-n4:
This issue is superseded by `#1154` and `#824`. Closing as duplicate.

PR #1032, opened 2018-08-06
GH-767: Handle early tester process termination more gracefully

lbeurerkellner:
Fixes https://github.com/eclipse/n4js/issues/1016

This PR improves the behaviour of the test results view in the N4JS IDE.

In case the tester process terminates earlily, the test results view still maintained the appearance of running tests (test cases without status, "running" as overall execution status).

With this PR, an early termination of the tester process is handled as a test error. If during a test case the process terminates unexpectedly, the case will be marked with an error status and all remaining test cases with a skipped status. Furthermore the case with error status will give information on the early termination when selected. As a consequence, the stop button of the results view is now also updated accordingly.

This PR was based on @jpilgrim's work on branch GH-767. Based on his changes, I also ported a minor improvement in `NodejsLaunchConfigurationTab.xtend` that avoids a non-deterministically shuffled order of configured environment variables due to the underlying `Map`-based implementation.


plugins/org.eclipse.n4js.tester/src/org/eclipse/n4js/tester/TesterFrontEnd.java
```diff
@@ -190,7 +195,42 @@ public Process test(TestConfiguration config, IExecutor executor) {
...
+
+		return process;
+	}
+
+	/**
+	 * Thread running in parallel which waits for a given process to terminate.
+	 *
+	 * On termination
+	 */
+	private final class TesterTerminationListener extends Thread {
```
yoosiba: could be in its own file. Also I don't get dangling `On termination`

Issue #1033, opened 2018-08-06
[mangelhaft] no option to disable timeouts for debugging


dbo:
Timeouts are hit you while debugging, we should allow to disable them by N4JS runtime option.

mmews-n4:
@qtran-n4 can this be closed?

qtran-n4:
Fixed

PR #1034, opened 2018-08-06
GH-1033: allow disabling timeouts 

dbo:
@qtran-n4 please


qtran-n4:
@dbo I will merge this PR when the builds are green.

PR #1035, opened 2018-08-07
GH-821: Filter type dependencies in dependency collection of the "big button"

lbeurerkellner:
Ticket: #821 

In this PR I made the following changes:

* I exposed the origin/type of a dependency (runtime, development or type as declared in the "dependencies", "devDependencies" and "typeDependencies" sections) in the ProjectDescription model via a new `DependencyType` enum type. Upon removal of type dependencies, this can be reduced to runtime and development.

* In the collection logic of the "big button", I filter the missing dependencies by their type (only collect non-type dependencies)

This PR is a preparation for merging the last subtask of #821, as it restores the backward compatibility of the introduced "typeDependencies". On the current master, the "big button" logic will attempt to install missing type dependencies. However, this fails since we do not publish type definitions yet.


mmews-n4:
Thank you!

Issue #1036, opened 2018-08-07
Investigate recent failures of nightly builds


mor-n4:
Recently the nightly builds were unstable. This issue is about briefly investigating these failures and implementing minor fixes, if possible. More extensive fixes should be done as part of separate issues.

mor-n4:
Added some minor fixes / work-arounds (see PRs linked above) and created other tickets for further investigation.

Closing this as FIXED.

PR #1038, opened 2018-08-07
GH-1036: Investigate recent failures of nightly builds

mor-n4:
See #1036.


mor-n4:
Builds had failures, not merging yet.

mor-n4:
Ran many builds and there was only a single other failure of a test for which a Jira ticket already exists (IDE-2270).

So merging now ...

PR #1040, opened 2018-08-07
GH-821: Make sure target definition package.json is re-created properly

lbeurerkellner:
This PR fixes a bug that was introduced by one of the subtasks of #821. 

On the current master, an invocation of the "big button" will delete the target definition `package.json` file but never re-create it. As a consequence, all following `npm install`-calls will output the following warnings:

```
npm WARN saveError ENOENT: no such file or directory, open '/.../workspace/.metadata/.plugins/org.eclipse.n4js.external.libraries/.n4npm/package.json'
npm notice created a lockfile as package-lock.json. You should commit this file.
npm WARN enoent ENOENT: no such file or directory, open '/.../workspace/.metadata/.plugins/org.eclipse.n4js.external.libraries/.n4npm/package.json'
```

More importantly however, the external `node_modules` folder managed by the IDE, will no longer have a corresponding `package.json` file. This can lead to unpredictable behaviour of npm when it tries to prune the `node_modules` folder and remove "unused" dependencies, as in "not listed in the package.json".

The following changes were applied to fix the bug:

* The method `TargetPlatformInstallLocationProvider#repairNpmFolderState` will now also re-create the target definition `package.json` file.
* Increase re-use of the `maintenance*` methods from `ExternalLibraryPreferencePage.java`. As a consequence, the maintenance action implementations that are used by the "big button" and by the library manager UI now overlap more (they still use slightly different versions of some actions, but I did not want to touch that too much)


mmews-n4:
Thank you!

Issue #1045, opened 2018-08-09
SemVer parser fails to parse specific version requirements with 's' and 'e' characters in prerelease tag


lbeurerkellner:
The following version requirements, as to be found in the packages `fbjs` and `es6-promise` resp., cannot be parsed by the SemVer parser:

```
"dependencies": {
	// XPECT FIXME error --> "Invalid version number "^2.1.0-revise": mismatched character '<EOF>' expecting 'm'".
	"A": "babel-preset",
	// XPECT FIXME error --> "Invalid version number "^2.1.0-revise": mismatched character '<EOF>' expecting 'm'.".
	"B": "^2.1.0-revise"
},
```

PR #1049, opened 2018-08-09
GH-1045: SemVer parser fails to parse specific version requirements with 'se' characters in prerelease tag

mmews-n4:
see #1045


mmews-n4:
Thanks!

Issue #1050, opened 2018-08-10
As a developer I want to install a missing dependencies using a quickfix


mmews-n4:
Functionality was lost when switching from manifest to package.json files. To re-enable it, a new extension point must be introduced in the bundle json.ui.

Write content assist and quickfixes similar to the current manifest, but now for the package.json.
### Acceptance Criteria
- [sub1]
  - [x] Bring back “Install Dependency” quick fix
  - [x] Recover “changeProjectTypeToValidation” quick-fix (see history of N4JSQuickfixProvider)
- [sub2]
  - [x] content assist and quickfixes as in n4mf
  - [ ] ~~spec updated~~
  - [x] Project reference hyperlinking
  - [ ] ~~Consider implementing support for ISemanticModification for the JSON language (e.g. partial formatting and serialization, also see PackageJsonChangeProvider)~~

PR #1051, opened 2018-08-10
GH-1050: As a developer I want to have UI support (content assist) for package.json

mmews-n4:
This solves sub1 of #1050


mmews-n4:
Thank you

Issue #957, opened 2018-07-09
Fix issues when a workspace project depends on one of the published runtime NPMs


mor-n4:
When a project in the workspace depends on a runtime project in a different version than the version provided as part of the shipped code, then the "Install missing dependencies" button does not install the correct version, probably because a project of that name is already present. However, the library manager supports shadowing shipped code projects by manually installed projects.

mor-n4:
Pending: FIXME's in file `ExternalLibrariesInstallHelper.java`!

mor-n4:
## Pending: FIXME's in file ExternalLibrariesInstallHelper.java + tests!

PR #1060, opened 2018-08-14
GH-957: Fix issues when a workspace project depends on one of the published runtime NPMs

mor-n4:
See #957.

This PR includes:
1. a refactoring of parts of the library manager to avoid converting version requirements to strings too early,
2. some refinements related to the changes of GH-957 that were pending.


plugins/org.eclipse.n4js.ui/src/org/eclipse/n4js/ui/wizard/dependencies/ExternalLibrariesInstallHelper.java
```diff
@@ -102,22 +112,24 @@ private void removeDependenciesToShippedCodeIfVersionMatches(Map<String, String>
...
+			Map<String, NPMVersionRequirement> versionRequirementsOfShippedCodeToInstall = projectNamesOfShippedCodeToInstall
 					.stream()
 					.map(id -> dependenciesToInstall.get(id))
-					.collect(Collectors.toSet());
-			if (versionConstraintsOfShippedCodeToInstall.size() > 1) {
-				// FIXME GH-957 warn about conflicting version constraints for shipped code!
-			}
-			String versionConstraint = versionConstraintsOfShippedCodeToInstall.stream().findFirst()
-					.orElse(null);
+					.collect(Collectors.toMap(SemverSerializer::serialize, Function.identity(), (vr1, vr2) -> vr1));
```
mmews-n4: Does it work with multiple version requirements that have a solution, like `>0.1.0` and `0.1.1`?

plugins/org.eclipse.n4js/src/org/eclipse/n4js/projectModel/dependencies/DependencyInfo.java
```diff
@@ -46,12 +53,16 @@ private static String toName(ProjectReference projectReference) {
...
-			version = SemverSerializer.serialize(prjDep.getVersionRequirement());
+			NPMVersionRequirement vr = prjDep.getVersionRequirement();
+			if (vr != null) {
+				return vr;
+			}
 		}
-		return version;
+		// by default, use an empty version rage set requirement (corresponds to empty string in package.json
+		// dependencies):
+		return SemverFactory.eINSTANCE.createVersionRangeSetRequirement();
```
mmews-n4: How about using the utils method here to create an empty version requirement?

plugins/org.eclipse.n4js.ui/src/org/eclipse/n4js/ui/wizard/dependencies/ExternalLibrariesInstallHelper.java
mor-n4: No, this is a very simple equality check. Otherwise we'd need sophisticated logic to check if one version requirement intersects with another, one implies another, etc. That's why I added a warning in case of multiple requirements remaining.

PR #1063, opened 2018-08-14
GH-1062: CSV export for performance data

### LINKED ISSUE #1062: [subtask 1018] As a smith I need to export collected performance data as CSV files
In order to statistically track the performance implications of the changes we make, we need to be able to compare before-and-after performance data for large-scale scenario, such as those provided by our integration tests.

By integrating CSV export with our existing performance tooling, we can add an export step to the after-methods of several tests. In this export step, we can export collected performance data which can in turn be archived as an artefact of the CI build.

Using these performance artefact, we will then be able to compare performance improvements and degradations across multiple revisions and builds.

lbeurerkellner:
Fixes #1062 

This PR consist of two main contributions:

* **CSV export of `DataCollector` instances**: The UI now exposes a new button `Snapshot and Save` that allows to take a performance snapshot and at the same time export the currently collected data as a CSV file (See Performance Graph View).
* **Performance data collection support in the headless compiler:** The headless compiler has been instrumented to make use of the `DataCollector` API. In order to obtain the collected performance data, one may use the new command line options `--performanceReport` and `--performanceKey` or the system environment variable `N4JSC_PERFORMANCE_REPORT`. Documentation on these options is to be found in `N4jscBase` since I assume that a smith would know about the internals of the headless compiler anyway.


mmews-n4:
👍

Issue #1018, opened 2018-07-25
As a user I want the LibraryManager to use new n4jsd npms 


mmews-n4:
Preferred solution: show all contents of the node_modules folder
Alternative: library manager only shows direct dependencies of the workspace projects

Adjust existing projects accordingly.

### Acceptance Criteria
- [x] remove package.json-markers and support for these markers, i.e. get rid of the markers in the (hidden) node_modules folder
- [x] adjust LibraryManager (remove transitive installations, adaptions, etc.)
- [x] adjust tests to no longer directly use N4JSD sandbox repo (either use published npms, npm@github, or verdaccio)
- [x] investigate performance and solve them if possible, otherwise put into separate task (10% increase would be ok since we will improve performance by installing all libraries at once later on)

PR #1098, opened 2018-08-28
GH-1018: As a user I want the LibraryManager to use new n4jsd npms

mor-n4:
See #1018.


plugins/org.eclipse.n4js.smith/src/org/eclipse/n4js/smith/TimedDataCollector.java
```diff
@@ -64,7 +62,7 @@ public ClosableMeasurement getClosableMeasurement(String name) {
 	/** This method must be synchronized to protect race conditions when calling data.add */
 	synchronized private void consume(TimedMeasurement measurement) {
 		TimedMeasurement timed = measurement;
-		data.add(new DataPoint(timed.name, timed.elapsed(NANOSECONDS)));
+		// data.add(new DataPoint(timed.name, timed.elapsed(NANOSECONDS)));
```
yoosiba: why is this needed? should not affect anything if you run without performance data collection and is undesired if you run with it turned on. So why?

mor-n4: Because Luca had turned on performance data collection on Jenkins, by default, leading this line to collecting >700MB `DataPoint`s and crashing the JVM during integration tests.

yoosiba: Then this is wrong fix.
Performance data collection should not be turned on by default, exactly because of this reason.

Issue #916, opened 2018-06-13
As a developer I want the library manager to install NPMs in a batch


mmews-n4:
The current implementation installs each NPM in a single call to NPM. This takes 4 minutes altogether when done on OPR. Tests showed that this duration can be decreased down to 1 minute if only one call to NPM is used instead. The drawback is, that this call takes longer and cancelling is more difficult.

### Acceptance Criteria
* use batch install
* cancellation of call to NPM is optional
* add quickfix "Install all missing npms"
* remove existing multi-fix and improve for single installation

PR #1101, opened 2018-08-31
GH-916: As a developer I want the library manager to install NPMs in a batch

mmews-n4:
see #916


mmews-n4:
tested with OPR: works. merging to topic branch of #1018.

PR #1108, opened 2018-09-03
GH-1107: Loader mode for API not impl, don't set by default, exit code improvements

### LINKED ISSUE #1107: n4js-mangelhaft marks broken dependencies as SKIPPED
The tool sets the loader's mode to be tolerant on missing dependencies, i.e. treats it as "API not implemented". The problem is that this mechanism applies on broken dependencies or uncompiled modules, too. We should only selectively set this mode when needed per n4js runtime option.

dbo:
fixes #1107 

@yoosiba please

### DIFF
--- a/n4js-libs/packages/n4js-mangelhaft-cli/src/n4js/org/eclipse/n4js/mangelhaft/runner/node/NodeTestAPI.n4js
+++ b/n4js-libs/packages/n4js-mangelhaft-cli/src/n4js/org/eclipse/n4js/mangelhaft/runner/node/NodeTestAPI.n4js
@@ -199,7 +199,6 @@ export default public class NodeTestAPI {
         const args = NodeTestCLI.toCommandLine(options);
         args.unshift(consoleTestRunner.join("/"));
 
-        env.N4JS_RT_test_mode = "1";
         env.N4JS_RT_keep_eventloop = "1";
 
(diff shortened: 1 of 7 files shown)


yoosiba:
@qtran-n4 build&release please

n4js-libs/packages/n4js-mangelhaft-cli/src/n4js/org/eclipse/n4js/mangelhaft/runner/node/NodeTestAPI.n4js
```diff
@@ -199,7 +199,6 @@ export default public class NodeTestAPI {
         const args = NodeTestCLI.toCommandLine(options);
         args.unshift(consoleTestRunner.join("/"));
 
-        env.N4JS_RT_test_mode = "1";
```
qtran-n4: shouldn't `env.N4JS_RT_test_mode` be set somewhere?

dbo: No, I want regular tests to fail when module loading fails. Just in case of stdlib spec tests (with not yet implemented API) this mode will be turned on.

qtran-n4:
Builds are fine. Merging...

PR #1109, opened 2018-09-04
Prepare IDE for transition to scoped n4jsd packages

mmews-n4:


plugins/org.eclipse.n4js/src/org/eclipse/n4js/external/NpmCLI.java
```diff
@@ -64,7 +64,7 @@
 
 	/** Simple validation if the package name is not null or empty */
 	public boolean invalidPackageName(String packageName) {
-		return packageName == null || packageName.trim().isEmpty() || packageName.contains("@");
+		return packageName == null || packageName.trim().isEmpty() || packageName.endsWith("@");
```
mor-n4: Consider using
```
return packageName == null || packageName.trim().isEmpty()
    || (packageName.contains("@") && !ProjectDescriptionUtils.isProjectNameWithScope(String));
```

mmews-n4: good idea, I think we should consider this for GH-1018!

Issue #1113, opened 2018-09-06
Local publishing of n4js-libs interferes with public publishing 


qtran-n4:
Local publish changes the package's version in lerna.json to 9999. This version is then used as the public version. Use `git checkout lerna.json` before public publishing.

PR #1114, opened 2018-09-06
GH-1113: Local publishing of n4js-libs interferes with public publishing

qtran-n4:


qtran-n4:
@dbo I am using a fixed version in `lerna.json` instead of `git checkout`.

Issue #1116, opened 2018-09-11
As a smith I want n4js-libs to be published with versions without a pre-release segment


mor-n4:
Currently, the version scheme for n4js-libs is as follows:
```
0.11.0-alpha.date.commit
```
For example:
```
0.11.0-alpha.20180910.882c614f
```
However, the use of a pre-release version segment (i.e. everything after the dash) leads to a number of issues:
1. different behavior of npm vs. yarn related to pre-release version segments when using an empty string as version requirement.
  Compare:
  `npm install "n4js-runtime-node@"` will install v0.11.0-alpha.20180910.3d12f9a4
  `yarn add "n4js-runtime-node@"` will install v0.6.0
2. ranges across versions with different major/minor/patch segment cannot be defined to include pre-release versions.

Therefore, this user story is about changing the publication scripts to use a version scheme without pre-release segment.

New version scheme:
```
0.12.p
```
where `p` will be increased automatically upon every publication. Later, major and minor segments may be incremented manually.

ACs:
1. adjust publication scripts for n4js-libs to use versions without pre-release segment
2. since the version no longer indicates the date/commit, the publication scripts should automatically add some meta-info elsewhere (before publication):
   1. add an appropriate key/value pair to the package.json (check what is common)
   2. add a readme file with date, commit, and link to n4js-libs folder on GitHub (showing the correct commit!) and make sure this readme is shown on pages such as https://www.npmjs.com/package/n4js-es5

mor-n4:
STATUS:
* most functionality has been merged with the PR linked above.
* pending: (1) readme (2) allow manually bumping of major/minor version segments.

PR #1117, opened 2018-09-12
GH-1116: As a smith I want n4js-libs to be published with versions without a pre-release segment

mor-n4:
See #1116.

This PR contains the main functionality, but there will be further PRs for this ticket.


qtran-n4:
Nice script! 👍 Before merge, one thing to you may want to consider removing
```
   # GH-1113: Since lerna.json may have been changed by local publishing, reverse all changes
    # We only publish if there are changes in n4js-libs since the last commit
    #PKG_VERSION=`cat lerna.json  | jq -r '.versionfixed' | xargs -t semver -i minor {}`
```
Background: This publishing script is used for publishing npms to local registry (e.g. during integration test) as well. After local publishing, `lerna` changes the version in `lerna.json` to the published version.

Issue #1121, opened 2018-09-18
Fix failing test due to error related with npm installation using the sandbox


mmews-n4:
Test in `ErrorsInExternalLibrariesPluginTest` fails and it seems to be related with the npm install of a sandbox project using a git-version during installation. The test fails only on Jenkins. 

**Output on Jenkins**
```
[17-09-2018 07:22:57] Installing NPM(s): MyLib@git+file:///tmp/1537204977437-0/ (protocol: git+file, url: /tmp/1537204977437-0/)
npm WARN targetplatform@ No description
npm WARN targetplatform@ No repository field.
npm WARN targetplatform@ No license field.

added 12 packages in 0.487s
npm WARN targetplatform@ No description
npm WARN targetplatform@ No repository field.
npm WARN targetplatform@ No license field.

+ MyLib@0.3.1
added 1 package and removed 11 packages in 0.422s
npm WARN targetplatform@ No description
npm WARN targetplatform@ No repository field.
npm WARN targetplatform@ No license field.

added 11 packages in 1.054s
[17-09-2018 07:23:01] External libraries built: MyLib, encoding, iconv-lite, is-stream, n4js-es5, n4js-node, n4js-runtime-es2015, n4js-runtime-html5, n4js.lang, node-fetch, safer-buffer, systemjs, when
```

**Output locally** (does neither install 11 other packages nor built 12)
```
npm WARN tasks@0.0.1 No description
npm WARN tasks@0.0.1 No repository field.
npm WARN tasks@0.0.1 No license field.

+ MyLib@0.3.1
added 1 package in 0.382s
```

PR #1126, opened 2018-09-20
GH-1121: Fix failing test due to error related with npm installation using the sandbox

mmews-n4:
see #1121


testhelpers/org.eclipse.n4js.ui.tests.helper/src/org/eclipse/n4js/tests/builder/ExternalLibrariesSetupHelper.java
```diff
@@ -57,15 +56,28 @@ public void setupExternalLibraries(boolean initShippedCode) throws Exception {
 			assertTrue("Error while saving external library preference changes.", result.isOK());
 		}
 
+		Injector n4jsInjector = N4JSActivator.getInstance().getInjector("org.eclipse.n4js.N4JS");
```
qtran-n4: Why can't we inject the instance directly?

mmews-n4: Because the test runner does not support that, unfortunately.

mmews-n4:
thanks!

Issue #1122, opened 2018-09-18
Jump to definition and searching for inherited fields does not work in external libraries


qtran-n4:
When we are inside a text editor displaying an N4JS file of an external library, jump to definition does not work. Moreover, when we search for inherited fields via Ctrl + 0, the inherited fields are not displayed. The outline view does not display inherited fields either.

This severely affects the usability of the IDE especially when working in a large codebase with huge external dependencies.

@yoosiba's comment: Suspicion: Navigation does not work to declarations for which only type information is there (and no N4JS code). These items are also not shown in the outline view for the same reason.

qtran-n4:
@yoosiba You can also add a 👍  :D

yoosiba:
enough emojis for my taste

PR #1129, opened 2018-09-21
GH-1122: Jump to definition and searching for inherited fields does not work in external libraries

mmews-n4:
see #1122

### DIFF
--- a/tests/org.eclipse.n4js.ui.tests/src/org/eclipse/n4js/tests/hyperlink/HyperlinkPluginTest.java
+++ b/tests/org.eclipse.n4js.ui.tests/src/org/eclipse/n4js/tests/hyperlink/HyperlinkPluginTest.java
@@ -23,6 +23,7 @@ import org.eclipse.core.runtime.MultiStatus;
 import org.eclipse.emf.common.util.URI;
 import org.eclipse.jface.text.IRegion;
 import org.eclipse.jface.text.Region;
+import org.eclipse.jface.text.TextSelection;
 import org.eclipse.jface.text.hyperlink.IHyperlink;
 import org.eclipse.jface.text.source.ISourceViewer;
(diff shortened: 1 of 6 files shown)


tests/org.eclipse.n4js.ui.tests/src/org/eclipse/n4js/tests/hyperlink/HyperlinkPluginTest.java
```diff
@@ -0,0 +1,128 @@
...
+		XtextEditor editorProcess = (XtextEditor) uriEditorOpener.open(uriProcess, true);
+		sourceViewer = editorProcess.getInternalSourceViewer();
+		region = new Region(556, 12);
+		IHyperlink[] hlinksInProcess = hyperlinkDetector.detectHyperlinks(sourceViewer, region, true);
+ 
+		assertTrue("Hyperlink in external library missing", hlinksInProcess != null && hlinksInProcess.length == 1);
+		assertTrue("Hyperlink must be of type XtextHyperlink", hlinksInProcess[0] instanceof XtextHyperlink);
+		XtextHyperlink hyperlinkToEvent = (XtextHyperlink) hlinksInProcess[0];
+		URI uriEvent = hyperlinkToEvent.getURI();
+		assertTrue("Hyperlink URI must be a file uri", uriEvent.isFile());
```
mor-n4: in addition to this assertion, we could test that URI `uriEvent` can actually be opened and the correct region is selected in the newly opened editor (by using the same infrastructure methods that the original hyperlink code is using; for testing the selection there is a public method`#getSelection()` in the editor, I think).

mmews-n4: 👍  I added further logic to assert that the selection text is as expected.

mmews-n4:
thanks!

PR #1132, opened 2018-09-26
GH-1116: As a smith I want n4js-libs to be published with versions without a pre-release segment (part 2)

mor-n4:
See #1116.

This adds version information to the read-me files of all n4js-libs (dynamically when publishing).


mor-n4:
This can only be fully tested by publishing to the public npm registry, so merging without review. Local tests were fine.

Issue #1135, opened 2018-09-27
Introduce cache for ApiImplMapping in transpiler


mmews-n4:
subtask of #1124 

Calculating the ApiImplMappings has a time complexity linear to all Projects which is currently on an internal project about 1300. Doing this for every resource takes ~50 seconds. Since the result does not change during one build, the result can be cached to reduce the time effort (from actually 100s since this method is called twice for every resource) down to ~0.5s.

PR #1136, opened 2018-09-27
GH-1135: Introduce cache for ApiImplMapping in transpiler

mmews-n4:
see #1135


tests/org.eclipse.n4js.ui.tests/src/org/eclipse/n4js/tests/misc/ClearCacheOnCleanPluginUITest.java
```diff
@@ -0,0 +1,171 @@
...
+		IResourcesSetupUtil.fullBuild();
+		waitForAutoBuild();
+
+		// use key of API_IMPL_MAPPING
+		SupplierWithPostAction testSupplier = new SupplierWithPostAction();
+		assertTrue(testSupplier.postActionTriggerCount == 0);
+
+		cache.clear(MultiCleartriggerCache.CACHE_KEY_API_IMPL_MAPPING);
+
+		// cache should be empty
```
qtran-n4: I don't understand this part. How can testing that cache is empty be done with the following code?

mmews-n4: Good question; when the method `get` is called it will invoke the supplier which will then increment the counter `postActionTriggerCount`. Hence, if the counter is equal to `1` it must have been zero before calling `get`. That means that the cache could not have contained the value of the supplier (before calling `get`).

mmews-n4:
thanks!

Issue #1138, opened 2018-09-28
Bug in string template literals sequences


dbo:
in general escape sequences should be kept "as is" in the compiled output, e.g.
```js
`\\`, `\t` ...
```
Currently these are rewritten, replaced by tab character etc.

PR #1140, opened 2018-10-01
GH-1138: Bug in string template literals sequences

mor-n4:
See #1138.


mor-n4:
ecma test suite is green (tested locally)

PR #1141, opened 2018-10-01
GH-1116: As a smith I want n4js-libs to be published with versions without a pre-release segment (part 3)

mor-n4:
See #1116.

This adds the possibility to conveniently bump the major/minor version segments of n4js-libs packages via file `n4js-libs/version.json`.


releng/utils/scripts/publish-n4js-libs.sh
```diff
@@ -110,32 +110,58 @@ if [ "${NPM_TAG}" = "latest" ]; then
...
+        exit -1
+    fi
+    if [ "$VERSION_MAJOR_REQUESTED" -gt "$VERSION_MAJOR_PUBLIC" ]; then
+        echo "New major version segment requested in file n4js-libs/version.json -> will bump major segment"
+        PUBLISH_VERSION="$VERSION_MAJOR_REQUESTED.$VERSION_MINOR_REQUESTED.0"
+    elif [ "$VERSION_MINOR_REQUESTED" -gt "$VERSION_MINOR_PUBLIC" ]; then
+        echo "New minor version segment requested in file n4js-libs/version.json -> will bump minor segment"
+        # for major segment we use the latest public version as template to make sure
+        # we do not end up with a lower version than the latest public version
+        PUBLISH_VERSION="$VERSION_MAJOR_PUBLIC.$VERSION_MINOR_REQUESTED.0"
```
mmews-n4: I would also be ok if we could set major and minor numbers directly here.

Issue #1124, opened 2018-09-19
As a smith I want to improve the build performance of external and user workspace


mmews-n4:
When building projects in general, the build is slowed down when the index contains a lot of resources.

AC
* check clients of findAllProjects (regarding ~1300 projects in the index)
* check implementation of findProjectWith and findAllProjects
* look for linear searches in IN4JSCore, etc.

Subtasks:
- [x] (#1135) Introduce cache for `ApiImplMapping` in transpiler
- [ ] other

mmews-n4:
Note: In `AbstractJSONValidatorExtension#getIssueSeverities()` all issues are changed into warnings iff the package.json is of a plainJS project. This might be changed when thinking about turning off/on validations for package.jsons / plainJS projects.

PR #1148, opened 2018-10-08
GH-1124: As a smith I want to improve the build performance of external and user workspace (part 1)

mor-n4:
See #1124.


mor-n4:
builds green, merging ...

PR #1149, opened 2018-10-09
GH-1124: As a smith I want to improve the build performance of external and user workspace (part 2)

mor-n4:
Mainly refactoring of data collection. Check commit messages.

See #1124.


plugins/org.eclipse.n4js.smith/src/org/eclipse/n4js/smith/TimedMeasurement.java
```diff
@@ -35,7 +35,7 @@ synchronized long elapsed(TimeUnit timeUnit) {
 	}
 
 	@Override
-	synchronized public void end() {
```
mmews-n4: `synchronized` was necessary before to avoid multithread problems. are you sure you can just remove it?

mor-n4: good point! didn't remove that on purpose

PR #1139, opened 2018-09-28
GH-1133: As a smith I want to reduce the size of the Xtext index

mmews-n4:
see #1133

PR #1150, opened 2018-10-09
GH-1124: As a smith I want to improve the build performance of external and user workspace (part 3)

mor-n4:
Disables validation of package.json in external workspace.

See #1124.


mor-n4:
NOT MERGING THIS, for now, because it would introduce another case of different handling in user vs. external workspace and after merging PR #1152 the gain would only be 10s in "reload external libraries" in OPR and after merging PR #1139 the gain would be almost zero.

Issue #1146, opened 2018-10-06
n4jsx emits warnings for svg elements


dbo:
JSX allows to define svg elements. N4JSX files emit warnings for some svg elements like `circle`, `path`, `g`.

##### foo.n4jsx.xt
```
// XPECT nowarnings --> "Tag g is an unknown HTML tag."
  <g>
  </g>

```

see https://reactjs.org/docs/dom-elements.html

PR #1166, opened 2018-11-05
GH-1146: n4jsx emits warnings for svg elements

mor-n4:
See #1146.


mor-n4:
thanks. builds green, merging ...

Issue #1183, opened 2018-11-29
Content assist not using alias in case of existing imports


mor-n4:
To reproduce:
```
import {SomeClass as SC} from "otherModule";

Some // <--
```
When doing content assist in last line after "Some" and selecting "SomeClass" from the list, the IDE replaces text "Some" by "SomeClass" which leads to an unresolved reference error. Instead, it should replace "Some" by "SC" (i.e. the alias of the existing import).

PR #1184, opened 2018-11-29
GH-1183: Content assist not using alias in case of existing imports

mor-n4:
See #1183.


mor-n4:
builds green, merging ...

Issue #926, opened 2018-06-21
As a user I want content assist to be faster and better sorted


mmews-n4:
As a user I want content assist to be faster and better sorted.

In particular invoking content assist for functions takes very long (in OPR/stdlib). It must be much faster!
The list of results should be organized better, e.g., the shortest prefix match should be the first hit.

This task is a bit unspecific on purpose - we need to investigate what we can actually do.

Ideas:
- Maybe looking for proposals can be made asynchronously (i.e. not UI blocking, and new proposals get added to the list when it is already opened).
- have different CA tabs, one only providing elements available in the current scope (which maybe can be computed without using the index!), one showing all elements (which definitely requires the index)

Ask OPR team for hints.

### Acceptance Criteria
- CA list shows up faster as at the moment
- proposal are better sorted

mmews-n4:
Current status:
* In https://github.com/eclipse/n4js/pull/933 a first improvement of the performance was merged (and icon support)
* Missing better sorting of proposals
* Missing even better time performance

Regarding
* wrong proposals: see other tickets like https://github.com/eclipse/n4js/issues/1142
* current time performance: This seems to be a result of the Xtext scoping architecture. Improving the situation here would therefor mean to either improve Xtext scoping or bypass and re-implement a faster version for content assist.

kduske-n4:
Here's an idea for sorting the proposals: Please put variables names at the top and sort them by scope, i.e., local variables first, then parameters, then attributes, etc.

Also rank the proposals by how well they match the case of the typed text, i.e. if I type `availa`, and there is a type `AvailableWhatever` and a variables name `availabilityStatus`, then the latter should go to the top because it matches the case better.

PR #1187, opened 2018-12-04
GH-926: As a user I want content assist to be faster and better sorted

mor-n4:
See #926.


mor-n4:
@szarnekow if you have a moment, it would be great if you could have a quick look at this.

Does it make sense or am I overlooking something?
Could it make sense doing this for Xtext in general?

plugins/org.eclipse.n4js.ui/src/org/eclipse/n4js/ui/proposals/imports/ImportsAwareReferenceProposalCreator.java
```diff
@@ -102,6 +103,10 @@ public void lookupCrossReference(
 			Function<IEObjectDescription, ICompletionProposal> proposalFactory) {
 
 		if (model != null) {
+			if (PickyCompletionProposalAcceptor.hasSeenScopeFor(acceptor, model, reference)) {
```
szarnekow: Here you could try to call `org.eclipse.xtext.ui.editor.contentassist.AbstractJavaBasedContentProposalProvider.announceProcessing(List<?>)` with list containing a unique first element, e.g. `"lookupCrossReference"` and the model and reference. That should already do the trick for a given ContentAssistContext.

Issue #1188, opened 2018-12-04
Some tests not executed in maven build


mor-n4:
The following tests from bundle `org.eclipse.n4js.hlc.integrationtests` are not executed on Jenkins:
```
CompileAndRunTestsAllPassTest.java
CompileAndRunTestsWithFailuresTest.java
N4jscJarTesterTest.java
PerformanceReportTest.java
```
This is caused by a misconfiguration. In this bundle, only test classes with a name ending in `JarTest` or `CliTest` are being executed.

Solution: rename test classes.

Bonus: do some clean up of the tests in this bundle, e.g. change `N4jsCliTest` to not inherit from `AbstractN4jscJarTest`.

PR #1189, opened 2018-12-04
GH-1188: Some tests not executed in maven build

mor-n4:
See #1188.


mor-n4:
Pending clean up:
change `N4jsCliTest` to not inherit from `AbstractN4jscJarTest`.

mor-n4:
builds green, merging ...

Issue #1046, opened 2018-08-09
java.lang.NullPointerException when refreshing external libraries


dbo:
[tt.log](https://github.com/eclipse/n4js/files/2273459/tt.log)

PR #1191, opened 2018-12-06
GH-1046: quickfix: java.lang.NullPointerException when refreshing external libraries #1046

jpilgrim:
#1046 

quickfix for NPE: external projects are not cleaned
and also added documentation of library manager to n4js

Needs to be reviewed and probably "really" fixed.


jpilgrim:
After further examination, I'm wondering why anything is working after this "quickfix". Won't merge that one, continue investigation.

Issue #1190, opened 2018-12-06
NullPointerException when running the Setup External Libraries Wizard


jpilgrim:
The N4JS IDE (Dec. 2 2018) I get a NullPointerException when running the Setup External Libraries Wizard (colleagues as well).

There is an old issue with the same title at #1046 -- but I assume it is a different issue now (at least the exception is thrown somewhere else.

*This is a blocker!*

```
java.lang.reflect.InvocationTargetException
	at org.eclipse.jface.operation.ModalContext.run(ModalContext.java:401)
	at org.eclipse.jface.wizard.WizardDialog.run(WizardDialog.java:977)
	at org.eclipse.n4js.ui.wizard.dependencies.ExternalLibrariesWizard.performFinish(ExternalLibrariesWizard.java:72)
    ...
Caused by: java.lang.NullPointerException
	at org.eclipse.xtext.builder.impl.ToBeBuiltComputer.doRemoveProject(ToBeBuiltComputer.java:203)
	at org.eclipse.xtext.builder.impl.ToBeBuiltComputer.removeProject(ToBeBuiltComputer.java:180)
	at org.eclipse.n4js.ui.external.ExternalLibraryBuilder$BuildOperation$2.getToBeBuilt(ExternalLibraryBuilder.java:464)
	at org.eclipse.n4js.ui.external.ExternalLibraryBuilder$BuildOperation.run(ExternalLibraryBuilder.java:513)
	at org.eclipse.n4js.ui.external.ExternalLibraryBuilder$BuildOperation.access$3(ExternalLibraryBuilder.java:495)
	at org.eclipse.n4js.ui.external.ExternalLibraryBuilder.doPerformOperation(ExternalLibraryBuilder.java:341)
	at org.eclipse.n4js.ui.external.ExternalLibraryBuilder.clean(ExternalLibraryBuilder.java:290)
	at org.eclipse.n4js.ui.external.ExternalLibraryBuilder.clean(ExternalLibraryBuilder.java:271)
	at org.eclipse.n4js.ui.external.EclipseExternalLibraryWorkspace.deregisterProjectsInternal(EclipseExternalLibraryWorkspace.java:270)
	at org.eclipse.n4js.ui.external.EclipseExternalLibraryWorkspace.deregisterAllProjects(EclipseExternalLibraryWorkspace.java:251)
	at org.eclipse.n4js.external.LibraryManager.installNPMsInternal(LibraryManager.java:223)
[...]

PR #1192, opened 2018-12-10
GH-1190: NullPointerException when running the Setup External Libraries Wizard

mor-n4:
See #1190.

See my comment here: https://github.com/eclipse/n4js/issues/1190#issuecomment-445810533


jpilgrim:
Maybe we can write a regression test for that?

jpilgrim:
and create a bug report in Xtext to make org.eclipse.xtext.builder.impl.ToBeBuiltComputer.doRemoveProject(IProject, IProgressMonitor) more robust: 

```
onlyOnThisProject = isSameProject 
           || storage2Project.getSecond()==null
           || !storage2Project.getSecond().isAccessible();
```

PR #1195, opened 2018-12-11
GH-926: As a user I want content assist to be faster and better sorted

mor-n4:
See #926.

This PR adds a minor fix to avoid some incorrect proposals during content assist.


mor-n4:
builds green ...

Issue #1199, opened 2018-12-12
Jump to definition / hyperlink not working in external libraries when opening file from Project Explorer


mor-n4:
Hyperlinks (e.g. jump to definition with CMD+Click) are not working in source files located in the external libraries IF the file was opened by double-clicking its entry in the Project Explorer view of Eclipse. If, on the other hand, the file was opened by navigating to it via another hyperlink OR by using "Open Type", everything works as intended.

To reproduce:

1. create a new N4JS project.
2. add the following dependencies:
   ```
   "dependencies": {
   	"@n4jsd/express": "4.0.0",
   	"@n4jsd/body-parser": "1.18.3",
   	"express": "4.0.0",
   	"body-parser": "1.18.3"
   }
   ```
3. install missing dependencies.
4. create a new N4JS file with:
   ```javascript
   import BodyParser from "body-parser"
   BodyParser;
   ```
5. CMD-click on "BodyParser".
   --> In the file that was opened now, hyperlinks are working as intended.
6. close that file and reopen it using the Project Explorer view.
   --> Now, hyperlinks are NOT working.

For a possible way how to fix this, see #1122.

mor-n4:
Simpler way to reproduce (with fewer dependencies):
as above, but use
```
"dependencies": {
	"n4js-runtime-node": ""
}
```
and
```javascript
import {Server} from "http";
Server;
```
instead.

PR #1200, opened 2018-12-17
GH-1199: Jump to definition / hyperlink not working in external libraries when opening file from Project Explorer

mor-n4:
See #1199.


mor-n4:
Builds green, merging ...

Issue #1206, opened 2019-01-02
Maven build fails with Xtend compile errors since latest Eclipse release


mor-n4:
Since the night from Dec 19 to Dec 20, 2018 the maven build fails with Xtend compile errors, produced by the `xtend-maven-plugin`. Note that no changes were merged to master during that time. Might be related to the release of Eclipse 2018-12 (i.e. new versions of plugins available that are among the transitive dependencies of our build).

To reproduce:
```
mvn clean compile
```
Excerpt of the build log:
```
((...))
05:04:29 [INFO] ------------------------------------------------------------------------
05:04:29 [INFO] Building org.eclipse.n4js.releng.utils 0.0.1-SNAPSHOT
05:04:29 [INFO] ------------------------------------------------------------------------
05:04:29 [INFO] 
((... some lines omitted ...))
05:04:29 [INFO] --- xtend-maven-plugin:2.13.0:compile (default) @ org.eclipse.n4js.releng.utils ---
05:05:03 [WARNING] Compilation of stubs had errors.
05:05:07 [ERROR] 
05:05:07 ERROR: 	N4JSAntlrGrammarGenerator.xtend - /jobs/genie.n4js/nightly/workspace/n4js/releng/org.eclipse.n4js.releng.utils/src/org/eclipse/n4js/antlr/N4JSAntlrGrammarGenerator.xtend
05:05:07 16: org.eclipse.n4js.antlr.CodeIntoGrammarInjector cannot be resolved to a type.
05:05:07 [ERROR] 
05:05:07 ERROR: 	N4JSAntlrGrammarGenerator.xtend - /jobs/genie.n4js/nightly/workspace/n4js/releng/org.eclipse.n4js.releng.utils/src/org/eclipse/n4js/antlr/N4JSAntlrGrammarGenerator.xtend
05:05:07 17: org.eclipse.n4js.antlr.UnicodeKeywordHelper cannot be resolved to a type.
05:05:07 [ERROR] 
05:05:07 ERROR: 	N4JSAntlrGrammarGenerator.xtend - /jobs/genie.n4js/nightly/workspace/n4js/releng/org.eclipse.n4js.releng.utils/src/org/eclipse/n4js/antlr/N4JSAntlrGrammarGenerator.xtend
05:05:07 18: org.eclipse.n4js.antlr.n4js.AutomaticSemicolonInjector cannot be resolved to a type.
((...))
```

mor-n4:
Pending:

[...]

PR #1213, opened 2019-01-10
GH-1206: Maven build fails with Xtend compile errors since latest Eclipse release

mor-n4:
Just some minor clean up in parent pom, that I wanted to test separately from the task's main changes.

#1206


mor-n4:
Builds green, merging ...

Issue #1215, opened 2019-01-14
As a smith I want to make tests in N4JS repo compatible with latest node/yarn version


mor-n4:
When running the builds with the latest node/npm/yarn version, around 20 tests are failing (including the stdlib integration test).

To reproduce:
bump version of node to `10.13.0` and yarn to `1.12.3`, e.g. as with commit 345f6e7d75743fa8ccb8e725fd4a2de85b3fc292. Then run Jenkins builds.

PR #1217, opened 2019-01-14
GH-1215: As a smith I want to make tests in N4JS repo compatible with latest node/yarn version

mor-n4:
See #1215.


mor-n4:
builds green, merging ...

Issue #1225, opened 2019-01-23
Nightly build on Eclipse fails due to the lack of docker (used for Eclipse helper generation)


qtran-n4:
See for example: https://ci.eclipse.org/n4js/job/nightly/87/

PR #1226, opened 2019-01-23
GH-1225: Nightly build on Eclipse fails due to the lack of docker (used for Eclipse helper generation)

qtran-n4:
Task: https://github.com/eclipse/n4js/issues/1225


torstenkraemer:
Hi @qtran-n4 ! I'm not sure what happens here or happened in the past - but the "dist" folder should not be checked in -  neither any large binaries like PDFs. 
This will generate commit-noise and bloat the repository quickly ...

torstenkraemer:
@qtran-n4 but as far as I understand these HTML pages are generated in the build-process and not hand-written. So they should not be checked in.

qtran-n4:
@torstenkraemer Good point. They have been there. I have deleted the `dist` and `src-gen` folders and let's wait for the build.

qtran-n4:
Thanks! The build result looks fine. Merging...

Issue #1248, opened 2019-02-14
Adjust N4JS Tasks Example and some tests to recent changes to the library manager


mor-n4:
Some adjustments required:

1. adjust N4JS Tasks Example (including corresponding test).
2. adjust tests that are currently on ignore (search for `@Ignore("side-by-side-use-case")`).

This is a follow-up of these user stories: `GH-1154/GH-1214/GH-824`.

PR #1249, opened 2019-02-14
GH-1248: Adjust N4JS Tasks Example and some tests to recent changes to the library manager

mor-n4:
See #1248.


mor-n4:
builds green, merging ...

Issue #1252, opened 2019-02-18
Revisit error messages in case npm/yarn is not installed after recent changes to the library manager


mor-n4:
1) Check how the current implementation (before this tasks) deals with the case that npm is not installed and align the yarn handling to that error behavior.

2) Make sure this error behavior is satisfactory. In particular, the error message should tell the user how to install node (for npm) / yarn!!

PR #1254, opened 2019-02-19
GH-1252: Revisit error messages in case npm/yarn is not installed after recent changes to the library manager

mor-n4:
See #1252.


mor-n4:
builds green...

Issue #1260, opened 2019-02-22
Validation does not respect dynamically accessed members that are also defined as static members


mmews-n4:
Consider this:
```ts
class Foo {
//	s: string;
	static s: int;
	m() {}
}
const f:Foo+ = new Foo();
// FIXME: "The static member s cannot be accessed from a non-static context."
f.s;
```

PR #1261, opened 2019-02-22
GH-1260: Validation does not respect dynamically accessed members that are also defined as static members

mmews-n4:
see #1260

### DIFF
--- a/plugins/org.eclipse.n4js/src/org/eclipse/n4js/scoping/members/AbstractMemberScope.java
+++ b/plugins/org.eclipse.n4js/src/org/eclipse/n4js/scoping/members/AbstractMemberScope.java
@@ -188,9 +188,18 @@ public abstract class AbstractMemberScope extends AbstractScope {
 			existingMember = findMember(nameAsString, !accessForWriteOperation, !staticAccess);
 		}
 		if (existingMember != null) {
-			return new WrongStaticAccessDescription(
-					EObjectDescription.create(existingMember.getName(), existingMember),
-					staticAccess);
+			boolean wrongStaticAccessDescription = true;
+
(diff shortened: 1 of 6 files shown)


plugins/org.eclipse.n4js/src/org/eclipse/n4js/scoping/members/AbstractMemberScope.java
```diff
@@ -188,9 +188,18 @@ protected IEObjectDescription getSingleLocalElementByName(QualifiedName name) {
...
 		}
 		if (existingMember != null) {
-			return new WrongStaticAccessDescription(
-					EObjectDescription.create(existingMember.getName(), existingMember),
-					staticAccess);
+			boolean wrongStaticAccessDescription = true;
+
+			if (this instanceof MemberScope) {
+				MemberScope thiz = (MemberScope) this;
+				wrongStaticAccessDescription = !thiz.type.isDynamizable();
```
mor-n4: `Type#isDynamizable()` will always return "true" for `TN4Classifier`s. You need `TypeRef#isDynamic()`, here.

In the following example you can see the problem:
```
class Foo3 {
//	s: string;
	static s: int;
	m() {}
}
const f3:Foo3/* no plus! */ = new Foo3();
// XPECT FIXME errors --> "The static member s cannot be accessed from a non-static context." at "s"
f3.s; // currently showing: "Couldn't resolve reference to IdentifiableElement 's'."
```

mmews-n4: Thanks, I changed it!

Issue #1262, opened 2019-02-25
Exception when opening same workspace after moving IDE executable to a new location


mor-n4:
To reproduce:

1. install N4JS IDE product open it and create a new workspace,
2. close IDE,
3. move IDE executable (i.e. the macOS App bundle) to a new location on disk,
4. start IDE from new location re-opening the previously created workspace.

Will throw something along the lines of:
```
java.lang.IllegalArgumentException: Cannot determine package name of file:/Users/mark-oliver.reiser/Home/Prog/Java/n4js-main/git-repo/n4js/plugins/org.eclipse.n4js.external.libraries/bin/runtime/n4js-runtime-v8/package.json: The nested location is not contained in the given workspace location file:/Users/m
	at org.eclipse.n4js.external.ExternalIndexSynchronizer.getPackageName(ExternalIndexSynchronizer.java:339)
	at org.eclipse.n4js.external.ExternalIndexSynchronizer.addToIndex(ExternalIndexSynchronizer.java:285)
	at org.eclipse.n4js.external.ExternalIndexSynchronizer.findNpmsInIndex(ExternalIndexSynchronizer.java:173)
	at org.eclipse.n4js.external.ExternalIndexSynchronizer.identifyChangeSet(ExternalIndexSynchronizer.java:227)
	at org.eclipse.n4js.external.ExternalIndexSynchronizer.checkAndClearIndex(ExternalIndexSynchronizer.java:195)
	at org.eclipse.n4js.ui.N4JSExternalLibraryStartup$CheckNodeModulesSyncOnRefresh$1.run(N4JSExternalLibraryStartup.java:119)
	at org.eclipse.core.internal.jobs.Worker.run(Worker.java:63)
```

PR #1263, opened 2019-02-25
GH-1262: Exception when opening same workspace after moving IDE executable to a new location

mor-n4:
See #1262.


mor-n4:
Thanks!
Builds green, merging ...

PR #1290, opened 2019-04-03
docs(faq): Mention TypeScript 3’s `unknown` type

ExE-Boss:
**TypeScript 3** added the `unknown` type, which behaves like **N4JS’s** `any` type.

See also https://github.com/Microsoft/TypeScript/issues/27265, which would add a compiler flag to **TypeScript** that would switch the implicit type from `any` to `unknown`.


jpilgrim:
@ExE-Boss Thank you very much for the contribution. And you are absolutely right.

jpilgrim:
@mor-n4 Would it be ok for you to merge this PR?

mor-n4:
Yes, good to mention this in the docs for completeness!

Issue #1293, opened 2019-04-10
As a smith I want an experimental LSP prototype


mmews-n4:
As a smith I I want an experimental LSP prototype. This prototype is not intended to be merged to master. We want to explore LSP and how Xtext/Eclipse support LSP, i.e. what features are available out of the box and which features are to be added/adapted manually. The goal is to be able to estimate the effort for supporting and using LSP.

-  preparation: activate support and find out problems which could be solved by SZ
-  activate via flag
-  builder: need a new builder:
-  xtext.ide based builder for LSP
-  ensure no Eclipse specific things are used in transpiler
-  Existing: Eclipse builder, Headless builder, External builder
-  write UI toolkit independent IDE services (xtext.ide)
-  use bridges to adapt Eclipse UI and LSP
-  Generic implementation for use in Eclipse native and LSP
-  Effort to migrate existing quickfixes, etc.

**Notes:**
- VSCode does not support "Clear Error Markers"
- VSCode hangs sometimes when N4JS-Language-Server is closed (maybe due to 146K errors)

PR #1294, opened 2019-04-10
GH-1293: As a smith I want an experimental LSP prototype

mmews-n4:
see #1293


mmews-n4:
continued here: #1298

Issue #1303, opened 2019-04-30
As a user I want to use a yarn workspace by importing only some of its contained packages into the IDE


mor-n4:
Yarn workspaces are supported by the N4JS IDE already, as long as all packages contained inside the yarn workspace are imported into Eclipse. The goal of this user story is to support importing only some of the packages contained in a yarn workspace into Eclipse, even though they might have dependencies to one or more other packages contained in the yarn workspace. Those other packages should be handled as external libraries by the IDE.

ACs:
1. add support for the above.
2. must work with scoped packages (e.g. `@myScope/Lib`).
3. packages that are imported into Eclipse must work as before. In particular, when the Eclipse project of an imported package is closed, this package should *NOT* be available as an external library (instead compile errors should show up).
4. add tests.

PR #1304, opened 2019-04-30
GH-1303: As a user I want to use a yarn workspace by importing only some of its contained packages into the IDE

mor-n4:
See #1303.


mor-n4:
builds were green, merging ...

Issue #1299, opened 2019-04-12
As a user I want array syntax similar to Typescript


dbo:
N4JS and TypeScript have slightly different syntax for similar semantics. This issue shows an example of such a difference.

N4JS supports the following syntax for arrays: `[string]` equivalent to `Array<string>`
whereas typescript uses `string[]`.

What's even worse about this is that it has similarities with typescript's tuple syntax: `[string, number]` is equivalent to N4JS' `Iterable2<string, number>`.

jpilgrim:
Note: [xxx] is the syntax for tuples in TypeScript

PR #1321, opened 2019-05-21
GH-1299: As a user I want array syntax similar to Typescript (part 1)

mor-n4:
See #1299.

1. changes array short-hand syntax from `[string]` to `string[]`.
   I.e. now `let a: string[][]` is equivalent to `let a: Array<Array<string>>`.
2. changes `[string]` to become the short-hand syntax for `IterableN`.
   E.g. `let i: [string,number]` is equivalent to `let i: Iterable2<string,number>`.

Overall, this aligns the syntax to TypeScript. However, `IterableN` is very different from TypeScript's tuples; these semantics are not changed by this PR, only the short-hand syntax.

**This is only part 1:** for backward compatibility `[T1, ..., Tn]` is still treated as `Array<T1>` for n=1  (for n=0 and n>1 it is already treated as `IterableN<T1, ..., Tn>`) and a quick fix is provided to transform `[T]` to `T[]`.


mor-n4:
Builds green.

Issue #1312, opened 2019-05-10
As a developer I want to have access to the propery 'name' of functions


mmews-n4:
Add a getter to provide access to the name of a function. Proposed solution below.

```js
@@Global @@ProvidedByRuntime

@Polyfill export external public class Function extends Function {
    public get name(): string;
}
```

dbo:
Requires the builtin `Function` to be non-final.

mor-n4:
See also:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/name

mor-n4:
> Requires the builtin `Function` to be non-final.

As an alternative, we should consider allowing polyfills for final classes in general. Polyfilling is different from sub-classing (just using a similar syntax).

mor-n4:
After looking into this a bit:

Polyfilling final classes had already worked for runtime polyfills. For static polyfills it did not work yet and I have fixed that on this branch. Also I added tests for both static and runtime polyfills of final classes.

PR #1322, opened 2019-05-22
GH-1312: As a developer I want to have access to the propery 'name' of functions

dbo:
#1312

@mor-n4 please; could be merged once the compiler either allows polyfills on final classes or the builtin final is removed.


mor-n4:
@dbo Actually this had already worked for your case, you just forgot the `@Final` on the polyfill class.

For static polyfills it did not work yet and I have fixed that on this branch, but you had a runtime polyfill. Also I added tests for both static and runtime polyfills of final classes.

Issue #1307, opened 2019-05-09
As a developer I want to bundle the IDE with a JRE


mmews-n4:
Acceptance Criteria:
* Bundle the IDE (not n4jsc.jar) with JRE-11
* Strip JRE-11 to a set necessary modules

jpilgrim:
also see https://bugs.eclipse.org/bugs/show_bug.cgi?id=506244

mmews-n4:
There are two options of bundling the JRE with the IDE (see: https://codeiseasy.wordpress.com/2012/07/31/including-a-jre-in-a-tycho-build/):
(1) just copy the JRE into the product and set the path in the eclipse.ini, or
(2) use the feature mechanism to bundle the JRE.

Notes:
- Drawback of (1) is that under Windows, the update function of the IDE will not update the bundled JRE (see: https://dev.eclipse.org/mhonarc/lists/tycho-user/msg04632.html)
- Option (2) makes it necessary to set file permissions (using `chmod`) for the JRE in Linux and MacOS
  - Using `chmod` does not work when Oomphing the N4JS-Eclipse-IDE (see: https://www.eclipse.org/forums/index.php/t/1074080/).

PR #1325, opened 2019-05-27
GH-1307: As a developer I want to bundle the IDE with a JRE

mmews-n4:
see #1307


mmews-n4:
The eclipsefdn/eca check currently fails on commiter @qtran-n4. This is not a problem since the last commit of qtran was checked fine without any issues. Only now and due to some changes in his e-mail address, this check fails. However, we merge this branch still, since as shown with the green check mark indicators above, all commits of qtran have been checked fine once.

Issue #1326, opened 2019-05-27
As a user I want same module specifiers in output code as in the N4JS source (as far as possible)


mor-n4:
Currently, all module specifiers that appear in imports in the Javascript output code are created by the transpiler from scratch, even in cases when the module specifier given in the N4JS source code could simply be reused). The reason for this is that the transpiler does not have enough information to tell if the module specifier from the source is of a format that can simply be reused in the output code or a format that is supported only in N4JS.

Hence, this user story requires some refactoring of scoping, where imports are resolved, to store the module specifier format somewhere and let the transpiler use this information.

Main uses cases in which module specifiers do need adjustment in the transpiler (also in the future):
1. import from another project without specifying the project name (only supported in N4JS),
2. import from same project (in N4JS the module specifier is always absolute, so transpiler has to change it to a relative module specifier in this case).

In particular, project imports (aka bare imports) should never be changed by the transpiler.

### ACs
1. adjust scoping to store information about format of module specifier somewhere,
2. adjust transpiler to use module specifier from source code by default and only change it if necessary,
3. add tests,
4. adjust documentation (if the adjusted behavior is documented, at the moment).

PR #1332, opened 2019-06-03
GH-1326: As a user I want same module specifiers in output code as in the N4JS source (as far as possible)

mor-n4:
See #1326.


mor-n4:
In the end, the changes were purely internal and the observable behavior did not change, except for a corner case. So no real spec update required. However, I added a clarification about main modules of definition projects.

PR #1333, opened 2019-06-03
GH-1299: As a user I want array syntax similar to Typescript (part 2)

mor-n4:
See #1299.

Part 2 of #1321 that will
1. change the meaning of the old short-hand syntax `[string]` from `Array<string>` to `Iterable<string>`.
2. remove the corresponding warning and quick fix for converting from old to new syntax.

This PR breaks backward compatibility!


mor-n4:
NOTE: most commits are from the part 1 PR. New commits start with fe90826.

Issue #1355, opened 2019-06-19
Scoping of to polyfillee and polyfilled types not consistent


mmews-n4:
**ABC.n4js in source folder src1**
```ts
@@StaticPolyfillAware
export public class ABC {
	method1() {
		let xy: ABC; // OK: CMD+Click jumps to src1
		(xy instanceof ABC); // OK: CMD+Click jumps to src1
	}
}
```

**ABC.n4js in source folder src2**
```ts
@@StaticPolyfillModule

@StaticPolyfill
export public class ABC extends ABC {
	method2() {
		let xy: ABC; // OK: CMD+Click jumps to src1
		(xy instanceof ABC); // FAIL: CMD+Click jumps to src2

		let m1: Module;
		let m2 = new Module();
		m1 = m2;
		m2 = m1; // XPECT no errors
	}
}
```

This is especially a problem when type guards are used, since the inferred type then relies on the type defined in the instanceof expression.

PR #1358, opened 2019-06-21
GH-1355: Scoping of to polyfillee and polyfilled types not consistent

mmews-n4:
see #1355

### DIFF
--- a/plugins/org.eclipse.n4js/src/org/eclipse/n4js/n4JS/extensions/SourceElementExtensions.xtend
+++ b/plugins/org.eclipse.n4js/src/org/eclipse/n4js/n4JS/extensions/SourceElementExtensions.xtend
@@ -139,9 +139,16 @@ class SourceElementExtensions {
 			val next = allContents.next
 			switch (next) {
 				N4ClassDeclaration: {
-					val polyfilledOrOriginalType = next.getTypeOrPolyfilledType() as TClass;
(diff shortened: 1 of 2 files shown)


plugins/org.eclipse.n4js/src/org/eclipse/n4js/n4JS/extensions/SourceElementExtensions.xtend
```diff
@@ -124,7 +139,17 @@ class SourceElementExtensions {
...
+					val nonNullClassDecl = if (polyfilledOrOriginalType instanceof TClass) {
+						val n4ClassDecl = polyfilledOrOriginalType?.astElement as N4ClassDeclaration;
+						if (n4ClassDecl === null)
+							next
+						else
+							n4ClassDecl;
+					} else
+						next;
+
+					nonNullClassDecl.collectVisibleTypedElement(addHere)
```
mor-n4: Could this be required for interfaces as well? (i.e. following case within this switch)

mmews-n4: Thanks for the hint. For the record: We looked into this and annotations of `@Polyfill` or `@StaticPolyfill` cannot be used for interfaces.

Issue #1359, opened 2019-06-24
As a smith I need a new version of Xpect fixing the Java 9 system class loader issue


mor-n4:
Xpect has a problem when running under a Java 9+ JRE: Xpect tests in `.xt` test files cannot be launched locally via the JUnit runner(s). See https://github.com/eclipse/Xpect/issues/264.

### ACs
1. Implement a fix,
2. submit a pull request for Xpect,
3. update the N4JS build to use the new version.

PR #1360, opened 2019-06-24
GH-1359: As a smith I need a new version of Xpect fixing the Java 9 system class loader issue

mor-n4:
See #1359.


mor-n4:
Thanks! Merging ...

Issue #1363, opened 2019-06-27
Unresolved reference error on ParameterizedTypeRefs when accessing Namespaces on dynamic imports


mmews-n4:
The following will create an issue which shows up during transpilation.

**file Module.n4js**
```ts
import * as SomeNPM+ from 'some-npm';
// XPECT fixme noerrors --> "unable to transpile file Module.n4js due to an unresolved reference in line 3 at column 20"
let clientOptions: SomeNPM.MyClass;
```

PR #1365, opened 2019-06-27
GH-1363: Unresolved reference error on ParameterizedTypeRefs when accessing Namespaces on dynamic imports

mmews-n4:
see #1363


mmews-n4:
Thanks!

Issue #1375, opened 2019-07-08
Test in N4JSResourceTest failing when run locally


mor-n4:
The test `N4JSResourceTest#testDisposeDocumentDoesntTriggerResolution()` is failing when executed locally in the Eclipse IDE but is working fine on maven/Jenkins.

Also, the test shows different failure behavior when executed as a plugin test.

PR #1376, opened 2019-07-08
GH-1375: Test in N4JSResourceTest failing when run locally

mor-n4:
See #1375.


mor-n4:
Builds green, merging ...

PR #1383, opened 2019-07-10
Typesafe wrappers for URIs and project names

szarnekow:


plugins/org.eclipse.n4js.generator.headless/src/org/eclipse/n4js/generator/headless/ConfiguredGeneratorFactory.java
```diff
@@ -65,7 +65,7 @@ ConfiguredGenerator getConfiguredGenerator(IN4JSProject project) {
 	 */
 	private void configureFSA(IN4JSProject project) {
 		File currentDirectory = new File(".");
-		File projectLocation = new File(project.getLocation().toFileString());
+		File projectLocation = new File(project._getLocation().toFileString());
```
szarnekow: `_getLocation` is a temporary name for this method. Same as `getSafeLocation`.
The former should be removed whereas the `getSafeLocation` should be renamed to `getLocation`

plugins/org.eclipse.n4js.generator.headless/src/org/eclipse/n4js/generator/headless/HeadlessHelper.java
```diff
@@ -68,7 +67,8 @@
 	 *             in error Case.
 	 */
 	public void registerProjects(BuildSet buildSet, FileBasedWorkspace workspace) throws N4JSCompileException {
-		Iterable<URI> projectUris = Iterables.transform(buildSet.getAllProjects(), p -> p.getLocation());
+		Iterable<FileURI> projectUris = Iterables.transform(buildSet.getAllProjects(),
+				p -> (FileURI) p.getSafeLocation());
```
szarnekow: The cast to `FileURI` looks suspicious at a first glance. Needs to be double checked.

plugins/org.eclipse.n4js.ui/src/org/eclipse/n4js/ui/internal/N4JSEclipseModel.java
```diff
@@ -171,32 +197,37 @@ public N4JSEclipseProject findProjectWith(URI nestedLocation) {
...
 		if (null != project && project.exists()) {
-			final Path path = new Path(location.toPlatformString(true));
-			final IResource resource;
-			if (1 == path.segmentCount()) {
-				resource = workspace.getProject(path.segment(0));
-			} else {
-				resource = workspace.getFile(path);
-			}
-			n4jsContainer = getN4JSSourceContainer(resource);
+			n4jsContainer = Optional.fromNullable(project.findSourceFolderContaining(location));
```
mmews-n4: Seems like code duplication with above if-condition

plugins/org.eclipse.n4js.ui/src/org/eclipse/n4js/ui/external/ExternalProjectProvider.java
```diff
@@ -49,7 +49,7 @@
 	private ExternalLibraryPreferenceStore externalLibraryPreferenceStore;
 
 	@Inject
-	private EclipseBasedN4JSWorkspace userWorkspace;
+	private IN4JSCore userWorkspace;
```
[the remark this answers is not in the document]
szarnekow: Good catch. I'll fix this.

plugins/org.eclipse.n4js/src/org/eclipse/n4js/projectModel/locations/SafeURI.java
```diff
@@ -0,0 +1,165 @@
...
+
+	public boolean isParent(U nestedLocation) {
+		return nestedLocation.toFileSystemPath().startsWith(toFileSystemPath());
+	}
+
+	public abstract void delete(Consumer<? super IOException> errorHandler);
+
+	public abstract Path toFileSystemPath();
+
+	protected abstract U self();
```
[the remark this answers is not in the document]
szarnekow: The very short answer: Java does not have self types and generics can be used to mimic those.

We want to expose methods on the SafeURI, that will return the same type when they are called, e.g. `public U withTrailingPathDelimiter()` - such that a `PlatformResourceURI.withTrailingPathDelimiter` guarantees a result of type `PlatformResourceURI`. Without `self()`, we'd have to do an unchecked cast to `U` whenever we would want to return `this`.

plugins/org.eclipse.n4js.ui/src/org/eclipse/n4js/ui/external/ExternalProjectMappings.java
```diff
@@ -286,33 +270,36 @@ private void computeNecessaryDependenciesRek(Map<String, List<N4JSExternalProjec
...
-		URI depLoc = userWorkspace.findProjectForName(projectName);
-		if (depLoc != null) {
-			// respect closed workspace projects by omitting them
-			String locStr = depLoc.toPlatformString(true);
-			IProject iProject = ResourcesPlugin.getWorkspace().getRoot().getProject(locStr);
-			if (iProject == null || !iProject.isAccessible()) {
-				depLoc = null;
-			}
-		}
+		String workspaceProjectName = ProjectDescriptionUtils.convertN4JSProjectNameToEclipseProjectName(projectName);
```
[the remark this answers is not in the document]
szarnekow: I'd prefer that, but I'm not sure which name has to be put where. Ideally on the IProject has the eclipse name and all other names are of the shape `@scope/name` but I don't know how to guarantee that.

plugins/org.eclipse.n4js/src/org/eclipse/n4js/projectModel/locations/SafeURI.java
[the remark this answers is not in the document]
mmews-n4: Yes. When `self()` returns `(U) this;`, no subclass would have to implement `self()`. Acknowledged that this is a (somewhat) unchecked cast, at least from the IDE-point-of-view.

plugins/org.eclipse.n4js/src/org/eclipse/n4js/naming/ModuleNameComputer.xtend
```diff
@@ -106,10 +107,8 @@ class ModuleNameComputer {
 	}
 
 	private def boolean uriStartsWith(URI resourceLocation, URI containerLocation) {
```
mmews-n4: Similar semantics: `UriUtil#isPrefixOf(URI, URI)`

plugins/org.eclipse.n4js/src/org/eclipse/n4js/projectModel/names/N4JSProjectName.java
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * A project name of the shape `(@scope/)?name`.
+ */
+public class N4JSProjectName implements Comparable<N4JSProjectName> {
+
+	private final String name;
+
+	public N4JSProjectName(String name) {
+		this.name = Preconditions.checkNotNull(name);
```
mmews-n4: While `EclipseProjectName` can be null, `N4JSProjectName` cannot?

plugins/org.eclipse.n4js/src/org/eclipse/n4js/projectModel/names/N4JSProjectName.java
```diff
@@ -0,0 +1,84 @@
...
+		Preconditions.checkArgument(name.indexOf(':') < 0, name);
```
mmews-n4: A similar check in `EclipseProjectName` could be introduced:
`Preconditions.checkArgument(name.indexOf('/') < 0, name);`

Issue #1373, opened 2019-07-05
As a smith I do not want to have compile errors after oomphing


jpilgrim:
After freshely "oomphing" N4JS with latest Eclipse version (Eclipse IDE for Eclipse Committers, 2019-06) and I got two problems:

- [ ] [org.eclipse.n4js.model/emf-gen](https://github.com/eclipse/n4js/tree/master/plugins/org.eclipse.n4js.model/emf-gen)
The whole folder contains errors due to wrong code generation. How can we fix that? Do we need a different EMF version somewhere?
- [ ] [AstSelectionProvider2PluginUITest.xtend](https://github.com/eclipse/n4js/blob/master/tests/org.eclipse.n4js.ui.tests/src/org/eclipse/n4js/tests/selection/AstSelectionProvider2PluginUITest.xtend) needs to be "touched" (i.e. changed, saved, undo change, save) in order to work

jpilgrim:
Screenshot of the two workaround:

1) emf-gen:
<img width="721" alt="ideWorkaround1" src="https://user-images.githubusercontent.com/1684899/60759659-2e283100-a029-11e9-90f6-262345fac2fe.png">

2) AstSelectionProvider2PluginUITest.xtend
<img width="542" alt="ideWorkaround2" src="https://user-images.githubusercontent.com/1684899/60759656-236d9c00-a029-11e9-8313-d652d2a1ea8f.png">

mmews-n4:
https://github.com/eclipse/n4js/pull/1390 fixes first item

szarnekow:
The other problem appears to be caused by a wrongly computed build order. Investigating.

szarnekow:
Further analysis revealed the following causal chain for the observed problem:

- We use bundle fragment to implement tests, lets say `n4js.runner.ui.tests` is a fragment for `n4js.runner.ui`
- The fragments required bundles, e.g. `n4js.ui.tests.helper` are hoisted into the host bundles dependencies internally by PDE
- Now we have a dependency from `n4js.runner.ui` to `n4js.ui.tests.helper` that is considered by the Eclipse incremental project builder.
- `n4jsui.tests.helper` depends again on `n4js.runner.ui` which forms a cyclic dependency.
- Cyclic dependencies are evil and cause the problem with the bogus error markers.
- Potential workaround: Refactor test fragments to bundles.

szarnekow:
see also https://bugs.eclipse.org/bugs/show_bug.cgi?id=536592
[...]

PR #1390, opened 2019-07-18
[#1373] More explicit type information as a workaround for Xcore bug

szarnekow:
[#1373] More explicit type information as a workaround for Xcore bug


mmews-n4:
Thanks!

mmews-n4:
builds are green, merging.

Issue #1388, opened 2019-07-16
As a smith I want to update to Eclipse 2019-06


mor-n4:
For the N4JS IDE, bump version of Eclipse to 2019-06 and also update all other dependencies (where applicable).

PR #1399, opened 2019-07-24
GH-1388: As a smith I want to update to Eclipse 2019-06 (part 2)

mor-n4:
See #1388.

This PR replaces `org.eclipse.equinox.ds` by `org.apache.felix.scr`, because `ds` is no longer part of the equinox feature, see [1], and would thus require explicit inclusion in the target platform. However, Oomph always removed this dependency to `ds` when re-generating the target platform, even when explicitly defining it as a dependency in a manifest.

Unrelated to the above, this PR also fixes a problem with dependency `javax.xml.stream`: Oomph always removed that dependency from the target platform when re-generating it, even though the dependency is required.


releng/org.eclipse.n4js.targetplatform/org.eclipse.n4js.targetplatform.target
```diff
@@ -75,7 +74,7 @@
       <unit id="com.google.inject" version="3.0.0.v201605172100"/>
       <unit id="io.github.classgraph" version="4.8.35.v20190528-1517"/>
       <unit id="javax.xml.bind" version="2.2.0.v201105210648"/>
-      <unit id="javax.xml.stream" version="1.0.1.v201004272200"/>
+      <unit id="javax.xml.stream" version="1.0.1.v201004272200"/> <!-- Oomph want's to remove javax.xml.stream, but it's required by rest-assured-2.4.0.jar in tester.tests.helper (tested July 2019) -->
```
torstenkraemer: try again with the extraUnits Annotation in the targlet-generator and use "," separation and only one entry.....

mor-n4:
This works! Seems like a bug in Oomph because the XML schema allows for several `extraUnits` elements, which I tested. Will look into this and maybe issue a ticket.

In any case: great idea to try this! :+1:

torstenkraemer:
As discussed the Annotations in the Setup-files are super-generic and I doubt that there will be any formal restriction on what you can enter.

PR #1401, opened 2019-07-24
GH-1400: Only add project if package.json has an "n4js" section

dbo:
#1400 

@mor-n4 please


mor-n4:
Builds green, merging ...

Issue #1422, opened 2019-08-21
npm install fails while installing n4js-cli on Windows 10


halitogunc:
n4js-cli installation failed at the **n4js-cli** `postinstall` script.

```
Windows Script Host
Line: 1
Char: 1
Error: Invalid character
Code: 800A03F6
Source: Microsoft JScript compilation error 
```

PR #1423, opened 2019-08-21
Fix npm install failure on Windows

halitogunc:
see: #1422

cc @mor-n4

### DIFF
--- a/n4js-libs/packages/n4js-cli/package.json
+++ b/n4js-libs/packages/n4js-cli/package.json
@@ -7,7 +7,7 @@
   "internalVersion": "2019-03-15T08:46:12Z",
   "description": "N4JS compiler and command line tools.",
   "scripts": {
-    "postinstall": "./npm-postinstall.js",
+    "postinstall": "node ./npm-postinstall.js",
     "build": "./npm-build.sh",
     "test": "eslint . && mocha tests"
   },
--- a/n4js-libs/packages/n4js-cli/npm-postinstall.js
+++ b/n4js-libs/packages/n4js-cli/npm-postinstall.js
@@ -1,5 +1,3 @@
-#!/usr/bin/env node
-
 /*
  * Copyright (c) 2018 NumberFour AG.
  * All rights reserved. This program and the accompanying materials


n4js-libs/packages/n4js-cli/package.json
```diff
@@ -7,7 +7,7 @@
   "internalVersion": "2019-03-15T08:46:12Z",
   "description": "N4JS compiler and command line tools.",
   "scripts": {
-    "postinstall": "./npm-postinstall.js",
+    "postinstall": "node ./npm-postinstall.js",
     "build": "./npm-build.sh",
```
mor-n4: And what about this shell script `npm-build.sh`? Does it work on Windows?
Also `n4jsc.js` a few lines further down: doesn't that one need to be adjusted, as well?

n4js-libs/packages/n4js-cli/package.json
```diff
@@ -7,7 +7,7 @@
   "internalVersion": "2019-03-15T08:46:12Z",
   "description": "N4JS compiler and command line tools.",
   "scripts": {
-    "postinstall": "./npm-postinstall.js",
+    "postinstall": "node ./npm-postinstall.js",
```
mor-n4: Have you tested that it still runs on macOS?

n4js-libs/packages/n4js-cli/package.json
mor-n4: Have you checked whether other package.json files contain similar cases? (shell scripts, .js-files launched via shebang)

n4js-libs/packages/n4js-cli/package.json
halitogunc: yes, it still works on macOS

n4js-libs/packages/n4js-cli/package.json
halitogunc: Windows does not have a built-in support for .sh files. They need to install a third-party tool like Cygwin, Git bash or similar.

halitogunc: yes, I have checked it and could not find any

mor-n4:
Builds green, merging ...

Issue #1440, opened 2019-08-27
Xpect test files cannot be executed individually from inside development Eclipse


mor-n4:
Currently, Xpect test files cannot be run individually from inside the IDE. Maybe this problem occurs only in some bundles.

To reproduce, either right-click file `01_CannotInstantiateInterface.n4js.xt` in Package Explorer and select "Run As ... > JUnit Test" OR open the file and right-click in the editor and select same context menu item; will throw:
```
java.lang.RuntimeException: java.lang.reflect.InvocationTargetException
	at org.eclipse.xpect.state.StateContainer.createInstance(StateContainer.java:197)
	at org.eclipse.xpect.state.StateContainer.createInstance(StateContainer.java:209)
	at org.eclipse.xpect.state.StateContainer.get(StateContainer.java:259)
	at org.eclipse.xpect.state.StateContainer.get(StateContainer.java:221)
	at org.eclipse.xpect.state.StateContainer.createFactory(StateContainer.java:161)
	at org.eclipse.xpect.state.StateContainer.getFactory(StateContainer.java:282)
	at org.eclipse.xpect.state.StateContainer.createInstance(StateContainer.java:185)
	at org.eclipse.xpect.state.StateContainer.createInstance(StateContainer.java:209)
	at org.eclipse.xpect.state.StateContainer.get(StateContainer.java:259)
	at org.eclipse.xpect.state.StateContainer.get(StateContainer.java:221)
	at org.eclipse.xpect.runner.XpectFileRunner.run(XpectFileRunner.java:139)
	at org.eclipse.xpect.runner.XpectRunner.runChild(XpectRunner.java:212)
	at org.eclipse.xpect.runner.XpectRunner.runChild(XpectRunner.java:1)
	at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
	at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
	at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
	at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
	at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
	at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
	at org.eclipse.xpect.runner.XpectRunner.run(XpectRunner.java:198)
	at org.eclipse.jdt.internal.junit4.runner.JUnit4TestReference.run(JUnit4TestReference.java:89)
[...]

PR #1441, opened 2019-08-27
GH-1440: Xpect test files cannot be executed individually from inside development Eclipse

mmews-n4:
see #1440


mmews-n4:
@szarnekow Would you take a brief look at this PR? I made changes to `LspCompatibleXpectTestFiles` and I introduced a `N4JSXpectRunner`?

plugins/org.eclipse.n4js.xpect/src/org/eclipse/n4js/xpect/common/SafeURIFileCollector.java
```diff
@@ -0,0 +1,68 @@
...
+	private final UriExtensions uriExtensions = new UriExtensions();
+
+	/** Constructor */
+	public SafeURIFileCollector(Class<?> owner, XpectTestFiles ctx) {
+		super(owner, ctx);
+	}
+
+	@Override
+	protected URI createURI(File file) {
+		throw new UnsupportedOperationException("Never called by super impl");
```
mor-n4: @szarnekow Why not just create a file URI, but with empty authority (corresponding to other overrides).

szarnekow: Since it was never called as far as I could tell, I wanted to add a tripwire.

Issue #1443, opened 2019-08-29
On Windows some imports cannot be resolved


mmews-n4:
This issue happens only on the windows platform when compiling N4JS code which compiles fine on MacOS platform.

`@issue: ERROR:Cannot resolve import target :: resolving simple module import : found no matching modules (file:///C:/.../node_modules/n4js-runtime-ecma402/src/n4js/Date.n4jsd line : 13 column : 37)`

PR #1446, opened 2019-08-29
GH-1443: On Windows some imports cannot be resolved

mmews-n4:
see #1443

### DIFF
--- a/plugins/org.eclipse.n4js/src/org/eclipse/n4js/internal/N4JSProjectSourceContainer.java
+++ b/plugins/org.eclipse.n4js/src/org/eclipse/n4js/internal/N4JSProjectSourceContainer.java
@@ -45,7 +45,8 @@ public class N4JSProjectSourceContainer extends AbstractSourceContainer implemen
 	public SafeURI<?> getLocation() {
 		String location = getRelativeLocation();
 		if (!Strings.isEmpty(location)) {
-			return project.getLocation().appendPath(getRelativeLocation());
+			String linuxPath = location.replaceAll("\\\\", "/");
+			return project.getLocation().appendPath(linuxPath);
 		}
 		return project.getLocation();
 	}
--- a/plugins/org.eclipse.n4js/src/org/eclipse/n4js/internal/N4JSProjectSourceContainer.java
+++ b/plugins/org.eclipse.n4js/src/org/eclipse/n4js/internal/N4JSProjectSourceContainer.java
@@ -10,10 +10,13 @@
  */
 package org.eclipse.n4js.internal;
 
+import java.io.File;
+
 import org.eclipse.n4js.projectDescription.SourceContainerType;
 import org.eclipse.n4js.projectModel.IN4JSProject;
 import org.eclipse.n4js.projectModel.IN4JSSourceContainer;
 import org.eclipse.n4js.projectModel.locations.SafeURI;
+import org.eclipse.n4js.utils.OSInfo;
 import org.eclipse.xtext.naming.QualifiedName;
 import org.eclipse.xtext.util.Strings;
 
@@ -45,7 +48,7 @@ public class N4JSProjectSourceContainer extends AbstractSourceContainer implemen
 	public SafeURI<?> getLocation() {
 		String location = getRelativeLocation();
 		if (!Strings.isEmpty(location)) {
-			String linuxPath = location.replaceAll("\\\\", "/");
+			String linuxPath = OSInfo.isWindows() ? location.replace(File.separatorChar, '/') : location;
 			return project.getLocation().appendPath(linuxPath);
 		}
 		return project.getLocation();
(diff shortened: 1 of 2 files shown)


plugins/org.eclipse.n4js/src/org/eclipse/n4js/internal/N4JSProjectSourceContainer.java
```diff
@@ -45,7 +45,8 @@ public IN4JSProject getProject() {
 	public SafeURI<?> getLocation() {
 		String location = getRelativeLocation();
 		if (!Strings.isEmpty(location)) {
-			return project.getLocation().appendPath(getRelativeLocation());
+			String linuxPath = location.replaceAll("\\\\", "/");
```
mor-n4: Not sure it's always correct to replace backslash with slash under non-windows platforms.

So, assuming `#getRelativeLocation()` returns a file system path as a string, I'd do something like:
```java
String linuxPath = File.separatorChar != '/' ? location.replace(File.separator, "/") : location;
```
(note that I use `#replace()` instead of `#replaceAll()` to avoid the need for `Pattern#quote()`)

Ideally, if `File.separatorChar!='/'`, we would also have to escape existing forward slashes somehow before changing `File.separatorChar` to `'/'`.

plugins/org.eclipse.n4js/src/org/eclipse/n4js/projectModel/locations/SafeURI.java
```diff
@@ -60,15 +60,14 @@ protected SafeURI(URI wrapped) {
...
-		} else {
-			for (String segment : segments) {
-				Preconditions.checkArgument(segment.length() > 0, "'%s'", given);
-			}
+
+		final int segCountMax = given.hasTrailingPathSeparator() ? segments.size() - 1 : segments.size();
+		for (int i = 0; i < segCountMax; i++) {
+			String segment = segments.get(i);
+			Preconditions.checkArgument(segment.length() > 0, "'%s'", given);
+			Preconditions.checkArgument(!segment.contains("\\")); // could happen on Windows platform
```
mor-n4: Similar as above. Backslash might appear in a segment on some non-Windows platform.

mmews-n4: How would a backslash on non-windows (which would not be semantically a slash) translate to a windows path in some URI?

mmews-n4: In other words, you like to allow backslashes in URIs? (However, not on Windows.)

mor-n4: I would have to check how macOS or Linux behave, but in principle a back slash might be a valid character for folder names on some non-Windows platform. In a URI it would have to be escaped, though I'm not even sure that escaping of backslash is strictly required by the standard, but at least it is good practice.

mor-n4: I just checked: on macOS backslash is working fine in file/folder names when using the command line (also file name completion with TAB supports it by auto-inserting additional backslashes for escaping).

mmews-n4: I see, thanks!

szarnekow: I have the feeling that this discussion should be about encoded / decoded URIs instead. Would URI.encode / decode handle backslashes correctly?

mor-n4: I don't think so, because as far as I understand the code, `Preconditions.checkArgument(!segment.contains("\\"))` is imposing a rule about what characters an individual segment may contain, no matter if they are URI-encoded or not.

BUT: in case `URI#segmentsList()` does not guarantee to return decoded strings (I don't know if it does), then the current code will "overlook" invalid backslashes if they are encoded, which is an unrelated problem (but not critical, I guess).

mmews-n4: @szarnekow I am not sure how you mean that. Let me explain what I have in my mind.

My intention was to disallow backslashes in URIs on windows (and throw exceptions). The reason for this was that backslashes are forbidden as part of folder or file names on windows. However, as @mor-n4 pointed out, backslashes actually can be part of folder or file names on e.g. MacOS, and hence should be supported on these OSs.

To en-/decode backslashes on Windows would mean that there is a directory named for instance "direc%5Ctory". To create its URI, we use the string "direc\tory" and get after encoding its URI direc%5Ctory"?

szarnekow: > My intention was to disallow backslashes in URIs on windows (and throw exceptions). 

Ok, I understand. In that case, I like the last patchset. It looks good to me.

mmews-n4:
thanks!

Issue #1449, opened 2019-08-30
On windows case-insensitive environment variables names are not respected


mmews-n4:
This results in a problem when searching for binaries in the PATH variable and adjusting the PATH variable (or others) to add other paths.

PR #1450, opened 2019-08-30
GH-1449: On windows case-insensitive environment variables names are not respected 

mmews-n4:
see #1449

### DIFF
--- a/plugins/org.eclipse.n4js/src/org/eclipse/n4js/binaries/Binary.java
+++ b/plugins/org.eclipse.n4js/src/org/eclipse/n4js/binaries/Binary.java
@@ -154,12 +154,12 @@ abstract public class Binary {
 			return defaultName;
 		}
 		if (OSInfo.isWindows()) {
-			if (System.getenv(PATH) == null) {
+			if (System.getenv(defaultName) == null) {
 				return defaultName;
 			}
-			final String lSimilarName = defaultName.toLowerCase();
+
 			for (String prop : environment.keySet()) {
-				if (lSimilarName.equals(prop.toLowerCase())) {
+				if (defaultName.equalsIgnoreCase(prop)) {
 					return prop;
 				}
 			}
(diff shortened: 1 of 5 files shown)


mmews-n4:
@mor-n4 Note that the fix is in method `updateEnvironment(...)` only. The remaining changes are refactorings to reduce code duplications.

plugins/org.eclipse.n4js/src/org/eclipse/n4js/binaries/Binary.java
```diff
@@ -107,37 +137,67 @@ default String getDescription() {
...
+		if (environment.containsKey(defaultName)) {
+			return defaultName;
+		}
+		if (OSInfo.isWindows()) {
+			if (System.getenv(PATH) == null) {
+				return defaultName;
+			}
+			final String lSimilarName = defaultName.toLowerCase();
+			for (String prop : environment.keySet()) {
+				if (lSimilarName.equals(prop.toLowerCase())) {
```
mor-n4: very minor: you could use `String#equalsIgnoreCase()` instead of converting to lower case twice.

plugins/org.eclipse.n4js/src/org/eclipse/n4js/binaries/Binary.java
```diff
@@ -107,37 +137,67 @@ default String getDescription() {
...
+	}
+
+	/** Deals with case-insensitivity on environment variables on windows platform */
+	protected String findActualPropertyNameOrDefault(Map<String, String> environment, String defaultName) {
+		if (environment.containsKey(defaultName)) {
+			return defaultName;
+		}
+		if (OSInfo.isWindows()) {
+			if (System.getenv(PATH) == null) {
+				return defaultName;
```
mor-n4: I don't understand the rationale for this case. Why does the return value of such a general-purpose utility method depend on the specific environment variable "PATH"?

plugins/org.eclipse.n4js/src/org/eclipse/n4js/binaries/Binary.java
mmews-n4: True, although I think that this change decreases performance of this loop. Change it anyway for readability.

PR #1456, opened 2019-09-04
GH-1455: n4js-cli postinstall: remove wget dependency in favor of usi…

dbo:
#1455

@mmews-n4 please

### DIFF
--- a/n4js-libs/packages/n4js-cli/README.md
+++ b/n4js-libs/packages/n4js-cli/README.md
@@ -17,6 +17,8 @@ For more information about N4JS, please refer to the official [project home](htt
 When installing the package, you can replace the packaged `n4jsc.jar` using the environment variable `N4JSC_JAR` referencing
 either a filesystem path or a HTTP/HTTPS URL.
 
+See [download page](https://download.enfore.com/n4js/).
+
 ## License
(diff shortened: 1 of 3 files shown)


n4js-libs/packages/n4js-cli/README.md
```diff
@@ -17,6 +17,8 @@ For more information about N4JS, please refer to the official [project home](htt
 When installing the package, you can replace the packaged `n4jsc.jar` using the environment variable `N4JSC_JAR` referencing
 either a filesystem path or a HTTP/HTTPS URL.
 
+See [download page](https://download.enfore.com/n4js/).
```
mmews-n4: We might better link to the Eclipse download page. There we list both the Eclipse and the enfore download pages.

mmews-n4: The page would be: https://www.eclipse.org/n4js/downloads.html

dbo: but only the latter links directly to the jars, isn't it?

dbo: or is there an eclipse jar download page?

Issue #1453, opened 2019-09-03
Asynchronous Jetty Server shutdown causes problems on Eclipse Jenkins


mmews-n4:
On the Eclipse Jenkins nightly builds, we see constantly tests failing with the error below. Comparing two different builds, the test cases differ (and also the number of test cases), but the error stays the same.

```
2019-08-240-11:12:20.020-0400 -                      main - ERROR -                   JettyManager - (org.eclipse.n4js.tester.server.JettyManager$1.java:121) - Cache failed to start new server instance at PORT=9415
java.io.IOException: Failed to bind to 0.0.0.0/0.0.0.0:9415
	at org.eclipse.jetty.server.ServerConnector.openAcceptChannel(ServerConnector.java:346)
	at org.eclipse.jetty.server.ServerConnector.open(ServerConnector.java:308)
	at org.eclipse.jetty.server.AbstractNetworkConnector.doStart(AbstractNetworkConnector.java:80)
	at org.eclipse.jetty.server.ServerConnector.doStart(ServerConnector.java:236)
	at org.eclipse.jetty.util.component.AbstractLifeCycle.start(AbstractLifeCycle.java:68)
	at org.eclipse.jetty.server.Server.doStart(Server.java:396)
	at org.eclipse.jetty.util.component.AbstractLifeCycle.start(AbstractLifeCycle.java:68)
	at org.eclipse.n4js.tester.server.JettyManager$1.load(JettyManager.java:119)
	at org.eclipse.n4js.tester.server.JettyManager$1.load(JettyManager.java:1)
	at com.google.common.cache.LocalCache$LoadingValueReference.loadFuture(LocalCache.java:3716)
	at com.google.common.cache.LocalCache$Segment.loadSync(LocalCache.java:2424)
	at com.google.common.cache.LocalCache$Segment.lockedGetOrLoad(LocalCache.java:2298)
	at com.google.common.cache.LocalCache$Segment.get(LocalCache.java:2211)
	at com.google.common.cache.LocalCache.get(LocalCache.java:4154)
	at com.google.common.cache.LocalCache.getOrLoad(LocalCache.java:4158)
	at com.google.common.cache.LocalCache$LocalLoadingCache.get(LocalCache.java:5147)
	at org.eclipse.n4js.tester.server.JettyManager.startServer(JettyManager.java:174)
	at org.eclipse.n4js.tester.tests.resource.BaseResourcesTest.before(BaseResourcesTest.java:78)
[...]

PR #1461, opened 2019-09-06
GH-1453C: Asynchronous Jetty Server shutdown causes problems on Eclipse Jenkins

mmews-n4:
Follow-up of GH-1453, since the actual problem was not solved before.

Turned out that on Eclipse Jenkins infrastructure, some defaults of Maven are different compared to enfore infrastructure (latter uses Docker). Namely, the maven-surefire plugin executes tests in parallel which caused races on hardware port 9415. Note that the race can be reproduced locally by setting the fork-count value to higher than one and setting the reuse-forks to true.

See Jenkins results of Eclipse of Branch GH-1453B here: https://ci.eclipse.org/n4js/job/GH-1453B/: The last five runs did not suffer anymore from the "could not bind" problem. Note that the one yellow build has a different problem.

Branch GH-1453C was introduced, since GH-1453B also contained lots of off-topic changes for testing reasons.

### DIFF
--- a/plugins/org.eclipse.n4js.tester/src/org/eclipse/n4js/tester/server/JettyManager.java
+++ b/plugins/org.eclipse.n4js.tester/src/org/eclipse/n4js/tester/server/JettyManager.java
@@ -108,6 +108,7 @@ public class JettyManager implements HttpServerManager {
 			try {
 				server = new Server(configureThreadPool(port));
 				ServerConnector connector = new ServerConnector(server);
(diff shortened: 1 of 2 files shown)


mmews-n4:
Thanks. Actually, I didn't search for the reason, but I suspect that this is since we do not use Docker. Also the Eclipse infrastructure provides a different version of Maven and it maybe does some pre-configuring. 😕

Issue #1464, opened 2019-09-09
As a user I want to have a download link for n4jsc.jar on the Eclipse page


mmews-n4:
The download for `n4jsc.jar` is missing on https://projects.eclipse.org/projects/technology.n4js/downloads. Add it.

PR #1465, opened 2019-09-09
GH-1464: As a user I want to have a download link for n4jsc.jar on the Eclipse page

mmews-n4:
see #1464


mmews-n4:
See updated Eclipse-Download page for new link to download `n4jsc.jar`.

Issue #1472, opened 2019-09-13
Test CheckDefaultPreferencesPluginUITest fails on some Jenkins builds


mmews-n4:
It fails often on Eclipse Jenkins nightly builds, and on some internal Jenkins builds. However, there are some Jenkins builds, where it never fails.

```
Running org.eclipse.n4js.tests.defaultPreferences.CheckDefaultPreferencesPluginUITest
[2019-09-12T11:32:53.033Z] Tests run: 1, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 0.301 s <<< FAILURE! - in org.eclipse.n4js.tests.defaultPreferences.CheckDefaultPreferencesPluginUITest
[2019-09-12T11:32:53.033Z] testDefaultPreferences(org.eclipse.n4js.tests.defaultPreferences.CheckDefaultPreferencesPluginUITest)  Time elapsed: 0.3 s  <<< FAILURE!
[2019-09-12T11:32:53.033Z] java.lang.AssertionError
[2019-09-12T11:32:53.033Z] 	at org.eclipse.n4js.tests.defaultPreferences.CheckDefaultPreferencesPluginUITest.testDefaultPreferences(CheckDefaultPreferencesPluginUITest.java:25)
```

mmews-n4:
seems to work now. closing.

PR #1473, opened 2019-09-13
GH-1472: Test CheckDefaultPreferencesPluginUITest fails on some Jenkins builds

mmews-n4:
see #1472


mmews-n4:
As discussed, the definition of the test execution order will be merged for now to evaluate whether some random tests will become reproducible. Later, this definition will be reverted again.

Issue #1482, opened 2019-09-19
As a smith I want to check if we can remove tycho-source-feature-plugin config from POMs


mor-n4:
Check purpose of maven plugin `tycho-source-feature-plugin` and decide if it can be removed from the `pom.xml` files.

### ACs

1. check purpose of maven plugin `tycho-source-feature-plugin`,
2. decide if it is still needed or obsolete,
3. in the latter case, remove from all `pom.xml` files and remove related things.

PR #1483, opened 2019-09-19
GH-1482: As a smith I want to check if we can remove tycho-source-feature-plugin config from POMs

mor-n4:
See #1482.


mor-n4:
Builds green.

Issue #1480, opened 2019-09-19
As a smith I want to update the N4JS IDE to Eclipse 2019-09


mmews-n4:
Due to the memory leak bug (#1420) in Eclipse 2019-06 this update is a long awaited change.

PR #1486, opened 2019-09-23
GH-1480 Part2: As a smith I want to update the N4JS IDE to Eclipse 2019-09

mmews-n4:
see #1480 

AC:
- Fix new warnings due to new libraries
- adjust target platform file to not change with every new run of setup tasks


mmews-n4:
AFAIK, we do not specify a version of guava in our sources anywhere. At the moment, we use version 27.1.0, which is also used by Xtext 2.19 and also in the [Eclipse Orbit](https://download.eclipse.org/tools/orbit/downloads/drops/R20190827152740/).

mmews-n4:
Thanks! (builds are green, merging)

Issue #1489, opened 2019-09-25
n4jsRuntimeLink may not be a symbolic link on Windows


jpilgrim:
Even though I have run BuildN4jsLibs.mwe2 the Xpect output tests do not work. E.g. I get 119 failures in the n4js.spec.tests, that is all output tests.

The failure is caused in org.eclipse.n4js.test.helper.hlc.N4jsLibsAccess.assertN4jsLibsAreBuilt(Path):

```
 Assert.assertTrue("not a symbolic link pointing to package \"" + N4JSGlobals.N4JS_RUNTIME + "\": "
		+ n4jsRuntimeLink + warning, Files.isSymbolicLink(n4jsRuntimeLink));
```
fails:

```
java.lang.AssertionError: not a symbolic link pointing to package "n4js-runtime": N:\oomph\n4js-master\git\n4js\n4js-libs\node_modules\n4js-runtime
******************************************************************
Maybe you forgot to run MWE2 workflow BuildN4jsLibs?
******************************************************************
```
I added a new assertion and uncommented the assertion above as follows:

```
Assert.assertTrue("path package does not exist \"" + N4JSGlobals.N4JS_RUNTIME + "\": "
				+ n4jsRuntimeLink + warning, Files.exists(n4jsRuntimeLink));
```

Now my tests are all working!

The assertion was added with #1281 by @mor-n4... Could you give me a hint what is going on?

Some background about my set up: It is Windows 10 with n4js installed on an external drive. Since the drive may be mounted with different drive names, I use a `subst n: d:\` to have a fixed drive name (i.e. n:).

Do we actually need a symbolic link there? Because in my case, the folder is there, but it is not a symbolic link.

mor-n4:
This symbolic link is created by yarn, so maybe yarn is working differently on Windows (creating copies instead of symbolic links). I'll have a look.

jpilgrim:
Update: as a matter of fact, `\n4js\n4js-libs\node_modules\n4js-runtime `is a symbolic link:

```
 <JUNCTION>     n4js-runtime [N:\oomph\n4js-master\git\n4js\n4js-libs\packages\n4js-runtime]
```

It seems something is wrong with the `Files.isSymbolicLink` (or how we use it..):

```
try {
			Path real = n4jsRuntimeLink.toRealPath();
			System.out.println("Link path: " + n4jsRuntimeLink);
			System.out.println("Real path: " + real);
[...]

PR #1490, opened 2019-09-25
GH-1489: n4jsRuntimeLink may not be a symbolic link on Windows

mor-n4:
See #1489.


jpilgrim:
See my comments at #1489

mor-n4:
@jpilgrim I'm now simply using a weaker assertion, as you have proposed. In the end, the stronger assertion not worth the hassle as it just implements a warning for the developer in case workflow `BuildN4jsLibs.mwe2` wasn't executed.

Merging now, please close the issue if it is now working for you.

Issue #1497, opened 2019-09-30
As a user I would like the transpiler to optimize trivial cases of reflection usage


mor-n4:
The transpiler should optimize some simple, reflection-related constructs:
```
ClassName.name
ClassName.n4type.name
ClassName.n4type.origin
ClassName.n4type.fqn
```
Instead of emitting the code as is and thus relying on reflection in the output code, the transpiler should directly emit the resulting string literal and thus keep the output code free of any reflection usage.

In these cases, related imports of the types (type `ClassName` in the example above) should be entirely removed from the output code if they are not required for other things in the module.

### ACs

1. adjust transpiler,
2. ensure now unused imports are removed,
3. create tests, including tests for cases in which the import has to be removed and cases in which it must not be removed.

PR #1499, opened 2019-10-01
GH-1497: As a user I would like the transpiler to optimize trivial cases of reflection usage

mor-n4:
See #1497.


mor-n4:
Builds green, merging ...

Issue #1518, opened 2019-10-18
As a smith, I want to be able to easily customize the LSP server


szarnekow:
The forked implementation as of #1515 has only a few solutions to be customized without relying on reflection or a lot of copy pasta. Thus we want to refactor the code to clean it up and adjust the design to allow for easier adjustments specific to n4js.

PR #1524, opened 2019-10-23
GH-1518: As a smith, I want to be able to easily customize the LSP server

mmews-n4:
see #1518


plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/build/XIncrementalBuilder.java
```diff
@@ -279,7 +279,7 @@ protected void unloadResource(URI uri) {
 						operationCanceledManager.checkCanceled(cancelIndicator);
 						// trigger init
 						resource.getContents();
-						EcoreUtil2.resolveLazyCrossReferences(resource, cancelIndicator);
+						EcoreUtil2.resolveLazyCrossReferences(resource, CancelIndicator.NullImpl);
```
mor-n4: Normally, resolving the cross-references can be a (very) long-running operation. Why is it ok to ignore the cancel indicator here by passing in `NullImpl`?

(similarly in line 293 below)

mor-n4: Thanks for the quick response!

In that case it would be good to have a TODO in this line to not forget to change it back once the LSP CancelIndicator is fixed. Or we could add a comment to some task to check all references to `CancelIndicator.NullImpl`.

Issue #1494, opened 2019-09-30
As a user I want support for optional chaining (aka elvis operator, etc.)


mor-n4:
There exists a Javascript proposal for null-safe/undef-safe property access, index access, and function invocation, called "optional chaining". This proposal is now in stage 3. Similarly, a closely related proposal for an "elvis operator" exists, called "nullish coalescing".

See:
https://github.com/tc39/proposal-optional-chaining
https://github.com/tc39/proposal-nullish-coalescing
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

There is also a Babel plugin available:
https://babeljs.io/docs/en/babel-plugin-proposal-optional-chaining

### ACs

1. implement syntax,
2. implement transpilation to ES5/6,
3. adjust type system and null/undef-analysis,
4. add tests,
5. adjust specification & documentation.

### Proposed transpilation

```
var foo = object.foo ?? "default";
```
should become
```
var _uniqueVarName;
var foo = (_uniqueVarName = object.foo) !== null && _uniqueVarName !== void 0 ? _uniqueVarName : "default";
```
---
```
a?.b.c
```
should become
```
var _uniqueVarName;

(_uniqueVarName = a) === null || _uniqueVarName === void 0 ? void 0 : _uniqueVarName.b.c;
```

szarnekow:
Merged

PR #1526, opened 2019-10-23
GH-1494: As a user I want support for optional chaining (aka elvis operator, etc.) (part 2)

mor-n4:
See #1494.

This part includes the transpilation of nullish coalescing operator and optional chaining to ES6.


mor-n4:
@mmews-n4 Please double check my assumption that a single helper variable will always be sufficient, not matter how many `?.` and `??` are used and no matter how they are nested.

mor-n4:

plugins/org.eclipse.n4js.transpiler.es/src/org/eclipse/n4js/transpiler/es/transform/ExpressionTransformation.xtend
```diff
@@ -219,11 +238,106 @@ class ExpressionTransformation extends Transformation {
...
+		} else {
+			replace(coalExpr, replacement);
+		}
+	}
+
+	/**
+	 * Converts
+	 * <pre>a?.b.c</pre>
+	 * to
+	 * <pre>($temp = a) == null ? void 0 : $temp.b.c</pre>
```
szarnekow: Is the `==null` check rewritten in subsequent step?

mor-n4: If the missing check for undefined is what you mean:
the cool kidz use `x == null` for `x === void 0 || x === null` these days (as I was told), they're equivalent (I know ... it's Javascript 😄 ).

szarnekow: Yes, that I was thinking about. Thank you for enlightening me, too :)

Issue #560, opened 2018-02-07
As a user I need informative error message for instanceof operator with use-site structural type


mor-n4:
```javascript
class A {
    public a: number;

    public f(): void {}
}

var a: ~A = { a: 2, f: function () {} };
let x = 5;

// XPECT FIXME errors --> "instanceof cannot be used with use site structural typing." at "~A"
// Getting: constructor{A} is not a subtype of number.
//          number is not a subtype of union{Function,type{Object},type{N4Enum}}.
a instanceof ~A;

// XPECT FIXME errors  --> "instanceof cannot be used with use site structural typing." at "~~A"
// Getting: number is not a subtype of union{Function,type{Object},type{N4Enum}}.
a instanceof ~~A;

// The following tests are only to ensure that existing correct functionality is maintained:

// XPECT errors --> "number is not a subtype of union{Function,type{Object},type{N4Enum}}." at "--x"
a instanceof --x;


// This needs to work! XPECT noerrors --> 
a instanceof A;

// XPECT errors --> "instanceof cannot be used with primitive types." at "number" 
a instanceof number; 
```
This is due to ~ being parsed as the INV (inverse, i.e. XOR) operator.

We need to do the following things:

Adjust the ExpectedTypeJudgment to get rid of wrong errors:
1. turn off expected type error for RelationalExpression, if operator is INSTANCEOF and right hand side is an UnaryExpression. We can do this by simply expecting anytype
2. turn off expected type error for UnaryExpression if parent is RelationalExpression and operator is INSTANCEOF. Same trick as in 1.

Add a new validation (in N4JSExpressionValidator) for exactly this case, extending the current INSTANCEOF validation. We need a new error message "'instanceof' cannot be used with use site structural typing.", since the error now stems from use site structural typeing (and not definition site, note that the typing strategy is not correctly parsed anyway.)

## Acceptance Criteria

The tests above must work.

## Hints
- the message is defined in message.properties, file IssuesCodes.xtend (in same package) must be touched to be recompiled
- in order to debug, run N4JS in debug mode. Unfortunately changing xtend code requires a restart. Do not simply "terminate" the runtime instance, but instead disable breakpoints and correctly quit the application

(moved from internal bug tracker, IDEBUG-163 and NumberFour/n4js#9; original reporter: @alisa-lisovskaya)

jpilgrim:
Assigned to Rodrigo, Eduard and Thanh (HAW Project).

PR #1530, opened 2019-10-28
GH-560: As a user I need informative error messages for instance operator with use-site structural type

rodrigoehlers:
See GH-560.

This PR fixes a bug which showed incorrect issues for a relational expression with `instanceof` and `~`.

1. `ExpectedTypeJudgement`
- If we're in a unary expression inside a relational expression with `instanceof` as operator, disable type issues by expecting any type.
- If we're in a relational expression with `instanceof` as operator, our right-hand-side is a unary expression and it's operator is `INV` eg. `~`, expect any type.

2. `messages.properties`
- Added error message to display when `instanceof` is used with use site structural typing.

3. `N4JSExpressionValidator`
- Add issue for relational expression, if right-hand-side is unary expression and it's operator is `INV` eg. `~`.

All spec tests passed on my local machine.


jpilgrim:
@mmews-n4 , @mor-n4 , @rodrigoehlers 
I'm not quite sure what went wrong in the tests:

```
Failed
org.eclipse.n4js.tests.project.NpmScopesPluginTest.testImportMainModule
Error Message
expected:<[]> but was:<[line 5: Project does not exist with project ID: n4js-runtime.]>
Stacktrace
org.junit.ComparisonFailure: expected:<[]> but was:<[line 5: Project does not exist with project ID: n4js-runtime.]>
	at org.eclipse.n4js.tests.project.NpmScopesPluginTest.testImportMainModule(NpmScopesPluginTest.java:255)
```

I would say that the changes are completely unrelated to that test. Note that this was the only failing test...

@rodrigoehlers Could you run that test on your machine.

EV21:
On my linux machine NpmScopesPluginTest runs fine (green)

> npm version
> {
>   npm: '6.12.0',
>   ares: '1.15.0',
>   brotli: '1.0.7',
>   cldr: '35.1',
>   http_parser: '2.8.0',
>   icu: '64.2',
>   llhttp: '1.1.4',
>   modules: '72',
>   napi: '5',
>   nghttp2: '1.39.2',
>   node: '12.13.0',
>   openssl: '1.1.1d',
>   tz: '2019a',
>   unicode: '12.1',
>   uv: '1.33.1',
>   v8: '7.7.299.13-node.12',
>   zlib: '1.2.11'
> }

mor-n4:
When I run `NpmScopesPluginTest` on this branch locally it works for me too. Have also started builds on our build nodes ...

mor-n4:
FYI: c14abcd built successfully (all tests green) on our Jenkins nodes.

rodrigoehlers:
Thanks @mor-n4.

I just added two more tests and changed implementation accordingly. All tests passed on my machine.

rodrigoehlers:
@jpilgrim Can you run our Jenkins for the nightly build?

plugins/org.eclipse.n4js/src/org/eclipse/n4js/typesystem/ExpectedTypeJudgment.java
```diff
@@ -287,6 +290,29 @@ public TypeRef casePostfixExpression(PostfixExpression e) {
...
+			EObject parent = e.eContainer();
+			if (parent instanceof UnaryExpression) {
+				parent = ((UnaryExpression) parent).eContainer();
+			}
+
+			if (parent instanceof RelationalExpression) {
+				RelationalExpression re = (RelationalExpression) parent;
+				boolean isInstanceOf = re.getOp().equals(RelationalOperator.INSTANCEOF);
+
+				if (isInstanceOf) {
```
jpilgrim: Could you add a comment here why we use "anyTypeRef" here -- e.g. reference AbstractN4JSDeclarativeValidator.checkRelationalExpression

plugins/org.eclipse.n4js/src/org/eclipse/n4js/typesystem/ExpectedTypeJudgment.java
```diff
@@ -287,6 +290,29 @@ public TypeRef casePostfixExpression(PostfixExpression e) {
...
 		@Override
 		public TypeRef caseUnaryExpression(UnaryExpression e) {
+			Expression innerExpression = e.getExpression();
+			TypeRef typeRef = ts.tau(innerExpression);
+
+			if (innerExpression instanceof UnaryExpression) {
+				typeRef = ts.tau(((UnaryExpression) innerExpression).getExpression());
+			}
+
+			boolean isNumber = RuleEnvironmentExtensions.isNumeric(G, typeRef);
```
jpilgrim: tau is not that cheap! Could you refactor the code so that tau (and typeRef etc.) are only used if necessary. E.g., first check the operator (I would assume INV) and if it is INV only then examine the rest  -- also taking care of the order in which things are checked and computed.

plugins/org.eclipse.n4js/src/org/eclipse/n4js/typesystem/ExpectedTypeJudgment.java
```diff
@@ -346,6 +372,23 @@ public TypeRef caseShiftExpression(ShiftExpression e) {
 		public TypeRef caseRelationalExpression(RelationalExpression e) {
 			switch (e.getOp()) {
 			case INSTANCEOF:
+				EObject rhs = e.getRhs();
+				if (rhs instanceof UnaryExpression) {
+					UnaryExpression ue = (UnaryExpression) rhs;
+					boolean isInv = ue.getOp().equals(UnaryOperator.INV);
```
jpilgrim: Similar to my comment above: if the operator is not INV, you can skip the rest of the extra handling. You do not even need a local variable then.

plugins/org.eclipse.n4js/src/org/eclipse/n4js/validation/validators/N4JSExpressionValidator.xtend
```diff
@@ -696,6 +697,24 @@ class N4JSExpressionValidator extends AbstractN4JSDeclarativeValidator {
...
+			
+			val rhs = relationalExpression.rhs;
+			if (rhs instanceof UnaryExpression) {
+				if ((rhs as UnaryExpression).getOp().equals(UnaryOperator.INV)) {
+					val innerExpression = (rhs as UnaryExpression).expression;
+					var rhsTypeRef = ts.tau(innerExpression);
+					
+					if (innerExpression instanceof UnaryExpression) {
+						rhsTypeRef = ts.tau((innerExpression as UnaryExpression).expression);
+					}
```
jpilgrim: First check the expression, only then compute tau. At the moment the computation in line 705 is not needed if the expression is an instance of UnaryExpression. In xtend, you can even write

```
val rhsTypeRef = tau (
        if (innerExpression instanceof UnaryExpression) {
           (innerExpression as UnaryExpression).expression;
        } else {
            innerExpression
        });
```

jpilgrim: Another remark: You check for unary expression in general. I'm not quite sure if this is not too general since we only want to handle the INV case here. I wasn't able to create a problem in N4JS, probably because the ExpectedTypeJudgment is handling this correct. I would suggest to only really limit the extra handling to the combination of INSTANCEOF and INV though.

tests/org.eclipse.n4js.spec.tests/xpect-tests/Ch05_05_StructuralTyping/Req76_3_StructuralAndInstanceof.n4js.xt
```diff
@@ -9,21 +9,36 @@
...
+// This needs to work! XPECT noerrors --> 
+a instanceof A;
 
-/* XPECT errors ---
-"constructor{T} is not a subtype of number." at "T"
-"number is not a subtype of union{Function,type{Object},type{N4Enum}}." at "~~T"
---- */
-t instanceof ~~T;
+// XPECT errors --> "instanceof cannot be used with primitive types." at "number" 
+a instanceof number;
```
jpilgrim: Maybe add a test with a different unary operator (such as ! or -) and different relational expression such as 

```
a instanceof (-A);
1 < ~A;
```

just to be sure that our extra handling is really limited to ~ and instanceof. (It does at the moment, just in case we will do some stupid refactorings in the future).

rodrigoehlers:
@jpilgrim Ready for merge, right?

Issue #1533, opened 2019-10-29
Reenable smoke tests on HAW build


jpilgrim:
In the context of, smoke tests were disabled since they failed to start due to some missing win32 modules. Which is weird as the build runs on Linux.

These tests should be re-enabled.

Log (example, messages are always similar):

```
!SESSION 2019-10-28 19:32:07.738 -----------------------------------------------
eclipse.buildId=unknown
java.version=11.0.4
java.vendor=Ubuntu
BootLoader constants: OS=linux, ARCH=x86_64, WS=gtk, NL=en
Framework arguments:  -application org.eclipse.tycho.surefire.osgibooter.headlesstest -testproperties /var/lib/jenkins/workspace/N4JSPR_n4js_PR-1520/tests/org.eclipse.n4js.smoke.tests/target/surefire.properties
Command-line arguments:  -data /var/lib/jenkins/workspace/N4JSPR_n4js_PR-1520/tests/org.eclipse.n4js.smoke.tests/target/work/data -application org.eclipse.tycho.surefire.osgibooter.headlesstest -testproperties /var/lib/jenkins/workspace/N4JSPR_n4js_PR-1520/tests/org.eclipse.n4js.smoke.tests/target/surefire.properties

!ENTRY org.eclipse.compare.win32 4 0 2019-10-28 19:32:14.138
!MESSAGE FrameworkEvent ERROR
!STACK 0
org.osgi.framework.BundleException: Could not resolve module: org.eclipse.compare.win32 [91]
  Unresolved requirement: Require-Capability: eclipse.platform; filter:="(osgi.os=win32)"

        at org.eclipse.osgi.container.Module.start(Module.java:462)
        at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel$2.run(ModuleContainer.java:1844)
        at org.eclipse.osgi.internal.framework.EquinoxContainerAdaptor$2$1.execute(EquinoxContainerAdaptor.java:150)
        at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel.incStartLevel(ModuleContainer.java:1837)
        at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel.incStartLevel(ModuleContainer.java:1778)
        at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel.doContainerStartLevel(ModuleContainer.java:1742)
[...]

Issue #1532, opened 2019-10-29
Move haw jenkins file to folder


jpilgrim:
Since we have more than one jenkins file, the nightly and other jenkins files have been moved to folder "jenkins". We need to do this with the HAW jenkins file as well.

PR #1534, opened 2019-10-29
GH-1532: moved jenkins to subfolder, disabled smoke tests

jpilgrim:
#1532

Side note: also use [ghef](https://www.npmjs.com/package/ghef) to filter events we can ignore on Jenkins (this does not change the repository)

Also disabled smoke tests since they caused problems... created #1533 for that.


jpilgrim:
@mmews-n4 Thanks!

Issue #1561, opened 2019-11-27
As a smith I want to reduce memory footprint of LSP builder


mmews-n4:
During #1535 the available memory was increased to build Stdlib, OPR, during tests etc. This is due to the lsp builder, which does not serialize the index and hence keeps resources in the memory.

Acceptance Criteria
- reduce memory footprint
- set memory limit back to 10G in Jenkins file
- set memory limit back to 4G in OPR/Stdlib build files

szarnekow:
The root cause of the high memory footprint appears to be that the entire set of resources in the workspace is kept in memory by the LSP server.

The situation is to be improved by only having open documents in memory with their full AST. Most of the other resources can be fully unloaded, the bigger part of the remaining files can be kept in memory without the AST. The type information is sufficient.

We'll approach this with the following multi-step solution:
- Put all built-in resources into a dedicated resource set. This is made transparent to the project's resource set by using the resource locator abstraction of EMF.
- After a project is built, unload all the AST and keep only the type information in memory.
- If the project is a leaf project in the dependency graph of the currently built projects, unload the entire resource. This is supposed to happen recursively.
- Projects with resources that are currently open, need a special treatment. Here we must switch from the per-project logic to a per-resource approach to figure which resources are safe to be unloaded. This needs to be refined.

PR #1579, opened 2019-12-04
GH-1561: Reduce memory footprint of LSP headless builder

szarnekow:
This PR ensures that resources are not loaded redundantly in the LSP builder case but
resolved from the dependent project's associated resource set instead. ResourceLocators to the rescue.

When a project is built initially, all its resources are loaded into a resource set including their AST. After the project build is done, the resource set is cleared.

Downstream projects in the dependency graph will attempt to use the resources from their dependencies. These are loaded from description into the resource set of their own project. Other projects may reuse these without the need to load them again. With the AST, the resources are very slim and can be kept in memory (for now).

Also the caching of validation issues was improved.

see #1561

### DIFF
--- a/plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/ProjectStatePersister.java
+++ b/plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/ProjectStatePersister.java
@@ -83,7 +83,7 @@ public class ProjectStatePersister {
 	 * - #vs times:
 	 * 	- source URI
 	 * 	- Number #vi of issues of source
-	 * 	- #vi times a validation issue
+	 * 	- #vi times a validation issue per N4JSIssue.writeExternal
 	 * </pre>
 	 */
 	private static final int VERSION_1 = 1;
@@ -112,7 +112,6 @@ public class ProjectStatePersister {
 		try {
 			File file = getDataFile(project);
 			try (OutputStream nativeOut = Files.asByteSink(file).openBufferedStream()) {
-
 				writeProjectState(nativeOut, N4JSLanguageUtils.getLanguageVersion(), state, files, validationIssues);
 
 			} catch (IOException e) {
@@ -183,16 +182,16 @@ public class ProjectStatePersister {
 		}
 	}
 
-	private void writeValidationIssues(Map<URI, ? extends Collection<Issue>> validationIssues, ObjectOutput ouput)
+	private void writeValidationIssues(Map<URI, ? extends Collection<Issue>> validationIssues, ObjectOutput output)
 			throws IOException {
 
 		int numberSources = validationIssues.size();
-		ouput.writeInt(numberSources);
+		output.writeInt(numberSources);
 		for (Map.Entry<URI, ? extends Collection<Issue>> srcIssues : validationIssues.entrySet()) {
 			URI source = srcIssues.getKey();
 			Collection<Issue> issues = srcIssues.getValue();
 
-			ouput.writeUTF(source.toString());
+			output.writeUTF(source.toString());
 
 			Collection<N4JSIssue> n4Issues = new ArrayList<>();
 			for (Issue issue : issues) {
@@ -202,9 +201,9 @@ public class ProjectStatePersister {
 			}
 
 			int numberIssues = n4Issues.size();
-			ouput.writeInt(numberIssues);
+			output.writeInt(numberIssues);
 			for (N4JSIssue issue : n4Issues) {
-				ouput.writeObject(issue);
+				issue.writeExternal(output);
 			}
 		}
 	}
@@ -325,7 +324,8 @@ public class ProjectStatePersister {
 			int numberOfIssues = input.readInt();
 			while (numberOfIssues > 0) {
 				numberOfIssues--;
-				N4JSIssue issue = (N4JSIssue) input.readObject();
+				N4JSIssue issue = new N4JSIssue();
+				issue.readExternal(input);
 				validationIssues.get(source).add(issue);
 			}
 		}
--- a/plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/ProjectStatePersister.java
+++ b/plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/ProjectStatePersister.java
@@ -83,7 +83,7 @@ public class ProjectStatePersister {
 	 * - #vs times:
 	 * 	- source URI
 	 * 	- Number #vi of issues of source
-	 * 	- #vi times a validation issue per N4JSIssue.writeExternal
+	 * 	- #vi times a validation issue as per {@link N4JSIssue#writeExternal(ObjectOutput) N4JSIssue.writeExternal}
 	 * </pre>
 	 */
 	private static final int VERSION_1 = 1;
--- a/plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/UriResourceMap.java
+++ b/plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/UriResourceMap.java
@@ -1,0 +1,85 @@
+/**
+ * Copyright (c) 2019 NumberFour AG.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   NumberFour AG - Initial API and implementation
+ */
+package org.eclipse.n4js.ide.xtext.server;
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+import org.eclipse.emf.common.util.URI;
+import org.eclipse.emf.ecore.resource.Resource;
+
+import com.google.common.collect.Collections2;
+import com.google.common.collect.ForwardingMap;
+import com.google.inject.Inject;
+
+/**
+ * A project specific map of resources per URI. It acts as a mere filter to the {@link GlobalUriResourceMap} such that
+ * an invocation of {@link #clear()} does not wipe out the entire map contents but only the resources from the current
+ * project.
+ */
+public class UriResourceMap extends ForwardingMap<URI, Resource> {
+
+	private final GlobalUriResourceMap globalMap;
+
+	private final Set<URI> localContents;
+
+	/**
+	 * Standard constructor
+	 */
+	@Inject
+	public UriResourceMap(GlobalUriResourceMap globalMap) {
+		this.globalMap = globalMap;
+		this.localContents = new HashSet<>();
+	}
+
+	@Override
+	protected Map<URI, Resource> delegate() {
+		return globalMap;
+	}
+
+	@Override
+	public void clear() {
+		delegate().keySet().removeAll(localContents);
+		localContents.clear();
+	}
+
+	@Override
+	public Resource put(URI key, Resource value) {
+		localContents.add(key);
+		return super.put(key, value);
+	}
+
+	@Override
+	public void putAll(Map<? extends URI, ? extends Resource> map) {
+		localContents.addAll(map.keySet());
+		super.putAll(map);
+	}
+
+	@Override
+	public boolean remove(Object key, Object value) {
+		if (localContents.remove(key)) {
+			return super.remove(key, value);
+		}
+		return false;
+	}
+
+	@Override
+	public Collection<Resource> values() {
+		if (localContents.isEmpty()) {
+			return Collections.emptyList();
+		}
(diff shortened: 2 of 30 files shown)


plugins/org.eclipse.n4js/src/org/eclipse/n4js/validation/helper/IssuesProvider.java
```diff
@@ -71,7 +73,7 @@ public IssuesProvider(IResourceValidator resourceValidator, Resource res,
 				: N4JSDataCollectors.dcValidations;
 		List<Issue> issues;
 		try (Measurement m = dc.getMeasurement("validation");) {
-			issues = rv.validate(r, CheckMode.ALL, ci);
+			issues = rv.validate(r, checkMode, ci);
```
mmews-n4: Thanks for adding code to respect the check mode.

plugins/org.eclipse.n4js/src/org/eclipse/n4js/validation/N4JSResourceValidator.java
```diff
@@ -102,6 +104,11 @@
 		return super.validate(resource, mode, cancelIndicator);
 	}
 
+	@Override
+	public List<Issue> validate(Resource resource, CheckMode mode, CancelIndicator cancelIndicator) {
+		return n4jsCache.getOrElseUpdateIssues(this::doValidate, resource, mode, cancelIndicator);
```
mmews-n4: Thanks for using the cache here (instead of in the N4JSCache itself).

plugins/org.eclipse.n4js.transpiler.es/src/org/eclipse/n4js/transpiler/es/transform/EnumAccessTransformation.xtend
```diff
@@ -41,14 +41,14 @@ class EnumAccessTransformation extends Transformation {
...
 	override assertPreConditions() {
-		member_StringBasedEnum_literals = state.G.n4StringBasedEnumType.findOwnedMember("literals", false, true);
 		assertNotNull("member of built-in type not found: StringBasedEnum#literals", member_StringBasedEnum_literals);
 	}
 
 	override assertPostConditions() {
 	}
 
 	override analyze() {
+		member_StringBasedEnum_literals = state.G.n4StringBasedEnumType.findOwnedMember("literals", false, true);
```
mmews-n4: 👍

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/XProjectManager.java
```diff
@@ -84,6 +84,16 @@
...
 
+	/**
+	 * The map for this project's resource set.
+	 */
+	@Inject
+	protected UriResourceMap uriResourceMap;
+
+	/** The workspace manager */
+	@Inject
+	protected XWorkspaceManager workspaceManager;
```
mmews-n4: This field seems to be not used in this class.

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/UriResourceMap.java
```diff
@@ -0,0 +1,85 @@
...
+import com.google.common.collect.Collections2;
+import com.google.common.collect.ForwardingMap;
+import com.google.inject.Inject;
+
+/**
+ * A project specific map of resources per URI. It acts as a mere filter to the {@link GlobalUriResourceMap} such that
+ * an invocation of {@link #clear()} does not wipe out the entire map contents but only the resources from the current
+ * project.
+ */
+public class UriResourceMap extends ForwardingMap<URI, Resource> {
```
mmews-n4: Could the class name reflect that this map is reading and writing to a global state, but deleting only those entries which it previously added?

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/XProjectManager.java
szarnekow: It is used in org.eclipse.n4js.ide.xtext.server.WorkspaceAwareResourceLocator

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/UriResourceMap.java
szarnekow: Hmm ... it probably should. What do you think about `ProjectAwareUriResourceMap`?

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/ProjectStatePersister.java
```diff
@@ -83,7 +83,7 @@
 	 * - #vs times:
 	 * 	- source URI
 	 * 	- Number #vi of issues of source
-	 * 	- #vi times a validation issue
+	 * 	- #vi times a validation issue per N4JSIssue.writeExternal
```
mor-n4: I don't get what "per" means, here. Maybe "as emitted by" instead?

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/UriResourceMap.java
mor-n4: Seems to me "Adapter" should be in the name, because it does not contain any key/value pairs itself and acts only as an adapter of the `GlobalUriResourceMap`.

What about `ProjectSpecificGlobalUriResourceMapAdapter` (I know, it's long ;-) )

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/WorkspaceAwareCanLoadFromDescriptionHelper.java
```diff
@@ -0,0 +1,37 @@
...
+import org.eclipse.xtext.resource.XtextResourceSet;
+
+import com.google.inject.Inject;
+import com.google.inject.Singleton;
+
+/**
+ * A specialization that will create the resources in the correct project specific resource set.
+ */
+@Singleton
+public class WorkspaceAwareCanLoadFromDescriptionHelper extends CanLoadFromDescriptionHelper {
```
mor-n4: The naming scheme "WorkspaceAware..." for LSP-specific adjustments is a bit unfortunate, because 1) it does not explicitly highlight the fact that it is specific to the LSP case and 2) the term "Workspace" is also used in Eclipse a lot.

However, I guess most of the Eclipse cases will soon be obsolete anyway (and will be removed), right? In that case it's just temporary, so not a problem.

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/XLanguageServerImpl.java
```diff
@@ -1149,30 +1149,31 @@ public InitializeResult getInitializeResult() {
...
+										if (coloringService != null && client instanceof LanguageClientExtensions) {
+											Document doc = ctx.getDocument();
+											List<? extends ColoringInformation> coloringInfos = coloringService
+													.getColoring(resource, doc);
+											if (!IterableExtensions.isNullOrEmpty(coloringInfos)) {
+												String uri = resource.getURI().toString();
+												((LanguageClientExtensions) client)
+														.updateColoring(new ColoringParams(uri, coloringInfos));
+											}
+										}
```
mor-n4: Wow!
I guess this is original code of the LSP implementation?

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/XProjectManager.java
```diff
@@ -113,23 +125,30 @@ public void initialize(ProjectDescription description, IProjectConfig projectCon
 	/** Initial build reads the project state and resolves changes. */
 	public XBuildResult doInitialBuild(CancelIndicator cancelIndicator) {
 		Set<URI> changedSources = projectStateHolder.readProjectState(projectConfig);
-
+		// TODO distinguish between changed sources and deleted sources
```
mor-n4: Could you add a ticket number?

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/XProjectManager.java
```diff
@@ -113,23 +125,30 @@ public void initialize(ProjectDescription description, IProjectConfig projectCon
...
+		// clear the resource set to release memory
+		boolean wasDeliver = resourceSet.eDeliver();
+		try {
+			resourceSet.eSetDeliver(false);
+			resourceSet.getResources().clear();
+		} finally {
+			resourceSet.eSetDeliver(wasDeliver);
 		}
+
+		persistProjectState();
```
mor-n4: ~Did you make sure to only do this if persisting wasn't turned off via command line option?~
(all good; it's being checked in the ProjectStateHolder)

plugins/org.eclipse.n4js/src/org/eclipse/n4js/resource/N4JSResource.java
```diff
@@ -1210,6 +1228,24 @@ private String defaultGetURIFragment(EObject eObject) {
...
 	}
 
+	/**
+	 * {@inheritDoc}
+	 * <p>
+	 * Overridden to make sure that we do not initialize a resource just to compute the root URI fragment for the parse
+	 * result.
+	 */
+	@Override
+	protected String getURIFragmentRootSegment(EObject eObject) {
```
mor-n4: Good find!

plugins/org.eclipse.n4js/src/org/eclipse/n4js/scoping/utils/CanLoadFromDescriptionHelper.java
```diff
@@ -243,4 +243,11 @@ private boolean projectContainsURI(IN4JSProject project, URI candidate) {
 		return Optional.ofNullable(description).flatMap(UserdataMapper::readDependenciesFromDescription);
 	}
 
+	/**
+	 * Create a resource in the context of the given resource set.
+	 */
+	public Resource createResource(ResourceSet resourceSet, URI resourceURI) {
```
mor-n4: Could you please explain what use cases this method is intended for? Because it is not intended (I guess) that _all_ Resource creation be handled by this method, in the future.

If I understand correctly, it's only intended for creation of resources "during load from index", right? (maybe "during load from index" can be worded/explained better)

tests/org.eclipse.n4js.lang.tests/src/org/eclipse/n4js/tests/resource/N4JSResourceLoadStatesTest.xtend
```diff
@@ -364,7 +364,7 @@ class N4JSResourceLoadStatesTest extends AbstractN4JSTest {
 				info.reconciled, res.reconciled);
 		}
 	}
-
+
```
mor-n4: Trailing white space! 😈

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/UriResourceMap.java
szarnekow: Offline discussion: Use `WorkspaceUriResourceMap` and `ProjectUriResourceMap`

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/WorkspaceAwareCanLoadFromDescriptionHelper.java
szarnekow: I'd like to keep it like this. If it turns out, that we are confused by the terminology, we should rename this late on. In the LSP context (read: java package), a Workspace is well defined.

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/XLanguageServerImpl.java
szarnekow: Nice, isn't it ;)

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/XProjectManager.java
szarnekow: Refactored

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/XProjectManager.java
szarnekow: #1583 - I won't add it to the code, since I'll work on it now

PR #1584, opened 2019-12-04
GH-1561: Reduce memory consumption when creating the test catalog

szarnekow:
The catalog used to be computed based on an empty resource set without any index information attached. This was fixed to use the available index also in the headless case.

see #1561

### DIFF
--- a/plugins/org.eclipse.n4js.cli/src/org/eclipse/n4js/cli/compiler/N4jscCompiler.java
+++ b/plugins/org.eclipse.n4js.cli/src/org/eclipse/n4js/cli/compiler/N4jscCompiler.java
@@ -15,6 +15,7 @@ import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.OutputStream;
+import java.nio.charset.StandardCharsets;
 import java.nio.file.Path;
 import java.util.Set;
 import java.util.SortedMap;
@@ -202,7 +203,7 @@ public class N4jscCompiler {
 					true); // do not include "endpoint" property here
 
 			try (OutputStream os = new BufferedOutputStream(new FileOutputStream(testCatalogFile))) {
-				os.write(catalog.getBytes("UTF-8"));
+				os.write(catalog.getBytes(StandardCharsets.UTF_8));
 				os.flush();
 
 			} catch (IOException e) {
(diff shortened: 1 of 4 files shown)


plugins/org.eclipse.n4js.tester/src/org/eclipse/n4js/tester/TestDiscoveryHelper.java
```diff
@@ -168,19 +167,28 @@ public boolean isTestable(final URI location) {
 	 *         tree may be empty.
 	 */
 	public TestTree collectTests(final List<URI> uris) {
-		final ResourceSet resSet = n4jsCore.createResourceSet(Optional.absent());
-		final IResourceDescriptions index = n4jsCore.getXtextIndex(resSet);
```
mor-n4: This line used to return an index with TModule information (not an empty index), at least in the UI case.
Has this behavior changed? (in this case, please adjust the API doc of `IN4JSCore#getXtextIndex()`) Or was there some other reason why - before this PR - everything was rebuilt from scratch?

plugins/org.eclipse.n4js.cli/src/org/eclipse/n4js/cli/compiler/N4jscCompiler.java
```diff
@@ -196,11 +198,12 @@ private void writeTestCatalog() throws N4jscException {
...
 
-			String catalog = testCatalogSupplier.get(true); // do not include "endpoint" property here
+			String catalog = testCatalogSupplier.get((uri) -> workspaceManager.getProjectManager(uri).getResourceSet(),
+					true); // do not include "endpoint" property here
 
-			try (FileOutputStream fos = new FileOutputStream(testCatalogFile)) {
-				fos.write(catalog.getBytes());
-				fos.flush();
+			try (OutputStream os = new BufferedOutputStream(new FileOutputStream(testCatalogFile))) {
+				os.write(catalog.getBytes("UTF-8"));
```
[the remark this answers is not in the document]
szarnekow: :+1:

plugins/org.eclipse.n4js.tester/src/org/eclipse/n4js/tester/TestDiscoveryHelper.java
szarnekow: `n4jsCore.getXtextIndex(resSet)` tries to find an adapter in the resource set that points to "the index". A newly created resource set doesn't have index information attached. I'm not sure how it worked in the old headless compiler, but it's perfectly reasonable to find an empty index in a an empty resource set in the standalone model. That's why we need to pass it into this method.

I'm _still_ not sure if it's a good idea to have `n4jsCore.createResourceSet` as an API at all.

mor-n4: With all the current changes in the LSP context, maybe now is a good time to reconsider these two methods in `IN4JSCore` in general. I think their intended use case is still valid, but maybe there's a better way to do it. Let's discuss this on Monday, also with @mmews-n4 (no need to postpone this PR, though).

szarnekow: I created ticket #1586

Issue #1597, opened 2019-12-10
Fix failing integration test related to missing checkout of master branch


mmews-n4:
Adjust git utils to check out master first, and then the snapshot branch. This is necessary for git commands on the command line which would fail otherwise.

PR #1598, opened 2019-12-10
GH-1597: Fix failing integration test related to missing checkout of master branch

mmews-n4:
see #1597


plugins/org.eclipse.n4js.utils/src/org/eclipse/n4js/utils/git/GitUtils.java
```diff
@@ -213,6 +221,9 @@ public static void hardReset(final Iterable<String> remoteUris, final Iterable<P
 			final int pathIndex = i++;
 			new Thread(() -> {
 				try {
+					// checkout master first
+					hardReset(remoteUri, paths[pathIndex], "master", cloneIfMissing, clean);
+					// now switch to requested branch
 					hardReset(remoteUri, paths[pathIndex], branch, cloneIfMissing, clean);
```
mor-n4: Wouldn't it make sense to pass in `false` for parameter `cloneIfMissing`, because we just cloned it in line 225 if it did not exist?

mor-n4: The problem was that the working copy was dirty and thus the branch switch in lines 154-155 failed, right?

Instead of invoking `#hardReset(String, Path, ...)` twice in lines 225, 227 I think it would be enough to change the order of hard reset (line 166f.) and branch switching (line 154f.). (Not second hard reset required after switching the branch, because the working copy should be clean after switching the branch if it was clean before, I guess.)

Also, that would fix the issue for clients that invoke `#hardReset(String, Path, ...)` directly.

mmews-n4: Since your update branch of node and yarn apparently and surprisingly fixes the problem, mind my comment as info only: The original problem here was, that branch master was not available in the snapshot clones, because the checkouts in lines 489-494 checked out the requested branch directly. So, regarding your paragraph "Instead of invoking ...", I can only say that the condition (`!currentBranch.equals(branch)` in line 147) of branch switching (line 154f) was always false (in case the snapshot was cloned before and not re-used from another test that checkout another branch).

However, since you got it working on #1600 I'll stop with this branch here and wait. In case #1600 will also work on master, I can close this branch.

Issue #1600, opened 2019-12-10
As a smith I want to update to the latest version of node / yarn


mor-n4:
Update the builds to use the latest LTS / stable version of node.js and yarn.

* node 12.13.1
* yarn 1.19.2

Adjust build setup and tests, if necessary.

mor-n4:
see https://bugs.eclipse.org/bugs/show_bug.cgi?id=558060

PR #1601, opened 2019-12-11
GH-1600: As a smith I want to update to the latest version of node / yarn

mor-n4:
See #1600.

Eclipse nightly was green with these changes.


mor-n4:
@jpilgrim If we merge this, you'd have to update to node 12.13.1 / yarn 1.19.2. Is this ok for you?

We could change the output Xpect method to ignore white space, but I did not want to do that unless there is true need for it, because it might weaken some test assertions.

README.adoc
```diff
@@ -62,7 +62,7 @@ Ensure you have
 
 - Java 11
 - Maven 3.2.x and
-- Node.js >=8
+- Node.js 12.3.x (recommended).
```
szarnekow: recommended or mandatory?

tests/org.eclipse.n4js.expectmatrix.tests/xpect-tests/builtin_js/array/array_concat.n4js.xt
```diff
@@ -38,7 +38,10 @@ console.log(numeric);
 <==
 stdout:
 [ 3, 4, 5, 1, 2, 3 ]
-[ 1, 2, 3, 4, 5, 6, 7, 8, 9 ]
+[
```
szarnekow: Should we use a JSON parser in the tests to assert the output?

mmews-n4:
> We could change the output Xpect method to ignore white space

@mor-n4 Do you know if the changes in the serializations are only due to newline characters placed at other locations? Or is it more than that?

mor-n4:
More than that. Sometimes more/fewer spaces are added and sometimes curly braces where moved such as: `{ Name ... }` changed to `Name { ... }`. The latter would not be solved by ignoring white space, but these were only few cases.

mmews-n4:
Ok, I see.

README.adoc
mor-n4: For just building the product(s) (no tests), versions 10+ should be ok.

tests/org.eclipse.n4js.expectmatrix.tests/xpect-tests/builtin_js/array/array_concat.n4js.xt
mor-n4: The output is not always valid JSON. But it might be an option in some cases (i.e. not changing output Xpect method in general but introducing a separate Xpect method that compares output via JSON).

mor-n4:
On second thought, I think we should always fix tests in case they depend on certain node versions. It does not make sense to make our build rely on a particular version of node and always force everyone to update node at the same time, just because of some `console.log()` output comparisons. I'll have a look at those 26 offending tests again.

@jpilgrim this means you do not have to update node on your side, if you don't want to.

mor-n4:
Thanks everyone!

Tested with new and old node version, all tests working now.

Merging ...

Issue #1608, opened 2019-12-12
As a user I don't want to compile dirty files


szarnekow:
Editing files with LSP clients will trigger incremental builds on the server. The builds that are not due to persistent changes on disk should not generate any files.

PR #1612, opened 2019-12-12
Proper shutdown synchronization, dont generate for dirty files

szarnekow:
Still work in progress because of the sysouts. We need to figure out how to do proper logging.

see #1608


plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/concurrent/XRequestManager.java
```diff
@@ -41,6 +43,8 @@
 	public void shutdown() {
 		queue.shutdown();
 		parallel.shutdown();
+		MoreExecutors.shutdownAndAwaitTermination(queue, 2500, TimeUnit.MILLISECONDS);
+		MoreExecutors.shutdownAndAwaitTermination(parallel, 2500, TimeUnit.MILLISECONDS);
```
mmews-n4: Thanks for adding timeouts here.

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/XWorkspaceManager.java
```diff
@@ -428,14 +429,46 @@ public XProjectManager getProjectManager(String projectName) {
...
-	 * Mark the given document as cloded.
+	 * Mark the given document as closed.
+	 *
+	 * When we close a document, being it a saved document or a dirty document, we need to run a build to make sure that
+	 * the generated code is consistent.
+	 *
+	 * Reason:<br>
+	 * Two files are open, dirty and refer to each other. One file is saved and code is generated. This generation is
+	 * done against the dirty (!) state of the other open document in LSP. When the other file is closed, the generator
+	 * needs to run again.
```
mmews-n4: Somehow I wonder how to make the implementation of the constraint "Source and transpiled files are always in sync on disk" more obvious. Maybe even including the n4js.projectstate. (Whereas the internal type index always relates to the current editor state, no matter of being in sync with the file on disk or being dirty instead.) ... Future refactorings.

szarnekow: 👍

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/XLanguageServerImpl.java
```diff
@@ -243,9 +245,13 @@ public void setWorkspaceManager(XWorkspaceManager manager) {
 
 		access.addBuildListener(this);
 
+		System.out.println("About to initialize the server with workspace " + baseDir);
```
szarnekow: What would be the correct way to do logging here, @mmews-n4 ?

mmews-n4: Maybe `LOG.info(...);` can be used here?

mmews-n4:
I continue to work on this branch. There is still the issue of how to deal with dirty state vs. file system state regarding source files vs. generated files, and also regarding which issues to show. This boils down to:
- when do we generate files from source files?

Especially consider in case of generation while having dirty state: 
- Should we generate at all?
- If yes, is it intuitive regarding error markers? (which will be not in sync either with dirty files, or with files on hdd?)
- If yes, is it intuitive to have source and generated files (both on hdd) out of sync (since generation would consider dirty state files)?
- Alternatively, is it intuitive to have source and generated files (both on hdd) __in sync__ but not in sync with dirty files in editor? -> mind error markers which might not match.

Also mind:
- https://github.com/eclipse/xtext-core/issues/373
- https://github.com/itemis/xtext-languageserver-example/issues/50

szarnekow:
Thank you for taking this over. I haven't come around yet to address the review remarks.

mmews-n4:
Regarding dirty files and generation, I can think of the following solutions:
- Manually and/or external generation: No automatic generation by LSP server.
- Generation only on clean editor state: Generate when the last dirty file is saved or closed.
- Mimic Eclipse dirty state support: Generate on save, ignore dirty states, create (special?) error markers for dirty files, since validation (before generation) runs against hdd file (note: if no generation is done, validation runs against dirty state)

I find that the following solutions come with too many disadvantages, or are too surprising for the user:
- Generate on every change (behaviour on master)
- Generate based on dirty files

Regarding "Mimic Eclipse dirty state support" mind that there is a problem with LSP, since LSP does not distinguish between issues of dirty files (in editor) and of hdd files, which becomes obvious when the hdd file contains sources that are deleted in the dirty file. Errors that would show up in this deleted region cannot be displayed in the editor anymore. More important: Mixing those issues in the same editor file is very confusing for the user. However, showing issues (errors) or some other information to the user would be reasonable, since errors will prevent the generation.

mmews-n4:
Regarding this user story, I chose the second alternative `Generation only on clean editor state`, because it is a simple and more intuitive solution than the current behaviour on master (which is `Generate on every change`). Later, this second alternative can be enhanced (e.g. by mimicking Eclipse dirty state behaviour).

For future reference, keep the following scenario in mind when activating generation on every save action (originally by @szarnekow):
>Two files are open, dirty and refer to each other. One file is saved and code is generated. This generation is done against the dirty (!) state of the other open document in LSP. When the other file is closed, the generator needs to run again.

mor-n4:
> For future reference, keep the following scenario in mind when activating generation on every save action (originally by @szarnekow):
> 
> > Two files are open, dirty and refer to each other. One file is saved and code is generated. This generation is done against the dirty (!) state of the other open document in LSP. When the other file is closed, the generator needs to run again.

Do we have a test implementing this scenario?

Also, the scenario description should be worded more precisely. For example: 1) "... When the other file is closed _without being saved_, the generator needs to run again." 2) Is a cyclic dependency between the files really necessary?

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/XProjectManager.java
```diff
@@ -162,21 +154,25 @@ public XBuildResult doInitialBuild(CancelIndicator cancelIndicator) {
...
 	}
 
-	/** Build this project. */
+	/** Build increments of this project. */
 	public XBuildResult doIncrementalBuild(Set<URI> dirtyFiles, Set<URI> deletedFiles,
-			List<IResourceDescription.Delta> externalDeltas, CancelIndicator cancelIndicator) {
-		return doIncrementalBuild(dirtyFiles, deletedFiles, externalDeltas, cancelIndicator, Function.identity());
+			List<IResourceDescription.Delta> externalDeltas, boolean doGenerate, CancelIndicator cancelIndicator) {
+
+		return doBuild(dirtyFiles, deletedFiles, externalDeltas, true, doGenerate, cancelIndicator);
```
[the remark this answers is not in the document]
mmews-n4: Yes. Mind that the other method that calls `doBuild()` is  `doInitialBuild()` which has more additional logic. Also, `doBuild()` is `protected` only.

mmews-n4:
>> Two files are open, dirty and refer to each other. One file is saved and code is generated. This
>> generation is done against the dirty (!) state of the other open document in LSP. When the other file 
>> is closed, the generator needs to run again.
>
> Do we have a test implementing this scenario?

No.

> Also, the scenario description should be worded more precisely. For example: 1) "... When the other 
> file is closed without being saved, the generator needs to run again." 2) Is a cyclic dependency 
> between the files really necessary?

1) - Thanks!
2) - "refer to each other" might not necessarily be meant as having an import cycle.

mmews-n4:
Thanks for all your feedback!

Issue #1617, opened 2019-12-17
Create warnings for variables with values of optional chained expressions


mmews-n4:
The following snipped shows the use of optional chaining and a subsequent use of a variable `x`. Semantically, the value of `x` must be either non-null or null, since the OptionalChainingPunctuator (ie. `?`) is a strong indicator for some value being `undefined` and put deliberately in the source code by the developer.

```ts
let x = a?.b;
x.toString; // FIXME: add a warning here, since x can be undefined
```

PR #1619, opened 2019-12-17
GH-1617: Create warnings for variables with values of optional chained expressions

mmews-n4:
see #1617

### DIFF
--- a/tests/org.eclipse.n4js.xpect.tests/model/validation/dataflow/NullUndefined_May_OptionalChaining.n4js.xt
+++ b/tests/org.eclipse.n4js.xpect.tests/model/validation/dataflow/NullUndefined_May_OptionalChaining.n4js.xt
@@ -1,0 +1,30 @@
+/*
+ * Copyright (c) 2017 NumberFour AG.
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   NumberFour AG - Initial API and implementation
+ */
+
+/*	XPECT_SETUP org.eclipse.n4js.xpect.tests.N4JSXpectTest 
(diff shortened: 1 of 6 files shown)


tests/org.eclipse.n4js.xpect.tests/model/validation/dataflow/NullUndefined_May_OptionalChaining.n4js.xt
```diff
@@ -0,0 +1,115 @@
...
+/*	XPECT_SETUP org.eclipse.n4js.xpect.tests.N4JSXpectTest 
+	IssueConfiguration {
+		IssueCode "DFG_NULL_DEREFERENCE" {enabled=true}
+	}
+	END_SETUP
+ */
+
+class C {
+	field: C;
+}
```
mor-n4: You might wanna add a test case for the "null not passed through optional chaining" rule:
```javascript
function f() {
	let c : C = new C();
	if (true) {
		c = null;
	}
	// XPECT warnings --> "Variable c may be null" at "x"
	c.field;

	let x = c?.field;
	// XPECT warnings --> "Variable x may be undefined" at "x"
	x.toString();
}
```

tests/org.eclipse.n4js.xpect.tests/model/validation/dataflow/NullUndefined_May_OptionalChaining.n4js.xt
```diff
@@ -0,0 +1,115 @@
...
+	// XPECT warnings --> "Variable x may be undefined" at "x"
+	x.toString();
+}
+
+
+function f5() {
+	let c : C;
+	let x = c?.field;
+	// XPECT warnings --> "Variable x may be undefined" at "x"
+	x.toString();
```
mor-n4: I think in this case it would make sense to warn with `Variable x is(!) undefined`.

mmews-n4:
Good find. I improved the analysis to also follow the data flow of the target of optional chaining.

Issue #1622, opened 2020-01-03
As a smith I want to remove the obsolete bundle 'generator.headless' and its test bundle


mmews-n4:
Also, adjust tests that still use this bundle like the access control tests. Adjust performance tests to either work with the new headless compiler or set them to ignore.

PR #1623, opened 2020-01-03
GH-1622: As a smith I want to remove the obsolete bundle 'generator.headless' and its test bundle

mmews-n4:
see #1622


plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/server/FileBasedWorkspaceInitializer.java
```diff
@@ -75,16 +82,97 @@ public IWorkspaceConfig getWorkspaceConfig(URI workspaceBaseURI) {
...
 		}
 	}
 
+	/**
+	 * Registers provided project URIs in the given workspace.
+	 *
+	 * Skips {@link IN4JSProject}s that are already registered with the given {@code workspace}.
+	 *
+	 */
+	public void registerProjectsToFileBasedWorkspace(Iterable<FileURI> projectURIs) {
```
mor-n4: This is only called from `#getWorkspaceConfig()` above, never from the outside. Would it make sense to make this private?

Issue #1624, opened 2020-01-06
As a smith I want to get rid of manually modified JS file N4BuiltInClasses.js


mor-n4:
Currently, there are two files in the repository called "N4BuiltInClasses":
```
/org.eclipse.n4js.environments/src/N4BuiltInClasses.n4js
/n4js-runtime/src/js/rt/N4BuiltInClasses.js
```
The JS file was generated from the N4JS using the N4JS transpiler a long time ago but has then been modified manually over the past years (without re-generating it from the N4JS file). Hence, the two files have deviated from one another.

The current situation is a huge problem when major changes to transpilation of classes, etc. are implemented in the transpiler, because then the JS would have to be adjusted manually which is difficult and error prone.

### ACs

1. remove JS file `N4BuiltInClasses.js` from the repository,
2. apply adjustments to the handling of built-in types such that everything still works without the JS file
   (there should only be an N4JS source file remaining for module N4BuiltInClasses),
3. investigate whether the redundancy in built-in type declaration/definition between files `builtin_n4.n4ts` and `N4BuiltInClasses` can be resolved without too much effort (maybe just partially by moving the reflection-related built-in types to a separate file),
4. check if bundle `org.eclipse.n4js.environments` is now obsolete and remove it if possible,
5. consider renaming `N4BuiltInClasses.n4js` to `N4BuiltInTypes.n4js`

PR #1625, opened 2020-01-06
GH-1624: As a smith I want to get rid of manually modified JS file N4BuiltInClasses.js

mor-n4:
See #1624.


mor-n4:
PENDING: revert commit bb7d6b7 before merging!

Issue #1632, opened 2020-01-23
As a smith I want to package the N4JS LSP extension for VSCode


mmews-n4:
AC:
* Change n4js-cli 
  * to install Java 11 on npm install if not globally available
  * to use installed (not global) Java when present
* package n4js-vscode-extension

PR #1633, opened 2020-01-24
GH-1632: As a smith I want to package the N4JS LSP extension for VSCode

mmews-n4:
see #1632


plugins/org.eclipse.n4js/src/org/eclipse/n4js/validation/validators/N4JSTypeValidator.xtend
```diff
@@ -518,27 +520,54 @@ class N4JSTypeValidator extends AbstractN4JSDeclarativeValidator {
 				return;
 			}
 
+			val usedInSpecConstructor = objectLiteral.eContainer?.eContainer instanceof NewExpression;
```
mor-n4: This will produce false positives (not every object literal passed into a constructor is the spec argument to a spec constructor).
The rule must be something like: (a) the object literal must be an argument to a ctor (b) the formal parameter of that ctor receiving the object literal as argument must be annotated with `@Spec`. (note that spec ctors may have more parameters than the spec parameter)

As Xpect test:
```javascript
class C {
	public prop: string;
	private propOther: string;

	constructor(arg0: ~i~this, @Spec arg1: ~i~this, arg2: ~i~this) {}
}

class D {
	constructor(arg0: ~C) {}
}

let c: ~C;

// works XPECT warnings --> "propUnknown is not defined in ~C; it will not be accessible from c." at "propUnknown"
c = {prop: "", propOther: "", propUnknown: ""};

function foo(arg: ~C) {}
// works XPECT warnings --> "propUnknown is not defined in ~C; it will not be accessible from the receiving parameter." at "propUnknown"
foo({prop: "", propOther: "", propUnknown: ""});

// this is not a spec constructor XPECT FIXME warnings --> "propUnknown is not defined in ~C; it will not be accessible from the receiving parameter." at "propUnknown"
new D({prop: "", propOther: "", propUnknown: ""});


new C(
	// not a spec argument XPECT FIXME warnings --> "propUnknown is not defined in ~C; it will not be accessible from the receiving parameter." at "propUnknown"
	{prop: "", propOther: "", propUnknown: ""},
	// works XPECT warnings ---
	// "propOther is not defined in ~i~this[C]; it will not have any effect in the spec constructor." at "propOther"
	// "propUnknown is not defined in ~i~this[C]; it will not have any effect in the spec constructor." at "propUnknown"
	// ---
	{prop: "", propOther: "", propUnknown: ""},
	// not a spec argument XPECT FIXME warnings --> "propUnknown is not defined in ~C; it will not be accessible from the receiving parameter." at "propUnknown"
	{prop: "", propOther: "", propUnknown: ""}
);
```

plugins/org.eclipse.n4js/src/org/eclipse/n4js/validation/validators/N4JSTypeValidator.xtend
```diff
@@ -518,27 +520,54 @@ class N4JSTypeValidator extends AbstractN4JSDeclarativeValidator {
...
+				AssignmentExpression: {
+					val lhs = container.lhs;
+					switch (lhs) {
+						IdentifierRef: lhs.idAsText
+						ParameterizedPropertyAccessExpression: lhs.propertyAsText
+						default: "the receiving object"
+					}
+				}
+				Argument: "the receiving parameter"
+				default: "the receiving object"
```
mor-n4: In the fall back case you could consider a wording like:
`propUnknown is not defined in ~C; it will not be accessible via type ~C.`

Maybe that's more clear than "the receiving object"? (not sure)

lsp-clients/n4js-vscode-extension/package.json
```diff
@@ -1,9 +1,9 @@
 {
 	"name": "n4js-vscode-extension",
-	"displayName": "N4JS VS Code extension",
-	"description": "LSP server extension for N4JS",
-	"version": "0.0.1",
-	"publisher": "eclipse/n4js",
+	"displayName": "N4JS VSCode extension",
```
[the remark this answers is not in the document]
mmews-n4: I wanted to use only one version.

mmews-n4:
Thanks for all the feedback!

@dbo Please still have a look when you can!

Issue #1626, opened 2020-01-08
Access to Class#n4type should be removed from types defined in n4jsd files


mmews-n4:
... unless these types do declare the annotation `@N4JS` (or state `extends N4Object` if possible at all).

Consider the following example:

**Library.n4jsd**
```ts
export external public class Library {}
```

**Main.n4js**
```ts
import {Library} from "Library";

// FIXME XPECT errors "Couldn't resolve reference to IdentifiableElement 'n4type'." at "n4type"
Library.n4type;

let lib = new Library();
// FIXME XPECT errors "Couldn't resolve reference to IdentifiableElement 'n4type'." at "n4type"
lib.constructor.n4type;

// note that the supertype of Library is not N4Object (as expected)
let n4obj: N4Object;
// XPECT errors "Library is not a subtype of N4Object." at "lib"
n4obj = lib;


// same goes for the constructor
let n4ctor: type{? extends N4Object};
// XPECT errors "constructor{? extends Library} is not a subtype of type{? extends N4Object}." at "lib.constructor"
n4ctor = lib.constructor;
```

mor-n4:
From a quick look, this seems to be caused by outdated code in the scoping:
https://github.com/eclipse/n4js/blob/f8f91a667156907388908fa17fc891e3e54bedd2/plugins/org.eclipse.n4js/src/org/eclipse/n4js/scoping/members/MemberScopingHelper.xtend#L342-L344

The if-condition is incorrect. Also, `RuleEnvironmentExtensions#collectAllImplicitSuperTypesOfType(RuleEnvironment, Type, RecursionGuard<Type>)` should probably be used there to avoid duplication of logic.

PR #1638, opened 2020-01-30
GH-1626: Access to Class#n4type should be removed from types defined in n4jsd files

mor-n4:
See #1626.


mor-n4:
This cannot be merged yet because of an open issue:

Classes in N4JS extending `Error` are not subclasses of `N4Object` but the transpiler creates an `n4type` reflection-getter for them. This PR would disallow accessing `n4type` on such classes, which would break certain client projects using `n4type` on custom error classes.

```javascript
class MyError extends Error {}

let n4obj: N4Object;
n4obj = new MyError(); // error "MyError is not a subtype of N4Object." (makes sense)

console.log(((MyError as Object) as type{N4Type}).n4type.fqn); // prints "Temp4/MyError"
// (so MyError does actully have an n4type getter)
```

mor-n4:
The open issue mentioned above was solved by changing the N4JS code base that was using the `n4type` getter on a subclass of `Error`.

mor-n4:
Builds green, merging ...

PR #1646, opened 2020-02-06
[#1645] Support for named groups in regex validator

### LINKED ISSUE #1645: As a developer, I want to use named capture groups in regular expressions
The named groups were added recently to the JS regular expression semantics. So far, the validation rejects these as being invalid.

szarnekow:
closes #1645


szarnekow:
The review should probably focus on the changes in the grammar and the tests. The other stuff is generated code due to newer EMF versions since the last time this language was generated.

plugins/org.eclipse.n4js.regex/build.properties
```diff
@@ -6,3 +6,4 @@ bin.includes = model/,\
                .,\
                plugin.xml,\
                about.html
+additional.bundles = com.ibm.icu
```
mmews-n4: why is this necessary?

mmews-n4: Due to newer EMF version?

szarnekow: com.ibm.icu is required to generate the language infrastructure. Using this mechanism here ensures that we do not introduce a real code dep. by accident

szarnekow:
> * this feature is quite new and not supported by all engines and browsers (e.g. Firefox); please consider adding a warning.

The validation for regular expression literals was improved. Please take a second look.

> * you had the idea of type system support for named capture groups. Please create a user story for that.

Done https://github.com/eclipse/n4js/issues/1653

mor-n4:
Very nice! Thanks!

One minor thing I noticed: the region for the warning is 2 characters too short. In the test below, for example, the warning appears on `<firs` and `<secon`.

Also, you might want to add a test for named capture groups:
```javascript
/* XPECT_SETUP org.eclipse.n4js.spec.tests.N4JSSpecTest END_SETUP */

// XPECT warnings ---
// "Named capture groups are not supported on all platforms." at "<first>"
// "Named capture groups are not supported on all platforms." at "<second>"
// ---
const p = /aaa(?<first>[0-9]*)bbb(?<second>[0-9]*)ccc/;

const m = p.exec('aaa111bbb222ccc');
console.log('first:', m.groups.first);
console.log('second:', m.groups.second);


/* XPECT output ---
<==
stdout:
first: 111
second: 222
stderr:
==>
--- */
```

:+1:

szarnekow:
Thank you for testing. Maybe it's an off-by-one error with the error marker. It should start at the '<' and so it did for me in VSCode IIRC.

What's wrong with these tests? What are you missing? https://github.com/eclipse/n4js/pull/1646/files#diff-3bd053e00c8d64c6e9337ab9e864d75e

mor-n4:
> What's wrong with these tests? What are you missing?

Sorry, I had overlooked those. Nothing missing then.

szarnekow:
I updated the error handling. Thanks @mor-n4 for catching the bug. It revealed even more problems when working with escape sequences and offsets.

Issue #1650, opened 2020-02-07
As a developer I want to remove the dependency from mangelhaft-cli to n4js-cli


mmews-n4:
* mangelhaft-cli should not invoke n4js-cli but rather use a given test catalogue
* simplify the cli (nomnom) of mangelhaft-cli accordingly

PR #1651, opened 2020-02-07
GH-1650: As a developer I want to remove the dependency from mangelhaft-cli to n4js-cli

mmews-n4:
closes: #1650


n4js-libs/packages/n4js-mangelhaft-cli/tests/npm-test.sh
```diff
@@ -12,8 +12,12 @@
 set -e
 cd `dirname $0`/TestPrj
 
+echo "Create test catalog"
+../../../n4js-cli/bin/n4jsc.js compile ../../../.. --testCatalog ./test-catalog.json
```
szarnekow: Should we have a way to pass the path to n4jsc.js as an env variable?

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/concurrent/XRequestManager.java
```diff
@@ -22,6 +22,10 @@
 import com.google.inject.Singleton;
 
 /**
+ * Handles requests from the LSP client.
```
szarnekow: I don't see the synchronization here apart from the comment

mmews-n4: The synchronization was done already before in commit 8e74fc218c6da26e56288fe271edce2aed063478

n4js-libs/packages/n4js-mangelhaft-cli/tests/npm-test.sh
mmews-n4: I agree and actually it is already available as `n4jsc` (like `mangelhaft-cli`), but I wanted to do it similar to the following call to mangelhaft.

PR #1660, opened 2020-02-11
[#1659] Port N4JS quickfixes to LSP code actions 

### LINKED ISSUE #1659: As a developer I want to have most complex quickfixes migrated to LSP server

szarnekow:
* Provide a generic means to support multifixes.
* Port exemplary quickfixes to code actions
* Minor refactorings to existing code

closes #1659


plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/XLanguageServerImpl.java
```diff
@@ -348,24 +371,6 @@ protected ServerCapabilities createServerCapabilities(InitializeParams params) {
 	@Override
 	public void initialized(InitializedParams params) {
 		clientInitialized.complete(params);
-		clientInitialized.join();
```
szarnekow: The join call here was surprising me given that the future was completed in the line above on the same thread.
Since we want to start something immediately after the client was initialized, I figured that a `.thenRun(..)` on the future might be the reasonable thing to do.

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/XLanguageServerImpl.java
```diff
@@ -185,11 +187,7 @@
 
 	private InitializeResult initializeResult;
 
-	private final CompletableFuture<InitializedParams> clientInitialized = new CompletableFuture<>();
-
-	private final CompletableFuture<?> initBuildFinished = new CompletableFuture<>();
-
-	private final CompletableFuture<?> initCleanFinished = new CompletableFuture<>();
+	private final CompletableFuture<InitializedParams> clientInitialized;
```
szarnekow: I removed these fields here since we use the same server instance on a clean-rebuild request. These two futures here would already be fulfilled at that point in time. So the methods to join on these were misleading.

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/XLanguageServerImpl.java
```diff
@@ -540,17 +543,9 @@ public void didChangeWatchedFiles(DidChangeWatchedFilesParams params) {
...
-				workspaceManager.clean(CancelIndicator.NullImpl);
-				initCleanFinished.complete(null);
-			} catch (Throwable t) {
-				t.printStackTrace();
-				initCleanFinished.completeExceptionally(t);
-			}
-
+	public CompletableFuture<Void> clean() {
+		return requestManager.runWrite("clean", () -> {
+			workspaceManager.clean(CancelIndicator.NullImpl);
```
szarnekow: Here I removed the completion of the no-longer available fields.

plugins/org.eclipse.n4js.cli/src/org/eclipse/n4js/cli/compiler/N4jscCompiler.java
```diff
@@ -117,7 +116,7 @@ private void performCompile() {
 		}
 		Stopwatch compilationTime = Stopwatch.createStarted();
 		languageServer.initialized(new InitializedParams());
-		languageServer.joinInitBuildFinished();
+		languageServer.getRequestManager().runRead("joinInitBuildFinished", (ci) -> null).join();
```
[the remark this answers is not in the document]
szarnekow: The method `joinInitBuildFinished` was misleading.
Maybe we should add a more general method to our test infra that says: Block, until the server has processed all requests.

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/server/codeActions/N4JSCodeActionService.java
```diff
@@ -10,29 +10,187 @@
...
+import com.google.common.base.Strings;
+import com.google.common.collect.HashMultimap;
+import com.google.common.collect.Multimap;
 import com.google.inject.Inject;
+import com.google.inject.Injector;
+import com.google.inject.Singleton;
 
 /**
  *
  */
```
szarnekow: TODO documentation

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/server/codeActions/N4JSCodeActionService.java
```diff
@@ -10,29 +10,187 @@
...
+	 * with {@link Fix}.
+	 */
+	@Inject
+	public void init(Injector injector) {
+		for (Class<?> quickFixProvider : quickfixProviders) {
+			Object qpInstance = injector.getInstance(quickFixProvider);
+			Method[] methods = quickFixProvider.getMethods();
+			for (Method method : methods) {
+				int modifiers = method.getModifiers();
+				if (Modifier.isPublic(modifiers) && !Modifier.isStatic(modifiers)) {
```
szarnekow: TODO add check for valid parameter types of the method

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/server/commands/N4JSCommandService.java
```diff
@@ -10,41 +10,249 @@
...
+				messageParams.setType(MessageType.Warning);
+				break;
+			case IStatus.ERROR:
+				messageParams.setType(MessageType.Error);
+				break;
+			default:
+				return null;
+			}
+
+			StringWriter sw = new StringWriter();
```
szarnekow: TODO message is ugly in VS code. Html did not work. Check what VSCode accepts to render a pretty message.

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/ExecuteCommandParamsDescriber.java
```diff
@@ -8,12 +8,19 @@
...
- * Marker interface for classes that contain quick-fix methods. Quick-fix methods must be annotated with {@link Fix}
- * (Mind the information in {@link Fix}).
+ * Return the argument types for the available commands.
  */
-public interface IQuickfixProvider {
-	// marker interface
+public interface ExecuteCommandParamsDescriber {
+
+	/**
+	 * Command id to argument types.
```
szarnekow: Add some info what these argumentTypes are, cf ExecuteCommandParams.getArguments()

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/XExecutableCommandRegistry.java
```diff
@@ -0,0 +1,47 @@
...
+import java.util.Map;
+
+import org.eclipse.xtext.ide.server.commands.ExecutableCommandRegistry;
+import org.eclipse.xtext.ide.server.commands.IExecutableCommandService;
+import org.eclipse.xtext.xbase.lib.util.ReflectExtensions;
+
+import com.google.common.collect.Multimap;
+
+/**
+ *
```
szarnekow: Documentation

mmews-n4:
I remember and was not sure myself. But as far as I could see, the supplier was either triggered immediately (for the single quickfix) or not returned for multi-quickfixes (whole file/project). To put it this way: I did not see that the supplier was necessary. I checked this by adding a `println()` inside method `N4JSQuickfixProvider#transformJavaTypeAnnotationToColonStyle()`. This was not triggered multiple times when hovering or executing the single-time-version of the quickfix. Only if the multi-quickfix was selected, the printout was done on the console for all the necessary locations. - Maybe I missed something?

szarnekow:
IIRC:
Suppose you have multiple fixes for the same issue code. If a fix is to be applied to the entire project / file, we need to find the correct text edits. The selection of a quickfix method is solely possible based on the issue code. Afterwards, we need to match against the title of the quickfixes. Let's say it is expensive to compute the text edits but cheap to compute the title. In that case, we want to defer the computation of the text edits to avoid computing edits for the fixes that do not match the given title.

mmews-n4:
How can I check with a `System.out.println("something")` that the expensive text edit computation is done only when necessary?

szarnekow:
You could try to duplicate the quickfix for transformJavaTypeAnnotationToColonStyle with another title but the same issue code.

mmews-n4:
I'm standing on a hose (ger: "Ich stehe auf dem Schlauch.")

Let me go this through. We have 3 subclasses of `ICodeActionAcceptor`:
`MultiQuickfixAcceptor`: delegates to
`CodeActionAcceptor`: returns the computed textedits for the single-quickfix already (with or without supplier). For all the other occurrences, changes are never computed (also no suppliers were computed before).
`TextEditCollector`: is used only after a user selects a multi-quickfix. Then, all textedits are computed.

> You could try to duplicate the quickfix for transformJavaTypeAnnotationToColonStyle with another title but the same issue code.

- Regarding the issue location for the quickfix: both of the textedits are computed for that location already. For the multi-quickfixes (for all the other locations), only commands are created.

mmews-n4:
Thanks a lot for your feedback! Merging.

PR #1662, opened 2020-02-13
GH-1661: n4js-cli does not always detect correct version of installed JRE

### LINKED ISSUE #1661: n4js-cli does not always detect correct version of installed JRE
At the moment, n4js-cli uses the return string of the call to `java -version` to figure out the JRE version. This is error prone since these returns string do not share a common format.

Goal here is to improve the logic to always return the correct JRE version.

mmews-n4:
closes #1661


n4js-libs/packages/n4js-cli/bin/PrintJavaVersion.java
```diff
@@ -0,0 +1,18 @@
...
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   NumberFour AG - Initial API and implementation
+ */
+
+/** Utility to detect and print out the Java version */
+public class PrintJavaVersion {
+	public static void main(String[] args) {
+		String versionStr = System.getProperty("java.version");
```
szarnekow: This could also be dependent on the JRE / JDK impl.
You may want to use `java.lang.Runtime.version().feature()`

mmews-n4: Ok, Thanks!

n4js-libs/packages/n4js-cli/package.json
```diff
@@ -18,6 +18,7 @@
   "files": [
     "bin/n4jsc.js",
     "bin/n4jsc.jar",
+    "bin/PrintJavaVersion.java",
```
mor-n4: It's confusing to have Java and Javascript source files in the `bin` folder.

Since the `src` folder already contains a sub-folder `n4js` which contains the N4JS sources, it would be best to create two sub-folders `js` and `java` below folder `src` and place those files there. I think this is then in line with the structure of other n4js-libs and N4JS code bases.

Also, `bin` folders in npm packages are often used for binaries that may make sense to be used directly. This applies to the JRE and the n4jsc.jar, but not the .js and .java files.

Issue #574, opened 2018-02-07
As a developer I want to limit reflection information on classes in the output code


mor-n4:
(moved from internal bug tracker, IDEBUG-412 and NumberFour/n4js#31; original reporter: @dbo)

There is a lot of reflection information emitted by the compiler without any possibility to limit it (e.g. only fields, only methods, just methods without param types, ...).

We should be careful not to bloat the compiled output without need, it slows down startup and it makes debugging harder.

Actually we need a solid concept what kind of reflection we provide or may not provide. I'm afraid simply adding an annotation `@noreflection` won't do it – what about super and sub-classes then, how can we validate the usage, are there other impacts in generated code etc.

mor-n4:
From a meeting with @mmews-n4 and @mor-n4:

As part of this task, it should be investigated if there might be an alternative to completely turning off reflection information at compile time (by way of an annotation or a property in the package.json): maybe, reflection information can be stored in a way such that existing, ordinary JS optimization tools that analyze code to remove unused modules (e.g. webpack) are able to remove the unused reflection information.

Possible approach: reflection information could be generated into a separate file (for each module) and the transpiler will then transpile the access to the reflection information (e.g. `MyClass.n4type`) to an access to the information in this separate file (including a corresponding implicit import).

However, it has to be considered how to deal with generic reflection access:
1. actual type not known:
   ```javascript
   let cls: type{MyClass};
   // ...
   cls.n4type; // might access reflection of any subclass of MyClass
   ```
2. reflection access is entirely generic by providing reflection information for any fully qualified name, etc. (if that is even supported).

A possible solution to such generic reflection access might be to channel it through a single implicit module that is created for every project, so at least reflection information could be removed by optimization tools as long as there is no generic access to reflection information of a particular project.

mmews-n4:
This ticket focusses on checking whether there is unnecessary information in the reflection output that could be removed, and whether the reflection information could be represented more compactly in the output code. Also consider adjusting pretty-printer to allow even more compact formatting of some parts of the output code (to improve readability). Turning reflection off would be handled in a different ticket.

Alternatives regarding compact reflection representation:
1. plain object literals instead of "new N4Class" etc. in output code; instances of N4Class will be created by utility function in runtime
2. plain object literals (as above); they will be used directly, instances of N4Class will never be created, N4Class will be changed to structural interface
3. as 1. or 2.: object literals will be represented as json string literals in output code

Use some kind of realworld scenario performance test to evaluate alternatives (within reason).

PR #1683, opened 2020-03-12
GH-574B: As a developer I want to limit reflection information on classes in the output code

mmews-n4:
Second part of #574 
Migration of reduced reflection information for ES6 branch


mmews-n4:
continued here: #1693

Issue #1691, opened 2020-03-17
As a smith I want to improve "add missing imports" functionality in code completion and quick fixes


mor-n4:
Since we decided to not add missing imports during "organize imports" (at least for the time being, see #1677 for rationale), imports are only added by way of code completion and quick fixes.

The main functionality is already available, but there are some problems and it would be a good opportunity for an overall review of the current functionality and its implementation. Also, tests have to be added for this.

Problems and missing features encountered, so far:

Given two classes with the same name `Duplicate` in two different modules and a third module `Main` without any imports, then
1. code completion inside `Main` will only show a single proposal when being triggered on the prefix `Dupli`.
2. when adding the unresolved reference `Duplicate` to `Main`, then the quick fixes for adding one of the two possible imports can be reached via the tool tip shown when hovering over the error message, but not via the keyboard short cut for showing quick fixes, i.e. `cmd` + `.`.
3. code completion does not work properly when executed for an element E which is not imported yet, in case a namespace import for the module containing E already exists.
4. multi-apply not supported yet by quick fixes for adding missing imports.

### ACs

1. check if there are more missing things related to "add missing imports" functionality,
2. solve problems described above and found in 1.,
3. review implementation and refactor if necessary,
4. add tests for all cases of "add missing imports" (via code completion or quick fixes).

This task is time boxed. Main motivation is to provide the base functionality. Create follow-up tasks as appropriate.

mor-n4:
Some remarks derived from looking through the old organize imports tests (because the old organize imports was adding missing imports, many tests are related to this user story):

* Optional functionality for adding missing imports (maybe for follow-up tasks):
  * if applicable, give the option to add a project import instead of an ordinary import
  * support add missing namespace imports
    For example: if `N.SomeClass` is given in code and N is unresolved and "SomeClass" exists in a module "M", then a quick fix could propose to add `import * as N from "M"` as import.

* General remarks related to implementation:
  * make sure all forms of adding missing imports works for unresolved IdentifierRefs as well as unresolved TypeRefs (all kinds of type refs, e.g. structural, union/intersection, TypeTypeRefs, etc.). When looking at the old tests, it seems this was a problem before (probably due to internal implementation differences between these cases).
  * when searching correct place to insert a new import in a module that does not yet have imports:
    * consider preamble comments such as copyright headers
      (consider that the top-most declaration might have an API documentation which must not be mistaken for a preamble comment!)
    * consider module-level annotations (must not add import before those!)
[...]

PR #1694, opened 2020-03-20
GH-1691A: As a smith I want to improve "add missing imports" functionality in code completion and quick fixes

mmews-n4:
See #1691.

### DIFF
--- a/tests/org.eclipse.n4js.ide.tests/src/org/eclipse/n4js/ide/tests/contentassist/CompletionWithImportsTest.xtend
+++ b/tests/org.eclipse.n4js.ide.tests/src/org/eclipse/n4js/ide/tests/contentassist/CompletionWithImportsTest.xtend
@@ -27,11 +27,9 @@ public class CompletionWithImportsTest extends AbstractCompletionTest {
 			"MBA" -> '''
 						export class B1 {}
 						export class A2 {}''',
-			"C"   -> '''
-						export class C01 {}
-						export class C02 {}
-						export class C03 {}
-						export class C04 {}''',
+			"MC"   -> '''
+						export class C1 {}
+						export class C2 {}''',
 			"Def" -> '''
 						export class Def01 {}
 						export class Def02 {}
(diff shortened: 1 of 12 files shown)


tests/org.eclipse.n4js.ide.tests/src/org/eclipse/n4js/ide/tests/contentassist/CompletionWithImportsTest.xtend
```diff
@@ -0,0 +1,262 @@
...
+						export class A2 {}'''];
+	}
+
+	@Test
+	def void test01() {
+		test('''
+			let x = new A1<|>
+		''', ''' 
+			(A1, Class, MA, , , 00000, , , , ([0:12 - 0:14], A1), [([0:0 - 0:0], import {A1} from "MA";
+			)], [], , )
```
mor-n4: What about applying the changes sent by the server and then using the resulting changed code for comparison (to make the test more readable)?
If `TextEdit`s are sent, here, then there is already method `AbstractIdeTest#applyTextEdits()` that might help.

mor-n4:
Two things I noticed (but they are more a problem of content assist in general, not the "add missing import"):

```javascript
import {A02} from "Temp";

A02
A01/* here, content assist does not work (caused by missing ; in line above) */
```

```javascript
import * as N from "Temp";
N.A01
N./* here, content assist offers only A01, not A02, A03, etc. */
N.A/* here it works as expected */
```
(module `Temp` exports classes `A01` ... `A04`)

tests/org.eclipse.n4js.ide.tests/src/org/eclipse/n4js/ide/tests/contentassist/CompletionWithImportsTest.xtend
mmews-n4: Thanks, good idea, but I also want to validate the label, icon, description besides the actual replacement string.

mmews-n4:
Regarding the "two things": I noticed that too, but think it is out of scope. I created a new ticket, however.

mmews-n4:
Thanks! Builds green, merging part A.

Issue #1696, opened 2020-03-26
Cast expression with syntax error throwing null-pointer exception


mor-n4:
To reproduce:
```javascript
const c = 42;
(c as )
```
Exception thrown:
```
Caused by: java.lang.NullPointerException
	at org.eclipse.emf.ecore.util.Switch.doSwitch(Switch.java:69)
	at org.eclipse.n4js.typesystem.BoundJudgment.applyUpperBound(BoundJudgment.java:56)
	at org.eclipse.n4js.typesystem.N4JSTypeSystem.upperBoundWithReopen(N4JSTypeSystem.java:215)
	at org.eclipse.n4js.validation.validators.N4JSExpressionValidator.checkCastExpression(N4JSExpressionValidator.java:1492)
	at jdk.internal.reflect.GeneratedMethodAccessor34.invoke(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.eclipse.xtext.validation.AbstractDeclarativeValidator$MethodWrapper.invoke(AbstractDeclarativeValidator.java:127)
	at org.eclipse.n4js.validation.N4JSValidator$N4JSMethodWrapperCancelable.invoke(N4JSValidator.java:113)
	... 42 more
```

PR #1697, opened 2020-03-26
GH-1696: Cast expression with syntax error throwing null-pointer exception

mor-n4:
See #1696.


mor-n4:
Thanks!
Builds green, merging ...

Issue #1698, opened 2020-03-26
As a smith I want to use performance data collectors with the headless LSP builder


mor-n4:
The performance data capturing based on class `DataCollector` is not yet fully functional in the LSP case (but the command line interface is already in place). Also, the data collectors and the places where they are turned on/off needs an overall review and clean up (for example, some data collectors defined in `N4JSDataCollectors` are unused and probably obsolete).

This can be used as an opportunity to implement some simple convenience functionality.

### ACs

1. test performance reporting with the LSP-based headless builder,
2. review all data collectors defined in `N4JSDataCollectors` and make sure they are applied at the correct locations in the code (if necessary, find better places),
3. overall clean up.

PR #1699, opened 2020-03-26
GH-1698: As a smith I want to use performance data collectors with the headless LSP builder

mor-n4:
See #1698.


plugins/org.eclipse.n4js.cli/src/org/eclipse/n4js/cli/N4jscOptionsValidater.java
```diff
@@ -174,10 +174,18 @@ private static void validateTestCatalogFile(N4jscOptions options) throws N4jscEx
 		}
 	}
 
-	private static void validatePerformanceReport(N4jscOptions options) throws N4jscException {
-		if (options.getPerformanceReport() != null && Strings.isNullOrEmpty(options.getPerformanceKey())) {
+	private static void validatePerformanceOptions(N4jscOptions options) throws N4jscException {
+		if (options.getPerformanceReport() == null) {
+			String msg = "Missing performance report.";
+			throw new N4jscException(N4jscExitCode.OPTION_INVALID, msg);
```
mmews-n4, line 180: Do we have a test for this (if possible)?

mor-n4: A value of `null` cannot be produced at the moment. I only added this (1) for consistency with the case "Missing performance key" right below this if-block and (2) in case we change the default in the future. So we cannot create a test for this, for now.

mor-n4:
Builds green, merging ...

Issue #1693, opened 2020-03-18
As a smith I want to improve reduced reflection feature to es6 classes


mmews-n4:

mor-n4:
It seems `origin` is not set in all elements, even though the property is declared in `N4Element` in file `builtin_n4.n4ts`. E.g.
```javascript
class C {
	m() {}
}
console.log(C.n4type.members()[0].origin); // prints "undefined"
```

mor-n4:
Reflection has to omit some implicit members created by the transpiler. Make sure this is handled consistently for all such members (probably the tag `hiddenFromReflection` in class `InformationRegistry` should be used in all those cases).

mor-n4:
I have rebased the contents of branch `GH-574B` onto the latest master. The result can be found on branch `GH-1693`. Branch `GH-574B` still exists, for reference.

PR #1705, opened 2020-03-31
GH-1693: As a smith I want to improve reduced reflection feature to es6 classes

mmews-n4:
see #1693


n4js-libs/packages/n4js-runtime/src/n4js/N4BuiltInTypes.n4js
```diff
@@ -70,7 +70,7 @@ abstract class N4NamedElement extends N4Element {
 	public name: string;
 }
 
-abstract class N4Type extends N4NamedElement {
+export project abstract class N4Type extends N4NamedElement {
```
mor-n4, line 73: At the top of this file there is a comment saying:
```
// This has the following consequences:
// 1) neither "@@Global" nor "export" must be used in this file,
```
Since you are now exporting all the classes, change the comment at the top of file to:
```
// This has the following consequences:
// 1) neither "@@Global" nor "export" with an accessibility higher than "project" must be used in this file
//   (to ensure only the implementations, not the types declared below, are available to ordinary N4JS code),
```

n4js-libs/packages/n4js-runtime/src/n4js/ReflectionBuilder.n4js
```diff
@@ -0,0 +1,437 @@
...
+
+const SYMBOL_IDENTIFIER_PREFIX = "#";
+const INTERFACE_FIELD_INITS = "$fieldInits";
+const INTERFACE_DEFAULT_MEMBERS = "$defaultMembers";
+const INTERFACE_EXTENDS = "$extends";
+const CLASS_IMPLEMENTS = "$implements";
+const PROPERTY_N4TYPE = "n4type";
+
+
+export function $interfaceFieldInits(obj: any+) : Object {
```
mor-n4, line 23: please remove the leading `$` from all functions that are only used inside the runtime and are not made available to the generated output code created by the transpiler.

I think it applies to:
```
$interfaceFieldInits
$interfaceDefaultMembers
$interfaceExtends
$classImplements
```
(but please double check if there are more, also in file `Runtime.n4js`)

n4js-libs/packages/n4js-runtime/src/n4js/ReflectionBuilder.n4js
```diff
@@ -0,0 +1,437 @@
...
+ * All rights reserved. This program and the accompanying materials
+ * are made available under the terms of the Eclipse Public License v1.0
+ * which accompanies this distribution, and is available at
+ * http://www.eclipse.org/legal/epl-v10.html
+ *
+ * Contributors:
+ *   NumberFour AG - Initial API and implementation
+ */
+
+import {N4Type, N4Classifier, N4Class, N4Interface, N4EnumType, N4Annotation, N4DataField, N4Accessor, N4Method, N4Member} from 'N4BuiltInTypes';
```
mor-n4, line 12: Or maybe a namespace import? Would also make the "cast from N4Class to N4Class" situation a bit more understandable.

mmews-n4: My first though was "cool" and I applied the change. But I had to change a lot of places and since this would be undone when we move the Builtin types to n4js-runtime, I think I rather leave it like this. Still, I added a comment to the N4Class cast.

mmews-n4:
Builds green. Thanks for your feedback! Merging.

Issue #1701, opened 2020-03-30
Null pointer exceptions in build log of nightly build


mor-n4:
In the log of the N4JS nightly build, many exceptions of type `NullPointerException` show up, even though the build is green.

For example:
```
Caused by: java.lang.NullPointerException: platform:/resource/com.enfore.opr.base/src/n4js/n4/i18n/i18nDateAndTime.n4js
	at com.google.common.base.Preconditions.checkNotNull(Preconditions.java:985)
	at org.eclipse.n4js.projectModel.locations.FileURI.validate(FileURI.java:58)
	at org.eclipse.n4js.projectModel.locations.SafeURI.<init>(SafeURI.java:51)
	at org.eclipse.n4js.projectModel.locations.FileURI.<init>(FileURI.java:38)
	at org.eclipse.n4js.internal.FileBasedWorkspace.safeFromURI(FileBasedWorkspace.java:59)
	at org.eclipse.n4js.internal.FileBasedWorkspace.fromURI(FileBasedWorkspace.java:55)
	at org.eclipse.n4js.internal.FileBasedWorkspace.fromURI(FileBasedWorkspace.java:1)
	at org.eclipse.n4js.internal.N4JSModel.fromURI(N4JSModel.java:141)
	at org.eclipse.n4js.internal.N4JSModel.findProjectWith(N4JSModel.java:129)
	at org.eclipse.n4js.internal.N4JSRuntimeCore.findProject(N4JSRuntimeCore.java:96)
	at org.eclipse.n4js.utils.N4JSLanguageHelper.isOpaqueModule(N4JSLanguageHelper.java:113)
	at org.eclipse.n4js.resource.N4JSResource.isOpaque(N4JSResource.java:340)
	at org.eclipse.n4js.resource.N4JSResource.doLoad(N4JSResource.java:677)
	at org.eclipse.emf.ecore.resource.impl.ResourceImpl.load(ResourceImpl.java:1563)
	at org.eclipse.emf.ecore.resource.impl.ResourceImpl.load(ResourceImpl.java:1342)
	at org.eclipse.xtext.resource.persistence.StorageAwareResource.load(StorageAwareResource.java:79)
	at org.eclipse.n4js.resource.N4JSResource.superLoad(N4JSResource.java:577)
	at org.eclipse.n4js.resource.N4JSResource.load(N4JSResource.java:859)
[...]

PR #1710, opened 2020-04-06
GH-1701: Null pointer exceptions in build log of nightly build

mor-n4:
See #1701.


mor-n4:
Builds green, merging ...

PR #1717, opened 2020-04-14
GH-1716: Project dependency not recognised when located inside a project node_modules folder of a yarn project only

### LINKED ISSUE #1716: Project dependency not recognised when located inside a project node_modules folder of a yarn project only
Given there is a project `P` in a yarn workspace setup (i.e. in the workspaces project) and `P` has a nested node_modules folder which contains a dependency `D` of `P`.  Then `D` is not recognised iff it is only inside the node_modules folder `P` (that means it must not be located in the node_modules folder of the yarn workspace project.)

mmews-n4:
closes #1716


plugins/org.eclipse.n4js/src/org/eclipse/n4js/utils/NodeModulesDiscoveryHelper.java
```diff
@@ -42,14 +44,25 @@
 	/** Defines whether a node_modules folder is a child of an NPM project or a YARN workspace project */
 	public static class NodeModulesFolder {
 		/** Location of a node_modules folder */
-		final public File nodeModulesFolder;
-		/** True iff the {@link #nodeModulesFolder} is in a yarn workspace */
-		final public boolean isYarnWorkspace;
+		final public File npmNodeModulesFolder;
+		/** Location of a node_modules folder */
+		final public File yarnNodeModulesFolder;
```
mor-n4, line 49: I don't understand the difference between `npmNodeModulesFolder` and `yarnNodeModulesFolder` (also the identical API doc does not help 😄 ). Yarn is compatible with npm, so all node_modules folders used by npm are also used by yarn (though not vice versa). What about `localNodeModulesFolder` (for those in each project) and `globalNodeModulesFolder` (for the one in the root folder of the yarn workspace)?

plugins/org.eclipse.n4js/src/org/eclipse/n4js/validation/validators/packagejson/N4JSProjectSetupJsonValidatorExtension.xtend
```diff
@@ -1405,7 +1410,7 @@ public class N4JSProjectSetupJsonValidatorExtension extends AbstractPackageJSONV
 						val NodeModulesFolder nmFolder = nodeModulesDiscoveryHelper.getNodeModulesFolder(projectPath);
 						if (nmFolder !== null) {
 							val projectName = new EclipseProjectName(project.name).toN4JSProjectName
-							res.put(projectName, nmFolder.nodeModulesFolder);
+							res.putAll(projectName, nmFolder.nodeModulesFolders);
```
mor-n4, line 1413: The map will contain the global node_modules folder (the one in the root folder of the yarn workspace) many times: 1 + n (if there are n projects in the yarn workspace). Does this make sense?

plugins/org.eclipse.n4js/src/org/eclipse/n4js/validation/validators/packagejson/N4JSProjectSetupJsonValidatorExtension.xtend
```diff
@@ -1166,8 +1166,13 @@ public class N4JSProjectSetupJsonValidatorExtension extends AbstractPackageJSONV
 				return;
 			}
 
-			val currNodeModulesFolder = allNodeModuleFolders.get(currentProjectName)?.toPath;
-			val currNPM = currNodeModulesFolder?.resolve(id.rawName);
+			var Path currNPM = null;
+			for (File nodeModulesFolder : allNodeModuleFolders.get(currentProjectName)) {
+				if (currNPM === null || !currNPM.toFile.exists) {
+					currNPM = nodeModulesFolder.toPath.resolve(id.rawName);
+				}
```
mor-n4, line 1173: This looks for the npm package `id` in both the local and global node_modules folders. It seems this check should be performed in a well-defined order, to always obtain the same result in case the project exists in both the local and global node_modules folders. Even if we used a LinkedHashMap (which we don't, I think?) it seems to me it would be good to make this more explicit.

plugins/org.eclipse.n4js/src/org/eclipse/n4js/validation/validators/packagejson/N4JSProjectSetupJsonValidatorExtension.xtend
mmews-n4: This is for caching purposes only since the related yarn/npm node_modules folders should not be computed everytime one of the six check methods is called. Regarding memory there might be a smarter way to cache it, I agree.

mmews-n4:
builds greenish. merging.

Issue #1720, opened 2020-04-16
As a developer I want the transpiler to simplify computed property names


mor-n4:
The transpiler should simplify computed property names in N4JS to ordinary property names in the output code, in case the expression is a string literal.

Background: in certain configurations, Babel et al. turn object literals with computed property names into very complex code for compatibility reasons. So it is good to avoid computed property names as far as possible.

PR #1722, opened 2020-04-16
GH-1720: As a developer I want the transpiler to simplify computed property names

mor-n4:
See #1720.


mor-n4:
Builds green, merging ...

PR #1730, opened 2020-04-21
GH-1700: n4jsc in n4js-cli returning exit code 0 even in case of compile errors

### LINKED ISSUE #1700: n4jsc in n4js-cli returning exit code 0 even in case of compile errors
The command line tool `n4jsc` contained in npm package `n4js-cli` returns an exit code of 0 even though there are compile errors in the N4JS source code being compiled.

To reproduce:

1. create a new, empty N4JS project.
2. add dev-dependency `n4js-cli` to package.json:
   ```json
   "devDependencies": {
   	"n4js-cli": "0.19.6"
   },
   ```
3. add a file to the source folder with a compile error:
   ```javascript
   const c: string = 42;
   ```
4. in bash, go to the root folder of the newly created N4JS project and run
   ```
   $ ./node_modules/.bin/n4jsc compile .
   ```
   which will report the compile error. When printing the exit code with
   ```
   $ echo $?
   ```
   we get `0` instead of the expected non-zero value.

### LINKED ISSUE #1723: As a smith I want to also publish n4js-cli and vscode extension in cases where only the n4jsc.jar changes (not n4js-libs)
At the moment, n4js-cli and n4js-vscode-extension are only published when the major.minor.patch version changes, i.e. they are not published when only the build-metadata changes. Consequently, changes that impact only n4jsc.jar wont get published unless the version is manually increased.

mmews-n4:
closes #1700

Note: Accidentally the branch name is GH-1723 (and not GH-1700 which would have been the correct branch name).


n4js-libs/packages/n4js-cli/src/n4js/n4jsc.n4js
```diff
@@ -27,10 +27,11 @@ import {getJavaCommand} from "JreProvider";
...
 export public async function runN4jscSync(log: {function(string)} = npmlog.info) {
     const args = getArgsFromProcess();
-
-    let exitCode : int = await runN4jsc(args, log);
-    if (exitCode !== 0) {
+    try {
+        let exitCode : int = await runN4jsc(args, log);
         process.exit(exitCode);
+    } catch (error) {
+        process.exit(error as int);
```
mor-n4, line 34: In case of unexpected exceptions, `error` might not be of type `int`. Maybe add a `typeof number` check?

plugins/org.eclipse.n4js.cli/src/org/eclipse/n4js/cli/N4jscExitState.java
```diff
@@ -0,0 +1,69 @@
...
+ * instead gracefully. In case the {@link N4jscExitState} is other than {@link #SUCCESS}, its message is printed out and
+ * the process terminates with its exit code (can be other than zero).
+ */
+public class N4jscExitState {
+	/** Singleton for plain success state without message for the user. */
+	final static public N4jscExitState SUCCESS = new N4jscExitState(N4jscExitCode.SUCCESS, true, null);
+
+	final N4jscExitCode exitCode;
+	final boolean suppressUserMessage;
+	final String message;
```
mor-n4, line 24: Since you added getters for those fields, you could make them private?

tests/org.eclipse.n4js.ide.tests/src/org/eclipse/n4js/ide/tests/compiler/N4jscTypeDefinitionsModuleShadowingTest.java
```diff
@@ -52,7 +53,7 @@ public void deleteWorkspace() throws IOException {
 	 */
 	@Test
 	public void testSimpleTypeDefsShadowing() {
-		CliCompileResult cliResult = n4jsc(COMPILE(workspace));
+		CliCompileResult cliResult = n4jsc(COMPILE(workspace), VALIDATION_ERRORS.getExitCodeValue());
```
mor-n4, line 56: `.getExitCodeValue()` not required here, right?

mmews-n4:
Builds green, thanks for feedback! Merging.

Issue #1724, opened 2020-04-20
As a smith I want to know the performance of scoping (i.e. IScope) and content assist


szarnekow:
Scoping of the Xtext framework relies on the methods `IScope#getAllLocalElements` and `IScope#getLocalElementsByName(...)` or `IScope#getLocalElementsByEObject`. Since the latter ones (and others in subclasses of `IScope`) rely on `getAllLocalElements` which returns an `iterable` only, many filter operations have linear time complexity. However, we do not know the actual performance impact of this implementation.

Goal of this user story is to add measurements (see `org.eclipse.n4js.smith.Measurement`) to all relevant scoping methods to estimate the impact of scoping regarding time performance. These measurements should measure both the overall impact and more fine grained measurements of specific methods of `IScope`.

A **follow-up** user story could be overwrite some methods of `IScope` to improve performance, e.g. like using maps as done in `org.eclipse.xtext.scoping.impl.MapBasedScope`.

A second big topic is performance of content assist, which should be measured independently to find out whether scoping or another part of the implementation is consuming much time during a content assist request. These measurements should be implemented in a way that they remain in the code base to be used at a later point in time for re-evaluation (also use `N4JSDataCollectors` if possible).

PR #1732, opened 2020-04-21
GH-1724: Add data collectors to content-assist

szarnekow:
This PR provides insights into the cost of the various steps during proposal computation in the LSP and the Eclipse case.


plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/N4JSIdeDataCollectors.java
```diff
@@ -0,0 +1,107 @@
...
+
+	private static DataCollector create(String key) {
+		return DataCollectors.INSTANCE.getOrCreateDataCollector(key);
+	}
+
+	private static DataCollector create(String key, DataCollector parent) {
+		return DataCollectors.INSTANCE.getOrCreateDataCollector(key, parent);
+	}
+
+	static class SequentializingDataCollector extends ForwardingDataCollector {
```
szarnekow, line 54: Consider promoting this to API in n4js.smith

szarnekow:
@mmews-n4 Please take a look at this PR

plugins/org.eclipse.n4js.ui/src/org/eclipse/n4js/ui/contentassist/CustomN4JSParser.java
```diff
@@ -326,126 +345,114 @@ public void initializeTokenTypes(ContentAssistTokenTypeMapper mapper, N4JSGramma
 		this.postfixGroup = grammarAccess.getPostfixExpressionAccess().getGroup_1();
 	}
 
-	// TODO remove the code below with Xtext 2.13
```
mor-n4, line 329: Thanks for the clean up!

plugins/org.eclipse.n4js.smith/src/org/eclipse/n4js/smith/SerializingDataCollector.java
```diff
@@ -0,0 +1,154 @@
...
+import java.util.Collection;
+import java.util.List;
+import java.util.concurrent.locks.ReentrantLock;
+
+import com.google.common.base.Supplier;
+
+/**
+ * A collector decorator that will ensure that measurements are not created concurrently.
+ */
+class SerializingDataCollector extends DataCollector {
```
mor-n4, line 22: Cool idea to solve the issue with such a wrapper!

But isn't it possible that this can introduce new deadlocks if the code being measured obtains locks on other resources? Not that critical, probably, because this new locking will only be active during debugging, but it's something we have to keep in mind.

plugins/org.eclipse.n4js.smith/src/org/eclipse/n4js/smith/DataCollectors.java
```diff
@@ -91,6 +113,12 @@ synchronized void setPaused(boolean paused) {
 		collectors.values().forEach(collector -> collector.setPaused(paused));
 	}
 
+	/** Stops all data collectors but allows ongoing measurements to complete normally. */
+	synchronized void stop() {
```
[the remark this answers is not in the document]
szarnekow: Makes sense, will do.

plugins/org.eclipse.n4js.smith/src/org/eclipse/n4js/smith/SerializingDataCollector.java
szarnekow: Hmm interesting. I think this could happen, if we'd have re-entrant invocations on the DataCollector (which is discouraged, but possible with the API).
Thread A passes by the disabled collector, aquires some other lock L2.
Thread B enabled the collector, acquires the collectors lock L1 and tries subsequently to work with L2.
Thread A is now reentrant on the collector and waits for L1 to become available. 

The configuration of the collectors does not offer this kind of control flow from the internals of a request back to the outer request manager, but yes, we should keep an eye on this. I'd like to wait a concrete deadlock situation to happen before revisiting the implementation.

Issue #1733, opened 2020-04-21
As a developer I want content assist to work in cases of missing semicolon


szarnekow:
cf #1703

PR #1737, opened 2020-04-22
GH-1733: Fixed lookAhead information in node model and speed up CA

szarnekow:


tests/org.eclipse.n4js.lang.tests/src/org/eclipse/n4js/tests/parser/LookAheadTest.xtend
```diff
@@ -0,0 +1,195 @@
...
+package org.eclipse.n4js.tests.parser
+
+import org.eclipse.n4js.n4JS.ExpressionStatement
+import org.eclipse.n4js.n4JS.NumericLiteral
+import org.eclipse.xtext.nodemodel.util.NodeModelUtils
+import org.junit.Test
+import org.eclipse.n4js.n4JS.IdentifierRef
+import org.eclipse.xtext.nodemodel.ICompositeNode
+
+class LookAheadTest extends AbstractParserTest{
```
mmews-n4, line 20: Would you add some insight what the test is about? Thing is that the look ahead implementation is not one of our strong suits and in case one of these tests fails, we would appreciate a hint where to look.

releng/org.eclipse.n4js.releng.utils/src/org/eclipse/n4js/antlr/n4js/AutomaticSemicolonInjector.java
```diff
@@ -62,7 +62,7 @@ private String addCustomMethodsToParser(String grammarContent) {
 		return replace(grammarContent, "@members {",
 				"@members {\n\n" +
 						"// injected by AutomaticSemicolonInjector\n" +
-						"protected boolean forcedRewind(int marker) { return true; } // overridden in subtype\n" +
+						"protected boolean forcedRewind(int marker, boolean advance) { return true; } // overridden in subtype\n" +
```
mmews-n4, line 65: 😨

szarnekow: It was a shortcut. I'll make this two methods :)

mmews-n4: Actually my comment was related to the idea that source code is modified with string replace. I did not see that before in our code base and was mildly shocked.

szarnekow: It's a beauty, I know ;)

Nowadays we could customize the grammar templates instead. But would still mean code generation == code in strings.

szarnekow:
As discusses with @mmews-n4 - merge into its base branch to have a single open review for this topic.

Issue #1734, opened 2020-04-22
Some tests in bundle org.eclipse.n4js.ide.tests are not executed on Jenkins


mor-n4:
Some of the tests in bundle `org.eclipse.n4js.ide.tests` are not executed on Jenkins.

For one of those tests there exists another ticket (see #1392) and that test is not working properly anyway. Fixing that test is not part of this ticket (just set it on ignore), it must be investigated why that test and others are not running on Jenkins and that must be fixed.

mor-n4:
NOTE: this is not done yet. One test is on ignore (search for `GH-1734`).

mor-n4:
Probably related:
```
Caused by: java.lang.ArrayIndexOutOfBoundsException: arraycopy: last source index 8 out of bounds for object array[7]
	at java.base/java.lang.System.arraycopy(Native Method)
	at org.eclipse.emf.common.util.BasicEList.grow(BasicEList.java:729)
	at org.eclipse.emf.common.util.BasicEList.addUnique(BasicEList.java:415)
	at org.eclipse.emf.common.util.AbstractEList.add(AbstractEList.java:304)
	at org.eclipse.emf.common.notify.impl.BasicNotifierImpl$EAdapterList.add(BasicNotifierImpl.java:224)
	at org.eclipse.n4js.resource.PostProcessingEntryTracker.lambda$0(PostProcessingEntryTracker.java:55)
```

szarnekow:
When running org.eclipse.n4js.ide.tests.RuntimeDependencyValidationIdeTest locally, I see a consistent single failure on `testLoadtimeDependencyConflict()`
```
java.lang.AssertionError: issues in module test-workspace/test-project/src/C.n4js do not meet expectation
EXPECTED:
(Error, [0:16 - 0:19], A load-time dependency target module B must only be imported once within the same runtime dependency cycle, but B is also imported by module X.
    Containing runtime dependency cycle cluster:
        *A.n4js --> Y.n4js
        *B.n4js --> A.n4js
        C.n4js --> B.n4js
        X.n4js --> B.n4js, C.n4js
        Y.n4js --> X.n4js)
ACTUAL:
    <none>
	at org.junit.Assert.fail(Assert.java:88)
	at org.eclipse.n4js.ide.tests.server.AbstractIdeTest.assertIssuesInModules(AbstractIdeTest.java:729)
[...]

PR #1744, opened 2020-04-29
GH-1734: Some tests in bundle org.eclipse.n4js.ide.tests are not executed on Jenkins

mor-n4:
See #1734 

Various fixes by @szarnekow required before certain tests can be activated in the maven build.
Also activates those tests (see change in pom.xml).


plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/XBuildManager.java
```diff
@@ -271,7 +274,7 @@ public XBuildable getIncrementalGenerateBuildable(List<URI> dirtyFiles, List<URI
 			throw ce;
 
 		} catch (Exception e) {
-			// recover
+			// recover and also discard the build queue - state is undefined afterwards.
 			this.dirtyFiles.clear();
 			this.deletedFiles.clear();
```
mor-n4, line 279: How do we get back into a well-defined state after this?

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/XLanguageServerImpl.java
```diff
@@ -445,8 +445,7 @@ public void exit() {
 				LOG.info("Shutdown done");
 				return result;
 			};
-		});
-		return CompletableFuture.completedFuture(null);
+		}).thenApply((any) -> new Object());
```
mor-n4, line 448: I guess the `#thenApply()` is only needed to satisfy the return type of `#shutdown()`?

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/concurrent/XRequestManager.java
```diff
@@ -74,6 +74,25 @@ synchronized public void shutdown() {
...
 	}
 
+	/**
+	 * Allows to join all requests that have been submitted so far.
+	 *
+	 * @return a future that can be awaited
+	 */
+	synchronized public CompletableFuture<Void> allRequests() {
+		CompletableFuture<?>[] cfs;
+		synchronized (this) {
```
mor-n4, line 84: Considering the method is declared synchronized, does this additional synchronized block have any effect?

plugins/org.eclipse.n4js/src/org/eclipse/n4js/validation/validators/packagejson/AbstractPackageJSONValidatorExtension.java
```diff
@@ -284,7 +286,11 @@ protected boolean isResponsible(Map<Object, Object> context, EObject eObject) {
 		if (!fileName.equals(IN4JSProject.PACKAGE_JSON)) {
 			return false;
 		}
-		IN4JSProject project = n4jsCore.findProject(pckjsonUri).get();
+		Optional<? extends IN4JSProject> optProject = n4jsCore.findProject(pckjsonUri);
+		if (!optProject.isPresent()) {
+			return false;
```
mor-n4, line 291: The exception thrown at this point before your fix hinted us towards all the other issues. So logging an error would be good, here.

tests/org.eclipse.n4js.ide.tests/src/org/eclipse/n4js/ide/tests/server/AbstractIdeTest.java
```diff
@@ -149,6 +149,10 @@ static final public void resetPrintStreams() {
 	/** Deletes the test project in case it exists. */
 	@After
 	final public void deleteTestProject() {
+		// clear thread pools
+		languageServer.shutdown().join();
+		languageServer.getRequestManager().shutdown();
```
mor-n4, line 154: Would a `joinServerRequests()` be an option here? (instead of lines 153, 154).
(except for the disadvantage that we would then always wait for those obsolete requests to complete instead of cancelling them)

szarnekow: joinServerRequests would not shutdown the server. The shutdown should be faster than a joinServerRequests.

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/server/N4JSLanguageServer.java
```diff
@@ -22,10 +22,12 @@
 
 import com.google.common.base.Optional;
 import com.google.common.collect.Lists;
+import com.google.inject.Singleton;
 
 /**
  * Language server for N4JS.
  */
+@Singleton
```
mmews-n4, line 30: [It's a trap!](https://www.youtube.com/watch?v=4F4qzPbcFiA)

plugins/org.eclipse.n4js/src/org/eclipse/n4js/validation/validators/packagejson/AbstractPackageJSONValidatorExtension.java
mor-n4: True, but you could already add something like `LOGGER.error(...)` to make sure we won't forget this at this location.

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/XBuildManager.java
mmews-n4: Some months ago I modified code here but we need to create some tests for this scenario (although this would be difficult).

plugins/org.eclipse.n4js/src/org/eclipse/n4js/validation/validators/packagejson/AbstractPackageJSONValidatorExtension.java
mmews-n4: or we add a general assertion for the redirected output string

szarnekow: I don't understand this idea.
What I often do to detect race conditions: add a lot of logging and run code in a loop. If one scenario fails, compare the logs of good and bad cases. Would be great, if that would work for the AbstractIdeTest.

mmews-n4: We talk about `AbstractIdeTest#redirectPrintStreams()` right? For testing/debugging it might be an idea to comment out the line inside that method body. As a result, the redirection would be disabled. Regarding "general assertion" I meant that we can still get the output via `SystemOutRedirecter#getSystemOut()` as a default and check that for certain things like `!getSystemOut().contains("Exception")` (If that helps?).

mor-n4:
Thanks @szarnekow!
Builds are green, except the pending test (did set that one on ignore). Merging ...

Issue #1728, opened 2020-04-20
As a developer I want the incremental LSP builder to re-generate all changed files when entering clean state


mor-n4:
Terminology: when one or more files are open *and* have unsaved changes, we say the LSP workspace is in "dirty state"; otherwise the workspace is in "clean state".

### Part 1

When leaving dirty state and entering clean state (i.e. saving the last of one or more changed files), the incremental builder must re-generate output files of all N4JS files that have been changed since leaving clean state (i.e. since changing the first file) and re-generate project-state files accordingly as needed. Currently, the builder re-generates only the last saved N4JS file and other N4JS files that are affected by that last file saved.

To reproduce:
1. create a new, empty N4JS project and open it in the LSP client.
2. create two N4JS files: `A.n4js` with `export public class A {}` and file `B.n4js` with `export public class B {}`.
3. execute "N4JS: Rebuild"
4. look at the two output files and assert that they have been generated correctly.
5. open both N4JS files and change the class name in each file *without* saving.
6. now save `A.n4js`.
7. now save `B.n4js` (don't use "save all files" functionality).
8. look at the two output files: only `B.js` has been re-generated; the output file `A.js` still contains the original class name!

Also make sure the project-state file is re-generated correctly (i.e. contains the new class name for `A.n4js`).

### Part 2

Changes in non-opened files must also be considered in this context. When a non-opened file F is changed (i.e. client sends LSP event "didChangeWatchedFiles"), then one of two cases applies:
1. iff the workspace is in dirty state, then F has to be added to the set of files that must be regenerated when entering clean state;
2. iff the workspace is in clean state, then F (and its dependants) must be regenerated immediately.

### ACs

1. add/improve clean/dirty state handling to cover part 1 and 2 above,
2. review clean/dirty state handling in general,
3. add tests (there are already some tests covering this in class `IncrementalBuilderGenerateTest`).

mor-n4:
There already exist some tests for this issue, search code for `FIXME GH-1728`.

mor-n4:
In this context, also consider `didClose()` events. There are two cases: closing a file that does not have unsaved changes and closing a file that has unsaved changes (and thus discarding the unsaved changes).

The latter case may lead to a transition of the workspace from dirty to clean state.

mor-n4:
Idea by @szarnekow:

Let `XBuildResult` tell whether it represents a complete or partial result:
```
XBuildResult {
  // zusätzlich zu dem, was schon drin ist
[...]

PR #1745, opened 2020-04-29
GH-1728: As a developer I want the incremental LSP builder to re-generate all changed files when entering clean state

mor-n4:
See #1728.


plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/XWorkspaceManager.java
```diff
@@ -97,6 +99,38 @@
...
+	/** See {@link #addFilesAwaitingGeneration(Collection, Collection)}. */
+	private final FileChangeTracker filesAwaitingGeneration = new FileChangeTracker();
+
+	/** Thread-safe tracking of file changes and/or deletions over time. */
+	private static class FileChangeTracker {
+
+		private final Set<URI> changedFiles = new LinkedHashSet<>();
+		private final Set<URI> deletedFiles = new LinkedHashSet<>();
+
+		public synchronized void clear() {
```
szarnekow, line 111: The synchronized keywords suggest that this code is called from different threads. Is that the case?

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/XProjectManager.java
```diff
@@ -197,6 +197,10 @@ protected XBuildResult doBuild(Set<URI> dirtyFiles, Set<URI> deletedFiles,
 
 		XBuildResult result = incrementalBuilder.build(request);
 
+		if (!request.isGeneratorEnabled()) {
```
szarnekow, line 200: I don't really like the flow from WorkspaceManager->BuildManager->ProjectManager->WorkspaceManager

Can we maintain the list of waiting files at the same location where we maintain the pending deltas?

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/XWorkspaceManager.java
```diff
@@ -137,6 +171,43 @@ public boolean hasContent(URI uri) {
 		}
 	};
 
+	/**
+	 * Tells whether the workspace is in dirty state. The workspace is said to be in dirty state iff at least one file
+	 * is open AND is dirty, i.e. has unsaved changes.
+	 */
+	public boolean isDirty() {
```
szarnekow, line 178: Can we call this `hasDirtyDocuments()`

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/XWorkspaceManager.java
```diff
@@ -256,23 +328,27 @@ protected XBuildable getIncrementalDirtyBuildable(List<URI> dirtyFiles, List<URI
...
-	/** Triggers an incremental build, and will generate output files. */
+	/**
+	 * Triggers an incremental build, and will generate output files. In addition to the given files, also all
+	 * {@link #filesAwaitingGeneration files awaiting generation} will be built.
+	 */
 	protected XBuildable getIncrementalGenerateBuildable(List<URI> dirtyFiles, List<URI> deletedFiles) {
-		XBuildManager.XBuildable buildable = buildManager.getIncrementalGenerateBuildable(dirtyFiles, deletedFiles);
+		addFilesAwaitingGeneration(dirtyFiles, deletedFiles);
+
+		Pair<List<URI>, List<URI>> changedAndDeleted = filesAwaitingGeneration.getAndClear();
```
szarnekow, line 346: Pair with getKey and getValue does not really express the intent of this data structure.
Maybe offer `FilesAwaitingGeneration.snapshot(): FilesAwaitingGeneration` that returns a snapshot of the original mutable structure?

tests/org.eclipse.n4js.ide.tests/src/org/eclipse/n4js/ide/tests/builder/IncrementalBuilderGenerateTest.xtend
```diff
@@ -124,6 +123,67 @@ cleanBuildAndWait(); // FIXME GH-1728 remove!
 		projectStateSnapshot.assertChanged();
 	}
 
+	/**
+	 * Given two modules Main, Other with a dependency from Main to Other, this test asserts that a change in Other
+	 * that has an effect on Main's output code will appear in Main's output file the moment Other is saved (assuming
+	 * Other is the only file being open).
+	 */
+	@Test
+	def void testChangeInOpenedFile_propagatesToOutputFileOfDependantModuleWhenSaved() {
```
szarnekow, line 132: `Dependant`->`Dependent` ?

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/XProjectManager.java
```diff
@@ -197,6 +197,10 @@ protected XBuildResult doBuild(Set<URI> dirtyFiles, Set<URI> deletedFiles,
...
+			workspaceManager.addFilesAwaitingGeneration(result.getAffectedResources());
```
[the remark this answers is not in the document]
mor-n4: I guess you mean `XWorkspaceManager#getIncrementalDirtyBuildable()` instead of `XWorkspaceManager#getIncrementalGenerateBuildable()`, right?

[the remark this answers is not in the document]
mor-n4: True, that looks like a *much* better location. The entire handling of "stuff awaiting generation" will then be local to a single class (at least for now), and also one of @szarnekow 's comments will be resolved by that.

However, this means we have to be sure that in case of cancellation we will receive a partial result instead of exiting with a CancellationException. So far, neither the API docs (e.g. that of `XBuildable#build(CancelIndicator)`) nor the actual code (e.g. [1]) looks as if that is guaranteed. But I'll have a look.

[1] https://github.com/eclipse/n4js/blob/514cf619820ab9c67f1e57313e475d380f8af2a1/plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/XBuildManager.java#L273-L274

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/XWorkspaceManager.java
mor-n4: It was a deliberate decision of Marcus and me to introduce this term also on the workspace level (simplifies talking about various special cases of handling theses states).

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/XProjectManager.java
mor-n4: Good idea! Done.

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/XWorkspaceManager.java
mor-n4: It was, but it doesn't matter anymore because due to moving this to XBuildManager it is no longer necessary.

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/XWorkspaceManager.java
mor-n4: I was thinking about that, too, but did not like the fact that (1) the snapshot is mutable and, more importantly, that (2) I'd have to add two getters to class `FilesAwaitingGeneration` that are only useful in the snapshot-case, and even dangerous to use in the non-snapshot case.

Anyways, thanks to your idea of moving the entire handling to XBuildManager this code is no longer necessary.

tests/org.eclipse.n4js.ide.tests/src/org/eclipse/n4js/ide/tests/builder/IncrementalBuilderGenerateTest.xtend
```diff
@@ -395,7 +395,7 @@ class IncrementalBuilderGenerateTest extends AbstractIncrementalBuilderTest {
 
 		assertFalse(outputFile.exists()); // never generate output files in node_modules folders
 		projectStateSnapshot.assertChanged();
-		assertNoIssues(); // error in Main.n4js should show up TODO GH-1727
+		assertIssues("Main" -> #[ "(Error, [1:16 - 1:31], any is not a subtype of number.)" ]);
```
szarnekow, line 398: Beautiful :)

mor-n4:
Builds green, merging ...

Thanks for the good ideas & feedback, everyone!

Issue #1729, opened 2020-04-20
As a developer I want to have changes of workspace projects (new/delete) be respected by lsp server


mor-n4:
Think of user interactions that create new projects, such as:
- move (in/out) folder into workspace
- create/remove package.json
- add/remove n4js section to package.json
- effect of fixing syntax errors in package.json to workspace
- rename project
- change yarn workspaces property
- consider node_modules folder (regarding above mentioned changes)
- add/remove source folder in package.json

Extend tests (see incremental builder tests; consider rename of "incremental builder tests")

mor-n4:
There already exists a test for this issue, search code for `FIXME GH-1729`.

mor-n4:
Status: workspace changes are detected and internal data structures of the LSP builder are updated and newly added projects are built. However, the effect of newly added/removed projects on other, unchanged projects in the workspace is not handled correctly by the incremental builder. Created follow-up for this: #1766.

mor-n4:
STATUS: workspace changes are detected and internal data structures of the LSP builder are updated and newly added projects are built. However, the effect of newly added/removed projects on other, unchanged projects in the workspace is not handled correctly by the incremental builder. Created follow-up for this: #1766.

PR #1746, opened 2020-04-29
GH-1727: As a developer I want the incremental LSP builder to propagate changes across projects

### LINKED ISSUE #1727: As a developer I want the incremental LSP builder to propagate changes across projects
The incremental LSP builder does not yet handle the case that a change in one project requires another project to also be rebuilt.

To reproduce: given two N4JS projects `ProjectA` and `ProjectB` contained in the same yarn workspace project, with `ProjectB` having a dependency to `ProjectA`, create the following two files:
```javascript
// ProjectA/src/A.n4js
export public const x = 42;
```
and
```javascript
// ProjectB/src/B.n4js
import {x} from "A";
const y: number = x;
y;
```
Now open file `A.n4js` in VS Code and change `42` to `"hello"` (without saving). This should lead to an error in `B.n4js` but this error does not show up. Now save `A.n4js` and the expected error still does not show up. Now execute command "Rebuild" and the expected error shows up.

NOTE: this user story is about adding support for cross-project incremental building in general, not just fixing the above scenario.

mmews-n4:
closes #1727


mmews-n4:
Includes slight change of write project state to disk: is capsuled into XProjectManger now.

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/XProjectManager.java
```diff
@@ -198,11 +195,13 @@ protected XBuildResult doBuild(Set<URI> dirtyFiles, Set<URI> deletedFiles,
 		XBuildResult result = incrementalBuilder.build(request);
 
 		projectStateHolder.updateProjectState(request, result);
+		if (doGenerate && result.getAffectedResources().isEmpty()) {
```
szarnekow, line 198: I think the condition `persistedProjectStateOutdated` was inverse to this guard here

mmews-n4: Maybe a `!` before `….isEmpty()` is missing?

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/XBuildManager.java
```diff
@@ -247,27 +250,28 @@ private XBuildable doGetIncrementalBuildable(List<URI> dirtyFiles, List<URI> del
 		try {
 			Map<ProjectDescription, Set<URI>> project2dirty = computeProjectToUriMap(this.dirtyFiles);
 			Map<ProjectDescription, Set<URI>> project2deleted = computeProjectToUriMap(this.deletedFiles);
+			SetView<ProjectDescription> changedPDs = Sets.union(project2dirty.keySet(), project2deleted.keySet());
+			IOrderInfo<ProjectDescription> orderInfo = projectBuildOrderProvider.get(changedPDs);
```
szarnekow, line 254: How much time does it take to compute the build order for OPR?

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/XProjectManager.java
szarnekow: Yes, I think so

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/XBuildManager.java
mmews-n4: I measured the method ProjectOrderInfo#init() from top to bottom with a Stopwatch. The output of the LSP server (started with `--clean`) on OPR initial build was:

```
LSP server starts
Listening for LSP clients on port 5007...
Connected to LSP client
LSP Server connected
8.379 ms
6.776 ms
```

So the build order was computed twice. First for the initial build, and second for an incremental build when the server sent the openFile call, I guess ("I guess", because the latter number `6.776 ms` was printed out after the initial build was complete).

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/IOrderInfo.java
```diff
@@ -0,0 +1,40 @@
...
+import java.util.List;
+
+import org.eclipse.xtext.resource.IResourceDescription;
+
+/**
+ * This interface is intended for use cases where sorting elements and computing an "is affected" information are
+ * combined. The combination might be reasonable since both of these computations might rely on similar data which would
+ * be cached for performance reasons.
+ *
+ * Implementations of this interface are sorted {@link Iterable}s. The idea is to iterate over a sorted set of elements
```
szarnekow, line 23: Formatting in JavaDoc is slightly improved by `{@link Iterable iterables}`

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/XBuildManager.java
```diff
@@ -125,13 +124,14 @@ public String toString() {
...
 	@Inject
-	private XTopologicalSorter topoSorter;
+	private ProjectOrderInfo.Provider projectBuildOrderProvider;
 
 	private final LinkedHashSet<URI> dirtyFiles = new LinkedHashSet<>();
 
 	private final LinkedHashSet<URI> deletedFiles = new LinkedHashSet<>();
 
-	private List<IResourceDescription.Delta> unreportedDeltas = new ArrayList<>();
+	/** Holds all deltas that */
```
szarnekow, line 133: There is something missing in this comment

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/XBuildManager.java
```diff
@@ -141,10 +141,12 @@ public String toString() {
 	public List<IResourceDescription.Delta> doInitialBuild(List<ProjectDescription> projects,
 			CancelIndicator indicator) {
 
-		List<ProjectDescription> sortedDescriptions = sortByDependencies(projects);
+		IOrderInfo<ProjectDescription> orderInfo = projectBuildOrderProvider.get(projects);
```
szarnekow, line 144: Renaming the orderInfo to orderedProjects would slightly improve the readability below

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/XBuildManager.java
```diff
@@ -302,52 +306,30 @@ protected void queue(Set<URI> files, Collection<URI> toRemove, Collection<URI> t
...
-	/** Get a sorted list of projects to be build. */
-	protected List<ProjectDescription> sortByDependencies(Iterable<ProjectDescription> projectDescriptions) {
-		List<ProjectDescription> sortedProjectDescriptions = topoSorter.sortByDependencies(projectDescriptions,
-				this::reportDependencyCycle);
-
-		String output = "Project build order:\n  ";
-		output += String.join("\n  ", IterableExtensions.map(sortedProjectDescriptions, pd -> pd.getName()));
+	/** Prints build order */
+	protected void printBuildOrder(IOrderInfo<ProjectDescription> orderInfo) {
+		String output = "Project build order:\n  "
```
szarnekow, line 331: I know that this is "just" a string concatenation, but please surround this by `Log.isInfoEnabled()`

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/XProjectManager.java
```diff
@@ -198,11 +195,13 @@ protected XBuildResult doBuild(Set<URI> dirtyFiles, Set<URI> deletedFiles,
 		XBuildResult result = incrementalBuilder.build(request);
 
 		projectStateHolder.updateProjectState(request, result);
+		if (doGenerate && !result.getAffectedResources().isEmpty()) {
```
szarnekow, line 198: Two questions about this:
- the build request does know if doGenerate is true or false. The build result knows the affected resources. Should the projectStateHolder decide on its own, if it should persist? E.g. should this logic be moved into the updateProjectState call above?
- projectStateHolder.writeProjectState has an internal guard that it should not write if the hashed contents is empty. Why is that?

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/build/XIndexer.java
```diff
@@ -224,13 +225,18 @@ public URI getURI() {
...
 			Collection<Delta> newDeltas, Collection<Delta> allDeltas, XBuildContext context) {
+
 		ResourceDescriptionsData originalIndex = context.getOldState().getResourceDescriptions();
-		List<URI> affectedURIs = IterableExtensions.toList(IterableExtensions.filter(remainingURIs, it -> {
-			IResourceDescription.Manager manager = context.getResourceServiceProvider(it)
-					.getResourceDescriptionManager();
-			IResourceDescription resourceDescription = originalIndex.getResourceDescription(it);
-			return isAffected(resourceDescription, manager, newDeltas, allDeltas, index);
-		}));
+		List<URI> affectedURIs = new ArrayList<>();
```
szarnekow, line 230: 👍

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/build/XIndexer.java
```diff
@@ -267,8 +274,8 @@ public URI getURI() {
 		try {
 			this.compilerPhases.setIndexing(context.getResourceSet(), true);
 			List<IResourceDescription.Delta> result = new ArrayList<>();
-			for (IResourceDescription.Delta delta : context.executeClustered(changedURIs,
-					it -> addToIndex(it, true, oldIndex, context))) {
+			List<Delta> deltas = context.executeClustered(changedURIs, it -> addToIndex(it, true, oldIndex, context));
```
szarnekow, line 277: Appreciated attention to detail!

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/build/XStatefulIncrementalBuilder.java
```diff
@@ -109,10 +109,13 @@ public XBuildResult launch() {
 			Set<URI> remainingURIs = new LinkedHashSet<>(oldIndex.getAllURIs()); // note: creating a copy!
 
 			XIndexer.XIndexResult result = indexer.computeAndIndexDeletedAndChanged(request, context);
-			operationCanceledManager.checkCanceled(request.getCancelIndicator());
-
 			ResourceDescriptionsData newIndex = result.getNewIndex();
 			List<Delta> deltasToBeProcessed = result.getResourceDeltas();
+			List<Delta> deltasFromExternal = indexer.computeAndIndexAffected(newIndex, remainingURIs,
```
szarnekow, line 114: The indexer does not have a chance to cancel this operation?

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/XBuildManager.java
```diff
@@ -247,27 +250,28 @@ private XBuildable doGetIncrementalBuildable(List<URI> dirtyFiles, List<URI> del
...
 			Map<ProjectDescription, Set<URI>> project2dirty = computeProjectToUriMap(this.dirtyFiles);
 			Map<ProjectDescription, Set<URI>> project2deleted = computeProjectToUriMap(this.deletedFiles);
+			SetView<ProjectDescription> changedPDs = Sets.union(project2dirty.keySet(), project2deleted.keySet());
+			IOrderInfo<ProjectDescription> orderInfo = projectBuildOrderProvider.get(changedPDs);
 
-			SetView<ProjectDescription> allDescriptions = Sets.union(project2dirty.keySet(), project2deleted.keySet());
-			List<ProjectDescription> sortedDescriptions = sortByDependencies(allDescriptions);
-
-			for (ProjectDescription descr : sortedDescriptions) {
+			for (ProjectDescription descr : orderInfo) {
```
mor-n4, line 256: There is quite some logic behind the iterator that is implicitly created here, esp. considering the fact that this iterator will be influenced by the `#visitAffected()` call inside the for loop's block and that several iterators may be created from a single iterable. Please try to simplify this and make it a bit more explicit.

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/XBuildManager.java
mmews-n4: tension is rising ... 😄

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/XProjectManager.java
mmews-n4: Regarding #1: I changed that, good idea. However, it was necessary to change the XBuildRequest also.
Regarding #2: I do not remember the reason. I removed it and test if it still works.

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/build/XStatefulIncrementalBuilder.java
mmews-n4: Do you mean that the indexer should call `operationCanceledManager.checkCanceled`?

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/ProjectBuildOrderInfo.java
```diff
@@ -0,0 +1,195 @@
...
+	}
+
+	/** Workspace manager */
+	@Inject
+	protected XWorkspaceManager workspaceManager;
+
+	/** Inversed set of project dependency information */
+	final protected Multimap<String, ProjectDescription> inversedDependencies = HashMultimap.create();
+	/** Build order of projects */
+	final protected List<ProjectDescription> sortedProjects = new ArrayList<>();
```
mor-n4, line 123: A bit unfortunate that this list is mutable, but I see why that is the case (initialized not in ctor but in `@Inject init()`).

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/ProjectStatePersister.java
```diff
@@ -301,14 +308,11 @@ private void writeFingerprints(Collection<? extends HashedFileContent> files, Da
...
-	private void writeValidationIssues(Map<URI, ? extends Collection<Issue>> validationIssues, DataOutput output)
-			throws IOException {
-
+	private void writeValidationIssues(Multimap<URI, Issue> validationIssues, DataOutput output) throws IOException {
 		int numberSources = validationIssues.size();
 		output.writeInt(numberSources);
-		for (Map.Entry<URI, ? extends Collection<Issue>> srcIssues : validationIssues.entrySet()) {
-			URI source = srcIssues.getKey();
-			Collection<Issue> issues = srcIssues.getValue();
+		for (URI source : validationIssues.keys()) {
```
mor-n4, line 314: Due to the change from `Map<URI,Collection<Issue>>` to `MultiMap<URI,Issue>` this for-loop's block won't be executed for resources without issues.

However, because the original code converted the map to an `ListMultiMap` and with `MultiMap#asMap()` back to a Map in method `#writeProjectState()`, it has the same effect if this is the only call-path that leads here. But double check if not done already.

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/ProjectBuildOrderInfo.java
mmews-n4: That's true. We can change this in GH-1729 maybe.

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/ProjectStatePersister.java
mmews-n4: Thanks, I agree. Also, I couldn't find indications for writing out empty lists to the project state.

mmews-n4:
Thanks for your help!

PR #1748, opened 2020-04-29
GH-1747: Write project state asynchronously from the build

### LINKED ISSUE #1747: As a user, I want fast incremental builds
Right now, the incremental build for dirty files is a lot faster than the incremental build for saved files - especially in larger modules.
The reason is the cost of writing the project state.
Since the project state is only a caching, it is save to move it out of the normal build processing and write the files to disk asynchronously.

szarnekow:
closes #1747


plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/ProjectStatePersister.java
```diff
@@ -329,12 +376,14 @@ public PersistedState readProjectState(InputStream stream, String expectedLangua
...
 		int version = stream.read();
-		if (version != CURRENT_VERSION) {
+		if (version != CURRENT_VERSION && version != VERSION_1) {
 			return null;
 		}
 
-		try (ObjectInputStream input = new ObjectInputStream(
-				new BufferedInputStream(new GZIPInputStream(stream, 8192)))) {
+		try (BufferedInputStream buffy = new BufferedInputStream(new GZIPInputStream(stream, 8192));
+				InputStream dataOrObjectInputStream = version == VERSION_1 ? new ObjectInputStream(buffy)
```
mmews-n4, line 384: I see. How about not having Version#1? At the moment, we don't have regular users yet.

szarnekow: The logged exceptions were annoying when switching branches ;)

Apart from that, it would be ok to remove the backwards compat there. Regular users would see a rebuild in any case since the language version would be differnt.

Issue #1749, opened 2020-05-04
LSP builder not using Xtext API for handling cancellation exceptions


mor-n4:
Some methods in the LSP builder catch `java.util.concurrent.CancellationException` but do not catch `org.eclipse.core.runtime.OperationCanceledException` which is thrown by our code, resulting in behavior for "unknown exception" to be executed in case of cancellation.

Probably the Xtext API for handling cancellation should be used instead, e.g. something like `operationCanceledManager.isOperationCanceledException(t)`.

This applies to methods:
```
org.eclipse.n4js.ide.xtext.server.XBuildManager.doIncrementalBuild(boolean, CancelIndicator)
org.eclipse.n4js.ide.xtext.server.build.XStatefulIncrementalBuilder.launch()
```
(check code for more cases).

Seems to be done properly in:
```
org.eclipse.n4js.ide.xtext.server.build.XIncrementalBuilder.build(XBuildRequest, IResourceClusteringPolicy)
```

mor-n4:
For reference:

As it turned out while working on this, using _only_ the `OperationCanceledManager` is not possible in all places in bundle `org.eclipse.n4js.ide`, because `OperationCanceledManager` does not handle Java's `java.util.concurrent.CancellationException`. So, in places where that exception might occur, it must be handled explicitly, maybe in addition to using `OperationCanceledManager` for the other kinds of cancellations.

Alternatively, support for `java.util.concurrent.CancellationException` could be added to `OperationCanceledManager`, but this would require a change in Xtext (not done as part of this task).

szarnekow:
cf https://github.com/eclipse/n4js/blob/master/plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/concurrent/XRequestManager.java#L125

PR #1752, opened 2020-05-05
GH-1749: LSP builder not using Xtext API for handling cancellation exceptions

mor-n4:
See #1749.

I wonder if a `java.util.concurrent.CancellationException` can even occur at the two places I changed, but haven't removed the support for it, for now.

### DIFF
--- a/plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/build/XStatefulIncrementalBuilder.java
+++ b/plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/build/XStatefulIncrementalBuilder.java
@@ -157,7 +157,9 @@ public class XStatefulIncrementalBuilder {
 		} catch (CancellationException e) {
 			// catch CancellationException here and proceed normally to save already resolved deltas
 		} catch (Throwable th) {
-			if (!operationCanceledManager.isOperationCanceledException(th)) {
+			if (operationCanceledManager.isOperationCanceledException(th)) {
+				// catch OperationCanceledException, etc. here and proceed normally to save already resolved deltas
+			} else {
 				throw th;
 			}
 		}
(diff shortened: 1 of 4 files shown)


plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/build/XStatefulIncrementalBuilder.java
```diff
@@ -156,6 +156,10 @@ public XBuildResult launch() {
 
 		} catch (CancellationException e) {
 			// catch CancellationException here and proceed normally to save already resolved deltas
+		} catch (Throwable th) {
+			if (!operationCanceledManager.isOperationCanceledException(th)) {
+				throw th;
```
mmews-n4, line 161: How about
```
if (operationCanceledManager.isOperationCanceledException(th)) {
	// catch OperationCanceledException here and proceed normally to save already resolved deltas
} else {
	throw th;
}
```

Also: Why catching a `Throwable`? Above it is also an `Exception`

mor-n4:
@mmews-n4 As discussed, I added the read-request change to this PR. See commit 8ad98d6684b3db6795183c9e8ab680ceed23d524.

tests/org.eclipse.n4js.ide.tests/src/org/eclipse/n4js/ide/tests/xtext/XRequestManagerTest.xtend
```diff
@@ -152,6 +155,33 @@ class XRequestManagerTest {
 		assertEquals(2, sharedState.get)
 	}
 
+	// TODO when re-enabling concurrent execution of read-requests:
```
mmews-n4, line 158: Maybe use a ticket id so we could search for the two locations easily.

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/build/XStatefulIncrementalBuilder.java
mor-n4: Regarding "how about": actually I thought about that too, but then didn't like the duplication of the comment. But if you also stumbled across the readability here, let's use the more readable variant you proposed.

Regarding Throwable: method `#isOperationCanceledException()` accepts a throwable, i.e. may return true for throwables that are not exceptions; this is the case, for example, if a `OperationCanceledError` is thrown; so to cover all possible cases a throwable should be used.

However, I now adjusted the corresponding location in XBuildManager (changed from catching Exception to catching Throwable) for consistency.

mor-n4:
Thanks @mmews-n4!

Builds green, merging ...

Issue #1757, opened 2020-05-07
As a smith I want infrastructure for custom LSP notifications/requests from server to client


mor-n4:
We would like to have infrastructure to send custom JSON-RPC notifications/requests from the N4JS LSP server to the LSP client, i.e. the N4JS extension for VS Code.

NOTE: since this feature needs special handling on client side and we do not want to implement special handling for all possible LSP clients, this will be supported only for VS Code (and maybe later for the Eclipse-based LSP client) and should be used only for temporary features or internal "smith" features. Therefore, the feature should be labeled as "smith" in the code.

Issue #1751, opened 2020-05-05
As a developer I want a super simple, preliminary progress indicator for the incremental build


mor-n4:
The incremental builder of the LSP server should report its progress to the client to let the user know if a build is currently running.

The LSP protocol and VSCode will soon have special support for such progress reporting. Therefore, this user story is only about a very simple, temporary solution until a proper solution can be implemented based on the upcoming LSP / VSCode features.

### ACs

1. Show when builds are started / aborted (i.e. cancellation) / completed.
   Percentage is not required.
2. Should be based on the ordinary logging we already have.
3. Optional: maybe log to a new, dedicated channel "N4JS Build Progress" instead of the existing channel "N4JS Language Server".

PR #1758, opened 2020-05-07
GH-1751/GH-1757: As a developer I want a super simple, preliminary progress indicator for the incremental build

mor-n4:
See #1751, #1757.

Slightly less super simple build progress indicator, using a separate output channel in VS Code.
Also includes first version of support for custom server-to-client communication.


mor-n4:
Builds green, merging!

PR #1760, opened 2020-05-11
GH-1734: Some tests in bundle org.eclipse.n4js.ide.tests are not executed on Jenkins

mor-n4:
See #1734.


mor-n4:
Builds green, merging ...

Issue #1684, opened 2020-03-13
Optional chaining on method calls looses call context


dbo:
```n4js
let foo: Object+;

foo.bar?.(5);
```

transpiles into
```js
($opt = foo.bar) == null ? void 0 : $opt(5);
```

thus `bar` is not called with correct call context `foo`.

Issue #1680, opened 2020-02-28
Chaining does not work for function arguments


dbo:
```n4js
export public interface I {
    qux: string
}
function foo(loo: I, bar = loo?.qux) {
}

foo(null);
```

Executing leads to:
```
ReferenceError: $opt is not defined
```

since transpiled to:
```js
function foo(loo, bar=($opt = loo) == null ? void 0 : $opt.qux) {
	let $opt;
}
```

mmews-n4:
`var` might fix that.

PR #1764, opened 2020-05-13
GH-1680/GH-1684: two bugs related to optional chaining

mor-n4:
See #1680, #1684.


mor-n4:
Builds green, merging ...

Issue #1765, opened 2020-05-14
As a developer I don't want duplicate/confusing error messages for unresolved imports


mor-n4:
In case an import cannot be resolved, several redundant error messages show up. All redundant messages should be suppressed (as far as feasible).

For example:
```javascript
import {OtherClass} from "Other";
new OtherClass().m();
```
Assuming module `Other` does not exist, the following error messages will be shown for the above code:
```
(Error, [0:8 - 0:18], Couldn't resolve reference to TExportableElement 'OtherClass'.)
(Error, [0:8 - 0:18], Import of OtherClass cannot be resolved.)
(Error, [0:25 - 0:32], Cannot resolve import target :: resolving simple module import : found no matching modules)
(Error, [1:4 - 1:14], Couldn't resolve reference to IdentifiableElement 'OtherClass'.)
```

This issue is mainly about suppressing redundant messages; improving the wording of the messages is not part of this user story, but some simple, easy-to-implement improvements should be made.

PR #1772, opened 2020-06-08
GH-1765: As a developer I don't want duplicate/confusing error messages for unresolved imports

mor-n4:
See #1765.

Also includes simplification of Xpect output tests, see IDE-1034.


mor-n4:
Builds green, merging.

Issue #1780, opened 2020-06-13
Transpiler bug when importing modules named after the same folder


dbo:
consider a module `/foo&bar.n4js` and a module `/foo.n4js` which export some variable:

foo.n4js
```js
export public const c = 5;

```
/foo/bar.n4js
```js
import {c} from "foo"
console.log(c)
```

the transpiled output is
```js
// Generated by N4JS transpiler; for copyright see original N4JS source file.

import 'n4js-runtime'
import {c} from './'

console.log(c);
//# sourceMappingURL=bar.map
```
which is wrong it should

```js
// Generated by N4JS transpiler; for copyright see original N4JS source file.

import 'n4js-runtime'
import {c} from '../foo'

console.log(c);
//# sourceMappingURL=bar.map
```

PR #1781, opened 2020-06-15
GH-1780: Transpiler bug when importing modules named after the same folder

mor-n4:
See #1780.


mor-n4:
Builds green, merging ...

Issue #1774, opened 2020-06-10
As a smith I want to separate editing functionality from the LSP builder (part 1)


mor-n4:
Each open file should have its own resource set and all editing functionality (content assist, hover, hyper links, ...) should be performed in the context of this resource set.

The goal of this separation is to solve the problem described in GH-1768.

This user story covers part 1. The goal of part 1 is to come up with a close-to-final implementation, including:
* most _existing_ tests in N4JS build should be green,
* early manual testing with large N4JS code bases,

but without:

* new tests for the new functionality,
* clean up,
* refactoring of existing LSP builder to remove all now obsolete functionality (e.g. `doGenerate==false`, workspace dirty state handling),
* merging to master.

PR #1782, opened 2020-06-18
GH-1774-sz: WIP for separating editing functionality from LSP builder

mor-n4:
Hi @szarnekow!
As discussed, here is the WIP for GH-1774. Thanks for taking a look at it!!

Summary:

* main goal: separation of "editing functionality" (e.g. content assist, hover, code actions, commands) from the builder.
* overall idea is that old classes `XWorkspaceManager`, `XProjectManager`, `XBuildManager`, etc. are now private to the builder and must not be directly accessed from outside the builder (e.g. from `XLanguageServerImpl`, `N4JSCodeActionService`, `N4JSCommandService`) and the new class `OpenFilesManager` was added for all functionality that needs EMF resources outside the builder.
* `XWorkspaceManager`'s role was slightly changed: it was meant as a global, top-level class accessible from everywhere (afaik), but it manages several things that are not required outside the builder; to limit changes in existing code as far as possible, I decided to keep `XWorkspaceManager` mostly unchanged and "move it into the builder", for now.
* the only information the builder provides to the outside is:
  (a) index = 1 resource description per file, (b) set of issues per file.
  This must be provided in a thread-safe manner.
  ~(NOTE: currently the builder also sends diagnostics to the client for non-opened files; this is a TODO and will be refactored).~
* main new classes are `OpenFilesManager`, `OpenFileContext`, `LSPExecutorService` (replaces XRequestManager).
  New helper classes include: `ConcurrentChunkedIndex`, `ConcurrentIssueRegistry`.
  To simplify "encapsulation" of builder-internal stuff, I had to introduce a new facade `LSPBuilder`.
* some changes I had to do in existing code:
  * refactor handling of issues to allow conversion between Issue and Diagnostic without requiring a Resource/XDocument (this was a pain!),
  * refactor handling of index and issues to allow thread-safe access to those from outside.

Current status:
* main functionality should be in place,
* all tests green,
* cancellation not yet activated in open file contexts (but active in builder),
* many small/medium details need improvement or reconsideration,
* NOT tested yet with real-world code.

A good intermediate commit that includes most new additions but few changes in existing code might be 55b01f3c90aa129f1cdeac5246261f479f6ebfb9.


plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/LSPBuilder.java
```diff
@@ -0,0 +1,223 @@
...
+
+	// FIXME GH-1774 thread safety!
+	public void addBuildListener(ILanguageServerAccess.IBuildListener listener) {
+		workspaceManager.addBuildListener(listener);
+	}
+
+	public CompletableFuture<Void> shutdown() {
+		return runBuildable("shutdown", () -> {
+			return (cancelIndicator) -> {
+				persister.pendingWrites().join();
```
szarnekow, line 193: Call `joinPersister` instead.

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/LSPIssue.java
```diff
@@ -29,13 +29,13 @@
  *
  * All line and column numbers are one-based. Offset is zero-based.
  */
-public class N4JSIssue extends IssueImpl {
+public class LSPIssue extends IssueImpl {
```
szarnekow, line 32: This rename here makes the review unfortunately unnecessary hard.

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/XLanguageServerImpl.java
```diff
@@ -471,132 +455,44 @@ public WorkspaceService getWorkspaceService() {
...
-
-	/**
-	 * Evaluate the params and deduce the respective build command.
-	 */
-	protected XBuildable toBuildable(DidChangeTextDocumentParams params) {
 		VersionedTextDocumentIdentifier textDocument = params.getTextDocument();
-		return workspaceManager.didChangeTextDocumentContent(getURI(textDocument),
-				textDocument.getVersion(), params.getContentChanges());
+		URI uri = getURI(textDocument);
+		if (openFilesManager.isOpen(uri)) { // FIXME GH-1774 reconsider
```
szarnekow, line 466: I think this is what the TODO is about, but I'll put my thoughts here nevertheless:
I'm not sure what the LSP client sends to the server if a file is changed on disk and not through the client. We must ensure that we don't miss these kinds of changes.

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/concurrent/ConcurrentChunkedIndex.java
```diff
@@ -0,0 +1,140 @@
...
+		return containerHandle2Data.get(containerHandle);
+	}
+
+	/** Sets the data for this container. */
+	public ResourceDescriptionsData setContainer(String containerHandle, ResourceDescriptionsData newData) {
+		Objects.requireNonNull(containerHandle);
+		Objects.requireNonNull(newData);
+		ResourceDescriptionsData oldData;
+		synchronized (this) {
+			oldData = containerHandle2Data.put(containerHandle, newData);
```
szarnekow, line 80: newData is mutable. Are we sure that no code path will mutate these in an unsafe fashion?

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/concurrent/ConcurrentIssueRegistry.java
```diff
@@ -0,0 +1,323 @@
...
+	}
+
+	protected void notifyListeners(IssueRegistryChangeEvent event) {
+		notifyListeners(Collections.singletonList(event));
+	}
+
+	protected void notifyListeners(List<IssueRegistryChangeEvent> events) {
+		List<IIssueRegistryListener> _listeners;
+		synchronized (this) {
+			_listeners = new ArrayList<>(listeners);
```
szarnekow, line 270: listeners could be a CopyOnWriteArrayList which would often avoid unnecessary pessimistic copies. Not sure how many listeners we expect, though.

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/concurrent/FutureUtil.java
```diff
@@ -0,0 +1,57 @@
...
+	 * Obtains the given future's result via {@link Future#get()}, but converts <code>java.util.concurrent</code>'s
+	 * {@link CancellationException} to Xtext's {@link OperationCanceledException} and wraps checked exceptions in a
+	 * {@link RuntimeException}.
+	 */
+	public static <T> T getCancellableResult(Future<T> future) {
+		try {
+			return future.get();
+		} catch (Exception e) {
+			Exception cancellation = getCancellation(e);
+			if (cancellation instanceof OperationCanceledException) {
```
szarnekow, line 34: Also consider OperationCanceledErrors.

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/concurrent/LSPExecutorService.java
```diff
@@ -0,0 +1,296 @@
...
+	/** The underlying executor service that is used to actually execute the tasks. */
+	@Inject
+	protected ExecutorService delegate;
+
+	/***/
+	@Inject
+	protected OperationCanceledManager operationCanceledManager;
+
+	/** Global queue of all pending tasks across all queue IDs. */
+	protected final List<QueuedTask<?>> pendingTasks = new ArrayList<>();
```
szarnekow, line 75: This should probably be an ArrayDeque

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/concurrent/LSPExecutorService.java
```diff
@@ -0,0 +1,296 @@
...
+	/** Queue IDs with currently running tasks. */
+	protected final Map<Object, QueuedTask<?>> runningTasks = new LinkedHashMap<>();
+
+	@SuppressWarnings("javadoc")
+	protected final class QueuedTask<T> implements Runnable, XCancellable {
+		protected final Object queueId;
+		protected final String description;
+		protected final Function<CancelIndicator, T> operation;
+		protected final QueuedTaskFuture<T> result;
+		protected boolean cancelled = false;
```
szarnekow, line 85: must be volatile

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/concurrent/LSPExecutorService.java
```diff
@@ -0,0 +1,296 @@
...
+				.forEach(t -> t.cancel());
+	}
+
+	/**
+	 * Blocks until all tasks complete that are running or pending at the time of invocation of this method OR are being
+	 * submitted by other threads while this method is waiting. Thus, this method waits for this executor to idle.
+	 */
+	public /* NOT synchronized */ void join() {
+		CompletableFuture<Void> allTasks;
+		while ((allTasks = allTasksOrNull()) != null) {
```
szarnekow, line 249: I don't think the JavaDoc is correct.
`allTasksOrNull may return null` and immediately thereafter another thread will submit a task. In that sense, the executor is not idle when `join` returns.

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/findReferences/XWorkspaceResourceAccess.java
```diff
@@ -25,27 +25,37 @@
...
+	public XWorkspaceResourceAccess(XLanguageServerImpl languageServer) {
+		this.languageServer = languageServer;
 	}
 
 	@Override
 	public <R> R readOnly(URI targetURI, IUnitOfWork<R, ResourceSet> work) {
-		XProjectManager projectManager = workspaceManager.getProjectManager(targetURI);
-		if (projectManager == null) {
+		OpenFilesManager openFilesManager = languageServer.getOpenFilesManager();
+		if (openFilesManager == null) {
```
szarnekow, line 41: I don't understand this condition.

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/findReferences/XWorkspaceResourceAccess.java
```diff
@@ -25,27 +25,37 @@
...
 			return null;
 		}
+		// FIXME GH-1774 consider re-using the resource set of the current open file context for other files:
+		// OpenFileContext currOFC = openFilesManager.currentContext();
+		// if (currOFC != null) {
+		// return doWork(currOFC.getResourceSet(), work);
+		// }
+		return openFilesManager.runInOpenOrTemporaryFileContextSync(targetURI, "XWorkspaceResourceAccess",
```
szarnekow, line 49: Two readonly may be nested in each other with different targetURIs. If the are from different projects, by transitivity we may have loaded two resources with the same URI into two different resource sets. I'm skeptical that we won't run into identity issues.

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/concurrent/LSPExecutorService.java
```diff
@@ -0,0 +1,296 @@
...
+ * <h2>Cancellation Handling</h2>
+ *
+ * Cancellation is handled entirely through Xtext's {@link CancelIndicator}s and a task's implementation can decide how
+ * to react to that (as usual with cancel indicators). Tasks will always be invoked, even if they were marked as
+ * cancelled while waiting on the queue, so implementors of a task may choose to implement a "non-cancellable" operation
+ * at the beginning of a task before reacting to the cancel indicator.
+ */
+// FIXME remove @Singleton?
+@Singleton // ensures queue IDs are "global" within a single injector
+public class LSPExecutorService {
```
szarnekow, line 62: This design looks like a simplified equivalent to Eclipses SchedulingRule.
What I don't understand is where a guarantee comes from that two tasks with different descriptions don't interfere with each other. As far as I can see, the description is usually the URI of the open resource of an editor (an editor id). This will sequentialise requests per editor. So far so good.

Now if two circular dependent editors run tasks that will be re-entrant on the executor service, we will dead-lock. How can we guarantee that this is not going to happen?

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/XLanguageServerImpl.java
mor-n4: In that case the client sends the separate LSP event `didChangeWatchedFiles`. The `OpenFilesManager` ignores these events entirely. Instead, they are sent to the `LSPBuilder`, triggering an incremental build which will eventually lead to an update of the persisted state index, which will then be forwarded to the `OpenFilesManager`.

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/concurrent/ConcurrentChunkedIndex.java
mor-n4: For now, I have solved this via the JavaDoc of the containing class `ConcurrentChunkedIndex`. There it is stated that clients are responsible for making sure not to change instances of `ResourceDescriptionsData` before/after passing them into / obtaining them from a `ConcurrentChunkedIndex` or creating a copy.

Last week I have already implemented an alternative introducing an `ImmutableResourceDescriptionsData` and enforcing its use in `ConcurrentChunkedIndex`, but I'm not sure yet if it might lead to far more copies than necessary (also, as a side note, I ran into a problem with fields of `ResourceDescriptionsData` being private and therefore not being able to copy `ImmutableResourceDescriptionsData` efficiently).

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/findReferences/XWorkspaceResourceAccess.java
mor-n4: The idea is that the `#runInOpen|TemporaryFileContext()` methods are used only in two cases:
1. by the LSP framework (mostly XLanguageServerImpl, maybe also N4JSCodeActionService / N4JSCommandService) to kick off the computation of some editing functionality,
2. somewhere nested inside editing functionality to run a computation in a resource entirely separate from the "main resource" of the open editor the editing functionality originated in.

No. 2 is expected to be a rather rare case and used only for operations that are self-contained and do not exchange `EObject`s (esp. `TModule` elements) with the outside. An example might be looking up the JSDoc string of a method in some arbitrary N4JS[D] file.

The standard case of using information of a resource B from within the context of an open file A is to simply load resource B in the resource set of the open file A (as usual with EMF and as the Eclipse editors are doing it usually).

So far my plan/intention.

HOWEVER: the current invocations of `#runInOpen|TemporaryFileContext()` on my branch are partly incorrect and thus give a wrong impression of the intention of those methods. The reason is that I haven't fully implemented loading of resources from source in the resource sets of open files (no `IExternalContentSupport` registered on the resource sets in `OpenFileContext` yet) and I have just added the `#runInOpen|TemporaryFileContext()` as "place holders", because I have to revisit those anyway.

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/concurrent/LSPExecutorService.java
mor-n4: `LSPExectorService` (with its queueId approach) and the `OpenFilesManager#runInOpen|TemporaryFileContext()` methods are not intended for fine-grained operations/computations that must be composable (see also my answer to your previous question). They correspond more to `XtextDocument#readOnly|modify()` when invoked on an XtextEditor's document, than to Eclipse's scheduling rules.

Another way to put it: the scenario you described ("Now if two circular ...") could also be replicated in the context of Eclipse if one Eclipse Xtext editior A would run an operation in another Xtext editor B via `getDocument().readOnly|modify()` which would then try to obtain a lock back on A. But this is unlikely because `XtextDocument#readOnly|modify()` is not intended for such composable operations. In 99% of cases editor should stick to their own context or create temporary, entirely unrelated and self-contained contexts.

However, it would probably be possible to detect such scenarios in `OpenFilesManager#runInOpen|TemporaryFileContext()` (or even in `LSPExecutorService`?) to at least fail more or less gracefully instead of running into a dead lock.

If we find that we need `OpenFilesManager#runInOpen|TemporaryFileContext()` more often and for more cases than I had in mind, we might have to introduce a separation of read-only and read/write operations in OpenFilesManager or some other locking mechanism, but I wanted to avoid this as long as we don't run into a true need for it.

szarnekow: > the scenario you described ("Now if two circular ...") could also be replicated in the context of Eclipse

In theory yes, practically all readOnlys of that shape in Xtext Eclipse would be implicitly sequential since they are triggered from the UI thread.

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/findReferences/XWorkspaceResourceAccess.java
szarnekow: > no IExternalContentSupport registered

Understood.

szarnekow:
I was thinking about this a little more and I'm wondering if the LspExecutorService suggests something different from what it actually does - it sequentialises the access to a document.
If we want to make access to a document sequential, I'd rather try to put an API for that on the document itself. It should probably be even more symmetric to the Eclipse impl.

Let's say we introduce a type
```
interface XtextLspDocument {
  Document document();
  <R> read(DocumentFunction<R> fun);
 interface DocumentFunction<R> {
  R compute(Resource r, Document d, CancelIndicator ci) throws CancellationException, 
ExecutionException;
  }
  void update("TextEditThingy" /* please put the correct type name from LSP here */ d);
  // probably more snapshot based accessors, e.g.
  ImmutableList<Issue> currentIssues(); 
}
```

Languages that have a semantic of "parallel reads are allowed" could implement the document based on a "real" ExecutorService. For N4JS this could be a singleThreadExecutor. It wouldn't spread the synchronisation mechanism across different types and wouldn't impose limitations based on some shared service.
What do you think?

mor-n4:
The LSPExecutorService is meant as a low-level utility that allows to easily sequentialize certain tasks by giving them equal "queueIds", without the need for a separate thread per queueId (as with single-thread executors). The API for sequential access to resources/documents is OpenFilesManager.

The idea with XtextLspDocument sounds interesting, esp. because it helps align the different cases. But I have to think about it a bit more.

> ... and wouldn't impose limitations based on some shared service.

By "shared service", do you mean the `OpenFilesManager`?

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/concurrent/LSPExecutorService.java
mor-n4: So you mean in Eclipse editors do not actually run in parallel?

If I make an edit in one of 10 open editors (not saving) and all other editors are affected by that edit, then scoping, resolution, validation, etc. of the other 9 editors is done one after another?

If that is the case and it never caused lag in Eclipse, we could think about that for LSP, too. So far, my goal on this branch was to allow to distribute work for open editors across multiple threads.

szarnekow: Hmm that's not what I meant.
In Eclipse, we have the UI thread. All user interaction is triggered from there. When a document is changed, a document event is scheduled and after a delay, the resource will be refreshed while holding a modify lock on the document. Linking etc will start while holding that lock and afterwards the other editors will receive a notification. They will be processed in own jobs subsequently, each holding a lock per document.

What I was referring to: This logic is under almost full control of the framework. That's why we "know" that the code won't escape to another document and attempt to do a modify().

Now back to the API on IXtextDocument - readOnly and modify. If a client implements code that aquires more than one modify on different docs, it could potentially deadlock, too. However, the modify(..) is usually only triggered from code that runs on the UI thread. Therefore it's would require other background threads that try to do concurrent modifies. In that case, we'd deadlock.

But: since that's highly unlikey, we won't see that. With LSP and Commands, that are executed on their own thread, the sequential UI thread is not present.

mor-n4: That's more or less what I had in mind and wanted to achieve in the LSP case too: that all the tricky stuff is handled by the framework. The difference of not starting out on the UI thread is a good point, though. Let's discuss this further once you are through with the diff.

mor-n4: After thinking about this some more, one additional comment about the deadlock scenario:

> Now if two circular dependent editors run tasks that will be re-entrant on the executor service, we will dead-lock. How can we guarantee that this is not going to happen?

With `OpenFilesManager`/`LSPExecutorService` a deadlock can occur in the scenario you mention only if one of the tasks being submitted/executed via `#runInOpenFileContext()` synchronously waits for the result of another task it (transitively) submitted. But this is easy to avoid because code that requires the result can be put into a completion stage (i.e. with `#thenApply()` etc. on the completable future returned by `#runInOpenFileContext()` when submitting the transitive task).

The only time we have to wait for a result is when existing API forces us to do so. So far, I have only found one case of this: `#readOnly(URI, IUnitOfWork<R, ResourceSet>)` in `IReferenceFinder.IResourceAccess` which is, unfortunately, used in LSP for more purposes than just reference finding. Hopefully we can refactor this or make sure it won't be a problem in other ways.

As a side note: this situation seems to be different with `XtextDocument` and Eclipse editors, because `XtextDocument#readOnly|modify()` return the result itself, not a future.

Having said all this, my comment above that `#runInOpenFileContext()` is not intended for use with fine-grained, composable operations still applies. When in the context of open file A the resource of another file B is required, that resource can simply be loaded into the resource set of A's open file context without switching contexts (at least in the standard case and once an IExternalContentSupport is in place).

szarnekow: > i.e. with #thenApply()

Wouldn't this escape the safety net that the OpenFilesManager tries to implement, e.g. only a single task with a given ID may run?
If I submit such a task and do a thenApply on the future, my previously isolated environment is no longer available. Or do I miss something?

Anyhow, if this is not the desired use-case, my feedback is more of academic value than practically relevant.

mor-n4: > ... my previously isolated environment is no longer available.

True! If access to the resource/document is required in the later stage as well, another `#runInOpenFileContext()` is needed. It's a bit nasty, but the same (or very similar) situation exists for the implementation of the transitive task being submitted to run in context of resource B: that must not access resource A either, even though it will often have access to the object reference in the code when implemented as a closure.

> Anyhow, if this is not the desired use-case, my feedback is more of academic value than practically relevant.

In any case, this discussion was very helpful for me! I will try to extract 2-3 recommendations and add them to the API doc somewhere.

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/concurrent/ConcurrentIssueRegistry.java
mor-n4: True, or a concurrent list could be used for `listeners`. But when making classes thread safe I usually try to stick to one of two approaches: either make fields thread safe by using concurrent collections, etc. or use synchronize on/in methods. That's why I wanted to stick to using "synchronize" even in the simple case of field `listeners` and even used `this` as argument to synchronize instead of just `listeners`. But this might be overly cautious, idk. What do you think?

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/concurrent/LSPExecutorService.java
mor-n4: `#pollNextPending()` is iterating over the elements and removing elements from other places than just head or tail, so its more used like a list.

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/concurrent/LSPExecutorService.java
mor-n4: Good catch!! Thanks!

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/findReferences/XWorkspaceResourceAccess.java
mor-n4: Removed.

mor-n4:
FYI: I noticed a problem. Currently an open file can see all resources (even those from other projects that are not in the list of dependencies of the project containing the open file). I thought this is taken care of on the language level by the scoping, but it seems an `IAllContainersState` has to be registered as adapter on the resource set for this to work. This isn't trivial because now more information has to be provided by the `LSPBuilder` to the `OpenFilesManager` (in a thread safe manner). I have implemented a solution but there are still some test failures and it needs clean up, so I won't push it here for now.

mor-n4:
Thanks a lot for part 2!!

> I'm still not entirely sure that I understood the threading model here. [...]

1) Simplicity: it's definitely true that this new approach is more complex. But I think that is a necessary price to pay for having the builder being independent of the editor functionality (and also each editor being independent of the other editors). The core idea of the original approach (i.e. trying to keep everything single-threaded as far as possible) made sense and would have its advantages, if we had not run into the limitations we discussed earlier (esp. editor functionality has to "wait" for the builder and read requests running on shared mutable state while it's difficult to guarantee they do not accidentally modify anything, esp. w.r.t. EMF models, adapter lists, caches, ...).

   My hope was that after some iterations we can get to a point where all increased complexity is in the framework (and there only in a few classes) and the code of the builder and editing functionality will become even simpler than before, because, for example, the builder no longer has to be adjusted to the needs of the editing functionality (e.g. mode `doGenerate==false` might be unnecessary). Also, I guess your ideas for parallelizing the builder might be easier to implement.

2) Clear ownership: let me try to summarize the separation/ownership I had in mind:

   * the builder is single-threaded and does not share any state with the outside: resource sets, project configurations/descriptions, etc. are all private to the builder (class `LSPBuilder` is the thread-safe entry point, classes `XWorkspaceManager`, ... have plain old single-threaded code),
   * each open file forms another such unit which is single-threaded and has its own private resource set, etc. (class `OpenFilesManager` is the thread-safe entry point for all open files together, `OpenFileContext` and the tasks running within those contexts contain plain old single-threaded code),
   * the only communication between those encapsulated units happens via the two classes `ConcurrentChunkedIndex` and `ConcurrentIssueRegistry`. They are thread safe, so they can be read from and written to at any time from any thread. They each own the few fields they contain.

   In addition, `LSPExecutorService` is just a minor utility. The queueIds it adds to an ordinary executor service are just used to sequentialize tasks of builder/open editors without the need for a separate thread per builder / open editor.

> [...] where a lot of the synchronization happens on the document which is - given that it has text and resource - rather self contained.

I suppose this is what you already mentioned above when proposing something like an `XtextLspDocument`. If I understand correctly, it would roughly mean renaming `OpenFileContext` to `XtextLspDocument`, making it thread safe and moving most of the logic from `OpenFilesManager` into `OpenFileContext` (just as a very rough summary). As an optional second step, `ConcurrentChunkedIndex` and maybe also `ConcurrentIssueRegistry` could be removed and the builder would directly invoke methods on `OpenFileContext` to update the persisted state index. I'm a bit wary about the second step but the first is definitely something we should consider; the exact way `IResourceDescription`s are forwarded through `OpenFilesManager` and fed into `OpenFileContext` is work-in-progress, anyway.

But I feel we have to continue with this particular topic offline, my comments are turning into walls-of-text already :-)

I'm gonna look through the individual code comments later today.

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/openfiles/OpenFileContext.java
```diff
@@ -0,0 +1,374 @@
...
+		}
+		if (!mainResource.isLoaded()) {
+			throw new IllegalStateException("trying to refresh a resource that is not yet loaded: " + mainURI);
+		}
+
+		// TODO the following is only necessary for changed files (could be moved to #updateDirtyState())
+		ResourceSet resSet = getResourceSet();
+		for (Resource res : new ArrayList<>(resSet.getResources())) {
+			if (res != mainResource) {
+				res.unload(); // FIXME better way to do this? (unload is expensive due to re-proxyfication)
```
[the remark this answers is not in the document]
mor-n4: With that approach: doesn't the mainResource then keep object references into the removed resources? Or is this unharmful, due to the code that follows (i.e. `mainResource.update(...)`)?

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/openfiles/OpenFilesManager.java
```diff
@@ -0,0 +1,384 @@
...
+		// to #discardOpenFileInfo() on the queue (note: this does apply to tasks being submitted after this method
+		// returns and before #discardOpenFileInfo() is invoked).
+		// FIXME GH-1774 reconsider sequence when closing files
+		return runInOpenFileContextVoid(uri, "closeFile", (ofc, ci) -> {
+			discardOpenFileInfo(uri);
+		});
+	}
+
+	/** Tries to run the given task in the context of an open file, falling back to a temporary file if necessary. */
+	public synchronized <T> CompletableFuture<T> runInOpenOrTemporaryFileContext(URI uri, String description,
```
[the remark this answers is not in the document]
mor-n4: The synchronization is only required and the corresponding lock on `this` is only retained while scheduling/submitting the task, not until that task starts nor while it is running.

For example, in the method you commented on, there are the lines:
```java
		if (isOpen(uri)) {
			return runInOpenFileContext(uri, description, task);
		} else {
			return runInTemporaryFileContext(uri, description, task);
		}
```
We have to ensure that after `isOpen(uri)` returns true, the file is still open in the following line when invoking `runInOpenFileContext(...)`. Other `runIn...()` methods contain similar cases.

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/openfiles/OpenFilesManager.java
```diff
@@ -0,0 +1,384 @@
...
+		// TODO GH-1774 closing a file may lead to a change in other open files (because they will switch from using
+		// dirty state to using persisted state for the file being closed)
+	}
+
+	/**
+	 * Tells whether the given resource set is one that is used to process an open file managed by this
+	 * {@link OpenFilesManager}.
+	 */
+	public synchronized boolean isOpenFileResourceSet(ResourceSet origin) {
+		OpenFileContext ofc = findOpenFileContext(origin);
```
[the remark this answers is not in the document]
mor-n4: Method `#findOpenFileContext()` called in first line inspects field `openFiles`. We need synchronized on the outer method (not just on `#findOpenFileContext()`) to guarantee the corresponding file is still open and thus the return value is still up-to-date once this method returns.

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/openfiles/OpenFilesManager.java
```diff
@@ -0,0 +1,384 @@
...
+	}
+
+	protected synchronized ResourceDescriptionsData createPersistedStateIndex() {
+		// TODO GH-1774 performance? (consider maintaining a ResourceDescriptionsData in addition to persistedState)
+		return new ResourceDescriptionsData(persistedState.getAllResourceDescriptions());
+	}
+
+	/** Creates an index containing the persisted state shadowed by the dirty state of all open files. */
+	public synchronized ResourceDescriptionsData createLiveScopeIndex() {
+		ResourceDescriptionsData result = createPersistedStateIndex();
```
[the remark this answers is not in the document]
mor-n4: As said in my first reply from today, the details how resource descriptions are channeled through `OpenFilesManager` and passed into `OpenFileContext` can probably be simplified. Let's discuss this as part of your "`XtextLspDocument`" proposal.

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/LSPIssueConverter.java
```diff
@@ -0,0 +1,86 @@
...
+		return result;
+	}
+
+	// TODO GH-1537: Remove this function when org.eclipse.xtext.validation.Issue.IssueImpl#lineNumber and #column are
+	// initialized with '0'
+	private int toZeroBasedInt(Integer oneBasedInteger) {
+		return oneBasedInteger == null ? 0 : (oneBasedInteger - 1);
+	}
+
+	protected XDocument createDocument(XtextResource resource) {
```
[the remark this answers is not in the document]
mor-n4: iirc I just moved this here from elsewhere. In any case, using the node model seems better, esp. because the node model caches some values. But it seems like a non-trivial change and I don't want to mix this with all the other topics on this branch. Added a TODO, for now.

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/WorkspaceAwareCanLoadFromDescriptionHelper.java
```diff
@@ -28,8 +29,16 @@
 	@Inject
 	private XWorkspaceManager workspaceManager;
 
+	@Inject
+	private OpenFilesManager openFileManager;
+
 	@Override
 	public Resource createResource(ResourceSet resourceSet, URI resourceURI) {
+		boolean inOpenFileContext = openFileManager.isOpenFileResourceSet(resourceSet);
+		if (inOpenFileContext) {
```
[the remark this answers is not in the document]
mor-n4: Should be fine, because it is basically single-threaded code, as long as resource sets are not passed around across contexts (and then all hell breaks loose, anyway ;). However, these two lines can be made more explicit by just saying `boolean inOpenFileContext = openFilesManager.currentContext() != null;`.

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/findReferences/XWorkspaceResourceAccess.java
```diff
@@ -25,27 +29,36 @@
...
 
 	/**
-	 * @param workspaceManager
-	 *            the workspace manager
+	 * @param languageServer
+	 *            the language server
 	 */
-	public XWorkspaceResourceAccess(XWorkspaceManager workspaceManager) {
-		this.workspaceManager = workspaceManager;
+	public XWorkspaceResourceAccess(XLanguageServerImpl languageServer) {
```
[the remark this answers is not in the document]
mor-n4: I think `XWorkspaceResourceAccess` and `ILanguageServerAccess` require some refactoring anyway. In particular, they seem to be (partially) redundant. But I'd like to do this in a separate step. Added a TODO.

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/openfiles/OpenFileContext.java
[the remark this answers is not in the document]
szarnekow: Exactly: mainResource.update() should do a full reparse so all the objects are gone. There wouldn't be a need to to proxify any objects.
If it's a partial reparse, things become more annoying and we'd need to go for the full proxy dance. I'd rather prefer to avoid that. Do we disable partial parsing already in the AntlrParser via `isReparseSupported`?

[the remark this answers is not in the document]
mor-n4: Yes, `N4JSSemicolonInjectingParser` overrides `isReparseSupported()` to always return `false`. But this is only N4JS. The code here has to work with all Xtext languages, if I'm not overlooking something.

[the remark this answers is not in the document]
szarnekow: I'd rather focus on making this work with N4JS atm. The current change is already huge and therefore prone to take a few month before it has been brought into a shape that fits all languages (and other constrains). 
So in general: yes, you are right. On the other hand, unloading is so very expensive, that we probably want to avoid this at the expense of always doing a reparse(..)

mor-n4:
Above changes were merged in another PR. Remaining topics were added to user stories.
Closing this now.

PR #1783, opened 2020-06-30
GH-1774: As a smith I want to separate editing functionality from the LSP builder (part 1)

mor-n4:
See #1774.

This is only part 1. More clean up, refactoring, API documentation, tests, etc. will be provided as part 2. Also the overall approach will be further discussed during July, possibly leading to design changes/refinements in some areas.

Also this does not yet remove (all) now unused/obsolete logic in the LSP builder. This is also done in part 2.

Feedback on these changes was provided by @szarnekow in a separate branch/PR, see #1782. Thanks a lot for that!


mor-n4:
Builds green, merging ...

Issue #1788, opened 2020-07-02
Imports cannot be resolved in newly created / renamed file when opened


mor-n4:
When an N4JS source file is created or renamed in VSCode and then opened, imports in the file cannot be resolved. This is fixed as soon as the file is being saved for the first time.

The issue is caused by method `OpenFileAllContainersState#getContainerHandle(URI)`, which cannot find the container handle for a URI that does not yet exist in the persisted index.

This is partly caused by the fact that VSCode does not send a `didChangeWatchedFiles` notification to the server after the create/rename, which might be caused by an incorrect configuration during server initialization. However, we should not rely on that in any case, because other LSP clients might support open files which do not exist on disk.

PR #1792, opened 2020-07-06
GH-1788: Imports cannot be resolved in newly created / renamed file when opened

mor-n4:
See #1788.


mor-n4:
Builds green, merging ...

Issue #1797, opened 2020-07-08
Headless builder emits some errors/warnings multiple times


mor-n4:
The headless builder prints out errors/warnings on a per-file basis but often several times: it seems that if a file contains _n_ issues, then the URI of the file and its issues are printed _n_ times.

For example:
```
packages/ProjectA/src/A.n4js
  ERR 1:20     string is not a subtype of number.
packages/ProjectA/src/B.n4js
  WRN 3:5      The local variable a is never used
  WRN 4:5      The local variable b is never used
packages/ProjectA/src/B.n4js
  WRN 3:5      The local variable a is never used
  WRN 4:5      The local variable b is never used
```

szarnekow:
https://github.com/eclipse/n4js/blob/master/plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/ProjectStatePersister.java#L311

https://github.com/eclipse/n4js/blob/master/plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/ProjectStatePersister.java#L313

Here we are serializing n times, where n is total number of entries in the Multimap. It should be the size of the `keySet() = Set<URI>` rather than `multimap.size()` and `multimap.keys() = Multiset<URI>` - probably not related but I'm not terribly sure.

mor-n4:
It's caused by this line:
https://github.com/eclipse/n4js/blob/51b5740f87eaae44455c23820c093301296e895f/plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/XProjectManager.java#L159

(so also an invocation of `Multimap#keys()`)

But great that you mention that other invocation. I will check that one too!

mor-n4:
FYI: 505999c1c0e82098a932a93392b0da5a7ea75626

BTW: we have a third invocation of `Multimap#keys()` in the code base, but it's followed by `.head`, so should be safe :)
https://github.com/eclipse/n4js/blob/51b5740f87eaae44455c23820c093301296e895f/plugins/org.eclipse.n4js/src/org/eclipse/n4js/validation/validators/N4JSTypeValidator.xtend#L786

PR #1798, opened 2020-07-08
GH-1797: Headless builder emits some errors/warnings multiple times

mor-n4:
See #1797.


mor-n4:
Builds green, merging ...

PR #1802, opened 2020-07-20
Fix security vulnerability of lodash 4.17.15

mmews-n4:


mmews-n4:
Thanks! Builds green.

Issue #1784, opened 2020-06-30
As a smith I want to separate editing functionality from the LSP builder (part 2)


mor-n4:
For part 1, see #1774.

The main implementation is already on master, see #1783, but is still experimental and requires finalization.

To be done as part of this user story:
* [x] change existing LSP builder to remove all now unused/obsolete functionality
  (e.g. remove unnecessary synchronized declarations, replace concurrent collections by ordinary collections, the doGenerate==false mode, build listeners, duplicate storage of issues XProjectStateHolder),
* [x] review all `@Singleton` annotations of the related classes and consider reducing the number of singletons
  (to improve encapsulation of builder-internal data structures, among other advantages),
* [x] check if there is a simpler solution to cross-project visibility than tracking the container structure in `OpenFileContext`,
* [x] consider creating an `ImmutableResourceDescriptionsData` and using it in `ConcurrentChunkedIndex`
* [x] consider additional cancellation when an index update arrives in an open file context
  (warning: we probably don't want a running completion request, etc. to be interrupted by index updates from the builder),
* [x] consider aggregating index updates and batch processing them if they arrive in very short intervals
* [x] go through all to-do's in the code (marked with `GH-1774`)
  (some are not really part of editor/builder separation, create dedicated user stories for them),
* [x] compare performance to pre-`GH-1774` version
* [x] recover old functionality:
  * [x] performance analysis infrastructure as implemented in old N4JSRequestManager
  * [x] tests in old XRequestManagerTest
* [x] further clean up / refactoring (e.g. consider introducing a top-level class `LSPManager` or `LSPFrontEnd`),
* [x] more API documentation,
* [x] write tests (esp. for `LSPExecutorService`).

mor-n4:
Some remarks:
* `@Singleton`s:
  --> keeping them as they are, for now; refactoring would require changes beyond the scope of this topic.

PR #1803, opened 2020-07-21
GH-1784: As a smith I want to separate editing functionality from the LSP builder (part 2)

mor-n4:
See #1784.


mor-n4:
Thanks!
Builds green-ish, merging ...

Issue #1631, opened 2020-01-22
Allow imported any+ namespaces to be used as functions


mmews-n4:
Mind the example below:

```ts
import * as lib_decompress+ from "decompress";
// FIXME XPECT noerrors --> "Not a function or method: lib_decompress." at "lib_decompress"
lib_decompress();
```

mor-n4:
After looking into this a bit: I think the namespace itself will never be a function (when using ES6 imports).

If there is a `plain.js` with either
```javascript
module.exports = ()=>console.log('hello from plain!');
```
or
```javascript
export default ()=>console.log('hello from plain!');
```
and another file `Main.n4js` with
```javascript
import * as P+ from "plain"
P();
```
will fail at runtime with `TypeError: P is not a function`. Instead, the code `P.default()` in the second line would work.

So what we actually want is probably a "dynamic default import", such as:
```javascript
import P+ from "plain" // not supported yet
P();
```

PR #1809, opened 2020-07-22
GH-1816: Type references pointing to namespaces of dynamic imports are not flagged as dynamic

mor-n4:
See #1816.

This PR implements a minor, largely unrelated improvement done while looking into GH-1631. See GH-1816 for details.


mor-n4:
Builds green, merging!

Issue #1811, opened 2020-07-23
As a developer I want to investigate missing events for watched files


mor-n4:
It seems VSCode does not send some file change events it is supposed to send. This story is about
1. going through notes and collecting problematic cases, verifying them again, and compiling a list.
2. finding out what the cause might be (maybe just a configuration problem on our side?).
3. solving the problem (if possible within the given time frame) or creating a more concrete follow-up task.

mor-n4:
This was caused by a configuration problem on our side, which was fixed by this task.
A task was created for further improvements of the file watching: `GH-1821`.

PR #1813, opened 2020-07-23
GH-1811: As a developer I want to investigate missing events for watched files

mor-n4:
See #1811.


mor-n4:
Builds green, merging!

PR #1817, opened 2020-07-27
GH-1723: As a smith I want to also publish n4js-cli and vscode extension in cases where only the n4jsc.jar changes (not n4js-libs)

mmews-n4:
see: #1723


n4js-libs/packages/n4js-cli/src/n4js/index.n4js
```diff
@@ -73,118 +68,233 @@ export public function n4jsc(
...
 
-    if (!vmOptions) {
-        vmOptions = {};
+/**
+ * Returns either:
+ * - the value of the environment variable 'N4JSC_JAR',
+ * - the value of property 'n4jscjar' of the n4js-cli configuration file,
+ * - or null.
+ */
+export public function findN4jscJarConfigProperty(workingDir: string, log: {function(string)}) : string {
```
mor-n4, line 208: The VSCode extension passes in the VSCode workspace root folder as "workingDir". But usually the user has some flexibility of choosing the VSCode workspace root folder and with certain combinations a n4js-cli config file might be overlooked.

For example: n4js-cli config file is located in root folder of the yarn workspace project (= root folder of Git repo) and the user opens its parent folder as VSCode workspace root folder.

(if people put the n4js-cli config file into projects in the "packages" folder there are more such cases)

n4js-libs/packages/n4js-cli/src/n4js/index.n4js
```diff
@@ -73,118 +68,233 @@ export public function n4jsc(
...
-            const isBool = typeof value === "boolean";
-            if (isBool && !value) {
-                continue; // skip false boolean flags
-            }
-            args.push((key.length > 2 ? "--" : "-") + key);
-            if (!isBool) {
-                args.push(String(value));
+    const fileName = path.basename(customN4jscName);
+    const jarInBinDir = path.resolve(BIN_DIR, fileName);
+    if (fs.existsSync(jarInBinDir)) {
```
mor-n4, line 268: With this check whether a jar was already downloaded, all URIs will be "ignored" that have a last segment equal to the file name of any previously downloaded jar. In particular, this applies to all URIs with a last segment equal to the default name "n4jsc.jar". This might be confusing for the user.

PR #1823, opened 2020-07-28
GH-1793: Implemented robust approach to proper recovery after cancellation

### LINKED ISSUE #1793: As a smith I want to review and improve the cancellation behaviour of the LSP builder
The cancellation behaviour of the LSP builder needs a thorough review and improvement / clean-up.

Concrete issues that need to be fixed:
* handling of cancellation
* the following code is bogus
    https://github.com/eclipse/n4js/blob/cba043f39d30627900ce8eb090da17b6ec60c78c/plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/build/XStatefulIncrementalBuilder.java#L162-L165
    (leads to exceptions in case of cancellation; see `GH-1775` for details)
* cancellation of one build can lead to the following build not building some files that actually require rebuilding.
    A temporary fix for this was implemented in GH-1785, but a better solution must be found. That task also added a test that can be used to reproduce the problem with a small example, see `IncrementalBuilderCancellationTest`.
    Search code base for `GH-1793`.

The above issues should be used as starting points for an overall review of the builder code. Also a simplification of the code in some places should be considered, if possible.

szarnekow:
- Reset the index of the resource set to the previous state
- Schedule all previously affected resources for for a build to ensure we fix previously sent validation issues in a subsequent run

closes #1793


plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/build/XBuildRequest.java
```diff
@@ -191,6 +199,29 @@ public void afterDelete(URI file) {
...
+	 * Getter.
+	 */
+	public AffectedListener getAffectedListener() {
+		return affectedListener;
+	}
+
+	/**
+	 * Notify the request that we are going to work with the given URI.
+	 */
+	public void markAffected(URI uri) {
```
szarnekow, line 219: TODO rename to afterDetectedAsAffected

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/build/XWorkspaceBuilder.java
```diff
@@ -274,21 +274,30 @@ private BuildTask toBuildTask(WorkspaceChanges workspaceChanges) {
...
-							allBuildDeltas, cancelIndicator);
-				} catch (BuildCanceledException e) {
-					// TODO GH-1793 remove this temporary solution
-					dirtyFiles.addAll(e.incompletelyBuiltFiles);
-					e.rethrowOriginalCancellation();
-					throw new IllegalStateException("should never get here");
+					projectResult = projectBuilder.doIncrementalBuild(new IBuildRequestFactory() {
+
+						@Override
+						public XBuildRequest getBuildRequest(String projectName, Set<URI> changedFiles,
```
szarnekow, line 289: Refactor to named class

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/build/XWorkspaceBuilder.java
```diff
@@ -53,6 +53,31 @@
...
+		private AffectedResourcesRecordingFactory(IBuildRequestFactory delegate) {
+			this.delegate = delegate;
+			this.affected = new HashSet<>();
+		}
+
+		@Override
+		public XBuildRequest getBuildRequest(String projectName, Set<URI> changedFiles,
+				Set<URI> deletedFiles, List<Delta> externalDeltas) {
+			XBuildRequest result = delegate.getBuildRequest(projectName, changedFiles, deletedFiles, externalDeltas);
+			result.setAffectedListener(affected::add);
```
mor-n4, line 69: I wonder if the anonymous class created here at `affected::add` might be a good place for breakpoints or temporary logging during debugging. In that case it would be good to write this more lengthy.

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/build/XWorkspaceBuilder.java
```diff
@@ -274,21 +299,27 @@ private BuildTask toBuildTask(WorkspaceChanges workspaceChanges) {
...
-				} catch (BuildCanceledException e) {
-					// TODO GH-1793 remove this temporary solution
-					dirtyFiles.addAll(e.incompletelyBuiltFiles);
-					e.rethrowOriginalCancellation();
-					throw new IllegalStateException("should never get here");
+					projectResult = projectBuilder.doIncrementalBuild(
+							recordingFactory,
+							projectDirty,
+							projectDeleted,
+							allBuildDeltas,
```
mor-n4, line 316: while at it, please consider renaming `allBuildDeltas` to `allBuiltDeltas` (I guess the name means it contains deltas that "were built")

tests/org.eclipse.n4js.ide.tests/src/org/eclipse/n4js/ide/tests/builder/IncrementalBuilderCancellationTest.xtend
```diff
@@ -28,13 +33,14 @@ import static org.junit.Assert.*
  */
 class IncrementalBuilderCancellationTest extends AbstractIncrementalBuilderTest {
 
+	/* Cancel after n resources */
 	private static final AtomicInteger cancelCounter = new AtomicInteger(0);
```
mor-n4, line 37: Proposal: "Iff greater zero, then cancel while building the n-th resource."

tests/org.eclipse.n4js.ide.tests/src/org/eclipse/n4js/ide/tests/builder/IncrementalBuilderChangesTest.xtend
```diff
@@ -225,4 +227,36 @@ class IncrementalBuilderChangesTest extends AbstractIncrementalBuilderTest {
 		joinServerRequests();
 		assertNoIssues();
 	}
+	
+	@Ignore("https://github.com/eclipse/n4js/issues/1822")
+	@Test
+	def void testTransitivelyAffected() {
```
mor-n4, line 233: Nice find!

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/build/XWorkspaceBuilder.java
```diff
@@ -274,21 +299,27 @@ private BuildTask toBuildTask(WorkspaceChanges workspaceChanges) {
...
+							cancelIndicator);
+				} catch (Throwable t) {
+					// re-schedule the affected files since a subsequent build may not detect those as affected
+					// anymore but we may have produced artifacts for these already
+					this.dirtyFiles.addAll(recordingFactory.getAffectedResources());
```
mor-n4, line 321: Could it be necessary to distinguish between dirty and deleted here? The root source of the affected resources is a list of deltas, so in theory there could be deltas with new==null.

At least this seems possible if the delete is detected late during resource loading in the `#executeClustered()` method (see `XClusteringStorageAwareResourceLoader#loadResource()`).

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/build/XWorkspaceBuilder.java
szarnekow: Do you think that `XBuildRequest. afterDetectedAsAffected` is a fair location for a breakpoint?

mor-n4: Maybe a bit more difficult to find when working on the XWorkspaceBuilder, but it should do the trick.

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/build/XWorkspaceBuilder.java
szarnekow: according to the JavaDoc these are deltas that have not been taken into account yet by downstream projects. I'll try to find a better name, since `all` is certainly a misleading prefix, too.

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/build/XWorkspaceBuilder.java
szarnekow: If there was an intermitted deletion or a change, we'd see a notification from the client for this. Here we only need the files, that have been built but not changed on disk to make sure we validate these again in a subsequent built (that is initiated with a different set of deltas if in doubt).
Therefore I don't think we need to record deletions in addition to the default mechanism.

tests/org.eclipse.n4js.ide.tests/src/org/eclipse/n4js/ide/tests/builder/IncrementalBuilderCancellationTest.xtend
szarnekow: :+1:

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/build/XBuildRequest.java
```diff
@@ -54,6 +54,12 @@
 
 	private CancelIndicator cancelIndicator = CancelIndicator.NullImpl;
 
+	/** Listener for affected resources */
+	public static interface AffectedListener {
+		/** Called if a resource is considered to be affected and will be build */
```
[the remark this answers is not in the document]
szarnekow: Done

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/build/XWorkspaceBuilder.java
szarnekow: `toBeConsideredDeltas`

PR #1825, opened 2020-07-28
GH-1824: The LSP server hangs in an infinite loop after the io streams have been closed when started with --stdio

### LINKED ISSUE #1824: The LSP server hangs in an infinite loop after the io streams have been closed when started with --stdio
Steps to reproduce
* Start VSCode with N4JS extension
* Modify inside extension code in `ExtensionProvider.getDeactivate()`: remove `n4jscProcess.kill();`.
* Reload VSCode so that modification takes effect
* Quit VSCode
* See with `jps` that `n4jsc.jar` is still running
* See with `jstack` that one process is inside a method of `LspServer`
* Notice that this process does never end

mmews-n4:
closes #1824


mmews-n4:
Thanks!

Issue #1842, opened 2020-08-05
Using method XtextResource#reparse() when refreshing resources causes exceptions


mor-n4:
After replacing `XtextResource#update()` by `XtextResource#reparse()` exceptions are thrown in `ResourceTaskContext#refreshContext()` when an existing resource is being changed. It seems that `#reparse()` in not behaving as expected / as it should, so we should switch back to `#update()`, for now.

```
java.lang.StringIndexOutOfBoundsException: begin 67, end 104, length 44
	at java.base/java.lang.String.checkBoundsBeginEnd(Unknown Source)
	at java.base/java.lang.String.substring(Unknown Source)
	at org.eclipse.xtext.nodemodel.impl.AbstractNode.getText(AbstractNode.java:157)
	at org.eclipse.xtext.nodemodel.util.NodeModelUtils.getTokenText(NodeModelUtils.java:421)
	at org.eclipse.n4js.resource.N4JSLinker.createProxy(N4JSLinker.java:270)
	at org.eclipse.n4js.resource.N4JSLinker.createAndSetProxy(N4JSLinker.java:223)
	at org.eclipse.n4js.resource.N4JSLinker.installProxies(N4JSLinker.java:164)
	at org.eclipse.n4js.resource.N4JSLinker.installProxies(N4JSLinker.java:170)
	at org.eclipse.n4js.resource.N4JSLinker.installProxies(N4JSLinker.java:130)
	at org.eclipse.n4js.resource.N4JSLinker$1.process(N4JSLinker.java:103)
	at org.eclipse.n4js.resource.N4JSLinker$1.process(N4JSLinker.java:1)
	at org.eclipse.xtext.util.concurrent.IUnitOfWork$Void.exec(IUnitOfWork.java:37)
	at org.eclipse.xtext.util.OnChangeEvictingCache.execWithoutCacheClear(OnChangeEvictingCache.java:128)
	at org.eclipse.n4js.resource.N4JSLinker.doLinkModel(N4JSLinker.java:91)
	at org.eclipse.xtext.linking.impl.AbstractCleaningLinker.linkModel(AbstractCleaningLinker.java:51)
	at org.eclipse.xtext.resource.XtextResource.doLinking(XtextResource.java:340)
	at org.eclipse.xtext.linking.lazy.LazyLinkingResource.doLinking(LazyLinkingResource.java:121)
	at org.eclipse.n4js.resource.N4JSResource.doLinking(N4JSResource.java:655)
[...]

PR #1843, opened 2020-08-05
GH-1842: Using method XtextResource#reparse() when refreshing resources causes exceptions

mor-n4:
See #1842.


mor-n4:
Note to self: check why existing tests did not show this problem.

mor-n4:
Builds green, merging ...

Issue #1844, opened 2020-08-06
Index changes not propagated correctly after commenting out an interface


mor-n4:
When commenting out an interface, a file using that interface is not updated correctly.

To reproduce:

1. Create a simple N4JS project with two files `IDataCollection.n4js`:
   ```javascript
   export public interface IDataCollection {
       public size() : number;
   }
   ```
   and `IDataList.n4js`:
   ```javascript
   import {IDataCollection} from "IDataCollection";
   
   export public interface IDataList extends IDataCollection {
       public get(idx: number): Object;
   }
   ```
2. clean everything, open VSCode and wait for the initial build to complete.
3. open file `IDataCollection.n4js` and comment out the interface. Then save and close this file.
   --> an error will show up for the non-opened file `IDataList.n4js`
4. open file `IDataList.n4js`
   --> the error will disappear
5. touch the open file (i.e. edit without save)
   --> the error won't show up
6. save the open file
   --> now the error will show up

PR #1845, opened 2020-08-06
GH-1844: Index changes not propagated correctly after commenting out an interface

mor-n4:
See #1844.


mor-n4:
Thanks!
Builds green, merging ...

Issue #1666, opened 2020-02-14
As a developer I want n4jsc to emit test catalogs for every project (instead of emitting one file for all projects)


mmews-n4:
Currently, a single test catalog is created for all projects of the corresponding workspace. This is a problem since sometimes not all tests should be executed.

AC
* create a test catalog for each project of a workspace that contains at least one test case
* adjust mangelhaft to support globs or multiple test catalogs when being executed

PR #1850, opened 2020-08-07
GH-1666: As a developer I want n4jsc to emit test catalogs for every project (instead of emitting one file for all projects) 

mmews-n4:
see #1666


plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/server/build/N4JSProjectBuilder.java
```diff
@@ -0,0 +1,107 @@
...
+		writeTestCatalog();
+
+		return buildResult;
+	}
+
+	@Override
+	protected void doClearWithoutNotification() {
+		super.doClearWithoutNotification();
+
+		removeTestCatalog();
```
mor-n4, line 65: Removing the test catalog is similar to deleting the project state file or deleting the output files. Since none of those two things are done in `doClearWithoutNotification()` but instead in `doClean()` directly, I think it would be nicer to override `doClean()` for invoking `removeTestCatalog()`.

mmews-n4:
@dbo Please have a short look at the changes of mangelhaft

mor-n4:
As discussed: it's a bit unfortunate that we have a code duplication between `ProjectDiscoverer.n4js` and our `ProjectDiscoveryHelper` and some related code in `n4js-cli`, but this has to be revisited in a dedicated task in the future, if necessary.

However, best rename `ProjectDiscoverer.n4js` to `ProjectDiscoveryHelper.n4js` and add a comment to this file pointing to the `ProjectDiscoveryHelper` in Java, so that in the future we can spot and find this redundancy more easily.

n4js-libs/packages/n4js-mangelhaft-cli/src/n4js/org/eclipse/n4js/mangelhaft/runner/node/NodeTestRunner.n4js
```diff
@@ -47,17 +51,29 @@ class NodeTestRunner  {
...
+        } else if (isHTTP_re.test(options.testCatalog)) { // running IDE
             const req = await fetch(options.testCatalog, {
                 headers: {
                     "Content-Type": "application/vnd.n4js.assemble_test_catalog_req.tm+json"
                 }
             });
             testCatalog = (await req.json()) as TestInfos;
         } else {
-            testCatalog = readJsonFile(options.testCatalog) as TestInfos;
+            throw "Unknown value for option 'testCatalog': " + options.testCatalog;
```
dbo, line 76: not sure whether throwing string here is OK

Issue #1849, opened 2020-08-06
Clean does no longer clear issues


mor-n4:
Clean as executed as part of command "N4JS: Rebuild" does no longer clear the issues.

This is probably not critical, but it could cause other follow-up problems, make other problems more difficult to notice, or might make debugging more difficult.

PR #1851, opened 2020-08-10
GH-1849: Clean does no longer clear issues

mor-n4:
See #1849.

### DIFF
--- a/plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/build/ProjectBuilder.java
+++ b/plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/build/ProjectBuilder.java
@@ -55,6 +55,7 @@ import org.eclipse.xtext.workspace.ProjectConfigAdapter;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableListMultimap;
 import com.google.common.collect.ImmutableMap;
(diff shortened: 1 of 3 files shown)


plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/build/ProjectBuilder.java
```diff
@@ -316,9 +317,23 @@ public void doClean(AfterDeleteListener deleteListener, CancelIndicator cancelIn
...
+		doCleanIssues();
+
 		doClear();
 	}
 
+	/**
+	 * Send the necessary 'publishDiagnostics' notifications to the LSP client to removing all existing issues located
+	 * in this project. Will not remove those issues from the current project state in {@link #projectStateSnapshot}.
+	 */
+	protected void doCleanIssues() {
```
szarnekow, line 329: Should this be named `publishCleanIssues` to avoid that somebody accidentally calls this without a following doClear?

mor-n4: Good point, it does not make sense to call this new method without doing a doClear() afterwards. I now renamed the method to `doClearWithNotification()` and invoke `doClear()` right from within the new method. See 1c3480f.

mor-n4:
Thanks everyone! Merging ...

PR #1857, opened 2020-08-13
GH-1856: The command N4JS:Rebuild should rebuild also projects in the node_modules folder

### LINKED ISSUE #1856: The command N4JS:Rebuild should rebuild also projects in the node_modules folder
At the moment, projects in the node_modules folder are not rebuild for performance reasons. Since these projects actually do undergo changes, especially due to `yarn install` and manual changes due to testing, those changes need to be picked up and dealt with during a rebuild. When those changes happen at the moment, all `.n4js.projectstate` files need to be removed manually, which is a not obvious solution for the developer.

However, rebuilding projects in the node_modules folder is probably a temporary solution only, since the negative impact on the performance is significant.

mmews-n4:
closes #1856


mmews-n4:
@mor-n4 Please have a second look - Thank you!

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/build/ProjectBuilder.java
```diff
@@ -298,13 +297,10 @@ protected XBuildResult doBuild(
 	/** Deletes the contents of the output directory */
 	public void doClean(AfterDeleteListener deleteListener, CancelIndicator cancelIndicator) {
 		deletePersistenceFile();
+		doClearWithNotification();
```
mor-n4, line 300: It would be more consistent to clear issues only if indexOnly == false, because in case we are not responsible for performing validation and thus not responsible for _creating_ issues, we should not be responsible for _deleting_ issues. However, in case of issues it is probably better to ensure a clean slate.

Maybe add a comment along the lines of "clear issues on client side even in indexOnly mode, because ...".

Issue #1867, opened 2020-08-20
As a smith I want to automatically publish the VSCode extension to the marketplace instead of S3


mmews-n4:
extend Jenkins file to publish the VSCode extension to the marketplace whenever a new extension was built.

mmews-n4:
using branch GH-1866 for this userstory

Issue #1866, opened 2020-08-19
As a smith I want to manually publish the VSCode extension to the VSCode marketplace


mmews-n4:
At the moment, the extension for VSCode is only published as a package [here](https://download.enfore.com/n4js/). While this should remain unchanged, the extension should also be published on the VSCode marketplace. This userstory is about setting everything up and publish the latest available extension manually. In a follow-up userstory, this publishing should be made automatically.

mmews-n4:
published here: https://marketplace.visualstudio.com/items?itemName=n4js.n4js-vscode-extension

PR #1871, opened 2020-08-24
GH-1866B: As a smith I want to automatically publish the VSCode extension to the marketplace instead of S3

mmews-n4:
follow-up issue of #1867 (original branch was #1866)


mmews-n4:
Close since there are no changes

Issue #1821, opened 2020-07-27
As a developer I want incremental builds to be triggered when plain JS files change


mor-n4:
When plain JS files are changed, the LSP server should trigger an incremental build.

This is tricky, because the builder also changes plain JS files (i.e. the generated output files) and the corresponding events produced by the LSP client must not confuse the builder. Originally we wanted to solve this by dynamically registering for didChangeWatchedFiles events; however, we decided to use a different approach.

For reference, here the original issue title and description:

> ### As a smith I want the LSP server to dynamically register for didChangeWatchedFiles events
>
> The goal of this user story is to let the LSP server register for `didChangeWatchedFiles` notifications dynamically for all source folders of N4JS projects plus the `package.json` files via LSP's `DidChangeWatchedFilesRegistrationOptions`. Maybe some of the projects in node_modules folders need to be watched as well (direct dependencies of N4JS projects?).
>
> ### Background
>
> Currently, file change events sent from the client to the server are filtered on the client side by a static glob pattern (search for `createFileSystemWatcher` in file `ExtensionProvider.n4js` in the VSCode extension).
>
> The current pattern does not include plain JS files, but changes to those files may actually require an incremental build (for example, when a JS file is in the source folder of an N4JS project, the compiler has to copy it over to the project's output folder whenever its content changes). Adding all JS files (also in node_modules folders) may lead to performance issues in certain situations (e.g. yarn install) and therefore the dynamic registration is preferable.

PR #1877, opened 2020-08-26
GH-1821: As a developer I want incremental builds to be triggered when plain JS files change

mor-n4:
See #1821.

NOTE THE BRANCH NAME! (with a `b`)

### DIFF
--- a/plugins/org.eclipse.n4js/src/org/eclipse/n4js/xtext/workspace/WorkspaceConfigSnapshot.java
+++ b/plugins/org.eclipse.n4js/src/org/eclipse/n4js/xtext/workspace/WorkspaceConfigSnapshot.java
@@ -77,22 +77,28 @@ public class WorkspaceConfigSnapshot {
 	}
 
 	/**
-	 * Find the project with the given nested location.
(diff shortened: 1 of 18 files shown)


plugins/org.eclipse.n4js/src/org/eclipse/n4js/xtext/workspace/WorkspaceConfigSnapshot.java
```diff
@@ -76,26 +76,59 @@ public URI getPath() {
...
+	 * Find the project with the given name.
+	 */
+	public ProjectConfigSnapshot findProjectByName(String name) {
+		return name2Project.get(name);
+	}
+
+	/**
+	 * Finds the project having a path that is a prefix of the given URI.
+	 * <p>
+	 * Note the difference to {@link #findProjectContaining(URI)}!
```
mmews-n4, line 88: Could you state here that it returns true also for files outside of source folders but under this project base dir (instead of letting the reader figuring this out by himself).

IMHO: Since there is no client for this method, I would not introduce it. Also it would be easy to add it inside a sublass.

mmews-n4 (approved):
👍

mor-n4:
Builds green, merging ...

PR #1880, opened 2020-08-27
GH-1876: The projectstate should use relative URIs based on the location of the projectstate file

### LINKED ISSUE #1876: The projectstate should use relative URIs based on the location of the projectstate file
At the moment, the project state persists absolute file URIs which is a problem in case the project state would be published or in case a project folder is moved.

Also: Deal with the situation when a file mentioned in the project state does not exist anymore on disk.

mmews-n4:
closes #1876

### DIFF
--- a/plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/build/ProjectStatePersister.java
+++ b/plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/build/ProjectStatePersister.java
@@ -78,7 +78,7 @@ public class ProjectStatePersister {
 	 * <pre>
 	 * - Language version as per {@link #getLanguageVersion() getLanguageVersion}
 	 * - Number #r of resource descriptions
-	 * - #r times a serializable resource description as per {@link #writeResourceDescription(SerializableResourceDescription, DataOutput)}
+	 * - #r times a serializable resource description as per {@link #writeResourceDescription(SerializableResourceDescription, URI, DataOutput)}
 	 * - A mapping of generated URIs as per {@link Source2GeneratedMapping#writeExternal(java.io.ObjectOutput) Source2GeneratedMapping.writeExternal}
 	 * - Number #f of fingerprints per URI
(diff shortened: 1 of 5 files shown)


plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/build/ProjectStatePersister.java
```diff
@@ -485,10 +497,32 @@ private File getDataFile(IProjectConfig project) {
...
+		return absoluteURI.toString();
+	}
+
+	private URI makeAbsolute(URI baseURI, String relativeString) {
+		if (relativeString.contains(":/")) {
+			return URI.createURI(relativeString);
+		}
+		URI relativeURI = URI.createFileURI(relativeString);
+		URI absoluteURI = relativeURI.resolve(baseURI);
+		return new FileURI(absoluteURI).toURI();
```
mor-n4, line 514: Why converting it to a FileURI and back to an URI?

tests/org.eclipse.n4js.lang.tests/src/org/eclipse/n4js/ide/xtext/server/ProjectStatePersisterTest.java
```diff
@@ -47,6 +49,8 @@
 @SuppressWarnings({ "restriction", "deprecation" })
 public class ProjectStatePersisterTest {
 
+	static final URI BASE_URI = new FileURI(new File("/")).toURI();
```
mor-n4, line 52: Wouldn't it be better to run the tests with a more realistic value?

mor-n4 (approved):
:+1:

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/build/ProjectStatePersister.java
mmews-n4: FileURI performs several assertions that I wanted to reuse here.

PR #1887, opened 2020-09-02
GH-1886: Fix escaping issue when reading project state files

### LINKED ISSUE #1886: Fix escaping issue when reading project state files
When uris in project state files became relative (done in GH-1876), deserialising escaped special symbols like `#` which results in errors.

mmews-n4:
closes #1886


plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/xtext/server/build/ProjectStatePersister.java
```diff
@@ -509,7 +509,7 @@ private URI makeAbsolute(URI baseURI, String relativeString) {
 		if (relativeString.contains(":/")) {
```
mor-n4, line 509: Not part of this PR's topic, but we should consider finding some other approach here to tell apart absoulte and relative URIs.

mor-n4 (approved):
:+1:
Great you found this so quickly!

Issue #1900, opened 2020-09-14
Organize imports turns dynamic imports into ordinary namespace imports


mor-n4:
Organize imports removes the `+` after the namespace name of dynamic imports.

PR #1901, opened 2020-09-14
GH-1900: Organize imports turns dynamic imports into ordinary namespace imports

mor-n4:
See #1900.


mor-n4:
Builds green, merging ...

Issue #1905, opened 2020-09-16
LSP incremental builder does not rebuild all files if previous build was cancelled (again)


mor-n4:
When the incremental builder in LSP is being cancelled, e.g. by saving a file while an incremental build is in progress, the next build triggered does not build all necessary files, sometimes.

A similar problem related to cancellation was fixed in GH-1785 (improved in GH-1793). It is unclear whether the current problem is a regression or a separate issue. In any case, the tests introduced with the earlier tasks are passing.

Detailed steps to reproduce this require an internal code base, so omitted here.

PR #1906, opened 2020-09-16
GH-1905: LSP incremental builder does not rebuild all files if previous build was cancelled (again)

mor-n4:
See #1905.


mor-n4:
Builds green, merging ...

Issue #1909, opened 2020-09-17
As a developer I want the initial build to be non-cancelable


mor-n4:
The initial build must not be canceled. To avoid accidental cancellation of the initial build, it should be non-cancelable.

Note that this will mean that incremental build tasks caused by file change events during the incremental build (e.g. when user saves N4JS files while initial build is running) will pile up in the queue. This has to be properly supported even for large numbers of pending tasks.

### ACs

1. [x] make initial build non-cancelable (also full build during "N4JS: Rebuild" command).
2. [x] check how a large number of pending and already cancelled incremental build tasks are handled
   (in particular, avoid repeated, unnecessary workspace configuration updates).
3. [x] ensure that (re-)initial build leaves state fields (e.g. `#dirtyFiles`, `#toBeConsideredDeltas`) in a valid state
   (they should probably be cleared at the end of the initial build).
4. [x] tests, etc.

PR #1910, opened 2020-09-17
GH-1909: As a developer I want the initial build to be non-cancelable

mor-n4:
See #1909.


mor-n4:
Builds green, merging ...

Issue #1911, opened 2020-09-17
As a developer I want newly added imports of plain-JS projects to use complete module specifiers


mor-n4:
When content assist or organize imports adds new imports, the module specifier should be a "complete module specifier" (i.e. including the project name) iff the target module is from a plain-js project.

Can be tested, for example, with content assist for function `throttle` defined in `redux-saga/effects`.

Make sure this also works in case the target project is the definition project of a plain-js project!

PR #1912, opened 2020-09-21
GH-1911: As a developer I want newly added imports of plain-JS projects to use complete module specifiers

mor-n4:
See #1911.


mor-n4:
Builds green, merging ...

Issue #1917, opened 2020-09-29
ProjectDiscoveryHelper returns duplicate project paths and applies incorrect priority between node_modules folders


mor-n4:
#### Problem 1: Duplicate Paths

In case there is a dependency from one member project of a yarn workspace to another (i.e. the standard case of yarn workspaces), then the `ProjectDiscoveryHelper` returns duplicate paths for the same project.

To reproduce:
1. Create a project setup on disk as follows:
```
+ testyarn
  | node_modules  (as created by yarn)
  + packages
  | + ProjectA
  | | - package.json  (with dependency to ProjectB)
  | + ProjectB
  | | - package.json  (no dependencies)
  - package.json
```
2. Invoke `ProjectDiscoveryHelper#collectAllProjectDirs(Path...)`.

The method will return the following paths:
```
P1
P1/packages/D1
P1/packages/D2
P1/node_modules/D2     <--- this is a duplicate for project D2!
```

#### Problem 2: Incorrect priority between `node_modules` folders

In a yarn workspace context, the projects in the global `node_modules` folder should have priority over those in local `node_modules` folders of member projects, but the `ProjectDiscoveryHelper` returns those in the local folders.
(Note: this is add odds with node's module look-up semantics, but since N4JS does not support multiple projects with the same name, we have to prioritize the projects in the global folder; see `GH-1314`).

PR #1918, opened 2020-09-29
GH-1917: ProjectDiscoveryHelper returns duplicate project paths and applies incorrect priority between node_modules folders

mor-n4:
See #1917.


mor-n4:
Builds green, merging ...

Issue #1914, opened 2020-09-24
As a developer I want the LSP builder to consider newly added/removed dependencies to NPM packages


mor-n4:
When adding a new dependency to a plain-JS NPM package in the `package.json` of an N4JS project and running npm/yarn install to update the `node_modules` folder, the incremental builder is unable to resolve the new dependency.

The original error "project does not exist" still shows up ...
* after npm/yarn has completed,
* after an incremental build was triggered manually (e.g. by saving the package.json),
* after running the command "N4JS: Rebuild",
* even after executing "reload window".

However, the new dependency is correctly resolved when running a rebuild _after_ having reloaded the window.

The goal of this story is to investigate the causes, fix small/medium bugs, and create follow-up user stories for larger fixes/improvements (if necessary).

mor-n4:
The fact that the incremental builder does not automatically react to new projects in the `node_modules` folder is caused by VSCode: there is a user(!) configuration property called `files.watcherExclude` with a default value of:
```
**/.git/objects/**
**/.git/subtree-cache/**
**/node_modules/**
**/.hg/store/**
```
Thus, VSCode does not send file change events for contents of `node_modules` folders. Since this is a setting in the user preferences, VSCode extensions do not have access to that and cannot change it.

So we cannot react to changes in the `node_modules` folder without user interaction. A solution might be a "N4JS: Scan for new projects" command or even a "N4JS: Scan for new changes" command (the latter even checking if files have changed).

mor-n4:
Status:

After the fixes done as part of this user story, added/removed dependencies should now be properly picked up ...
* after running the command "N4JS: Rebuild", and also
* after executing "reload window" (i.e. after an initial build).

Having added/removed dependencies in the `node_modules` folder being picked up automatically by the incremental builder is not possible, because VSCode does not send file-changed-notifications for those, so this requires a user action; created a user story for adding a new "N4JS: Refresh" command for this purpose, see GH-1921.

PR #1919, opened 2020-09-29
GH-1914: As a developer I want the incremental builder to consider newly added/removed dependencies to NPM packages (PART 1)

mor-n4:
See #1914.

This includes two fixes that should make sure that a "N4JS: Rebuild" command will properly pick up newly added/removed dependencies. Does not fix the automatically triggered incremental build (problems related to that are caused by VSCode not sending file-changed-events and thus needs some other solution).


mor-n4:
Builds green, merging ...

Issue #1860, opened 2020-08-13
Issues persisted in the project state can get out of sync and become incorrect


mmews-n4:
I add two tests to reproduce this:
- `IncrementalBuilderFileChangesTest#testFixFileBetweenSessions`
- `IncrementalBuilderFileChangesTest#testBreakFileBetweenSessions`

mor-n4:
This happens not only if files are changed (as covered by above tests) but also when projects are being added/removed between LSP server sessions.

PR #1920, opened 2020-09-30
GH-1914 / GH-1860: As a developer I want the LSP builder to consider newly added/removed dependencies to NPM packages

mor-n4:
See #1914 and also #1860.

This includes several fixes that should make sure that an **initial build** will properly pick up newly added/removed dependencies.


mor-n4:
Builds green, merging ...

Issue #1923, opened 2020-10-05
Bogus "duplicate module" error in modules of implementation projects of API/Impl concept


mor-n4:
When using the API/Implementation concept, bogus errors show up in all modules of the implementation project, with the message: "A duplicate module MyModule is also defined in X" (where X is the path to the corresponding module in the API project).

This error shows up only in open editors under LSP, not in the builder (neither LSP builder nor headless builder).

PR #1924, opened 2020-10-05
GH-1923: Bogus "duplicate module" error in modules of implementation projects of API/Impl concept

mor-n4:
See #1923.


mor-n4:
Builds green, merging ...

Issue #1921, opened 2020-10-01
As a developer I want a refresh command for the LSP builder


mor-n4:
As a developer I want an LSP command that scans the entire workspace for changes (added/removed projects, added/removed source files, textual changes in source files, etc.) and triggers a minimal incremental build to bring everything up to date.

Motivation for having such a command:
1. VSCode does not send `didChangeWatchedFiles` notifications for the contents of folders called `node_modules`.
   Note: this can be configured in VSCode via user preference property `files.watcherExclude`, but only the user can do this and the default is to ignore changes in `node_modules` folders.
2. other clients might implement similar limitations or might not reliably support file watching or not at all.
3. there might be other cases that the builder state gets out-of-sync with the workspace on disk (e.g. OS not reporting changes, bugs in LSP server).

Running an "N4JS: Rebuild" command would solve the above cases, but that runs a full build and therefore will take much longer in many cases. Running a "Reload window" (and thus an initial build) is not an elegant alternative and might also rebuild more than necessary.

PR #1925, opened 2020-10-07
GH-1921: As a developer I want a refresh command for the LSP builder

mor-n4:
See #1921.


mor-n4:
Builds green, merging ...

Issue #1846, opened 2020-08-06
Initial build in LSP server shows more errors than the incremental build


mor-n4:
It seems there is an inconsistency between the initial build and incremental build in the LSP server: after removing an interface from a module and saving, certain errors show up after the initial build completes. Then, when running a clean and initial build (with command "N4JS: Rebuild"), additional errors show up. A comparison with the behavior in Eclipse suggests that the result of the incremental build is correct.

To reproduce:

1. create a yarn workspace with two projects `ProjectA` and `ProjectB`, with the latter having a dependency to the former.
2. add these files to `ProjectA`:
   IDataCollection.n4js
   ```javascript
   export public interface IDataCollection {
       public size() : number;
   }
   ```
   IDataList.n4js
   ```javascript
   import {IDataCollection} from "IDataCollection";
   
   export public interface IDataList extends IDataCollection {
       public get(idx: number): Object;
   }
   ```
   MainA.n4js
   ```javascript
   import {IDataList} from "IDataList";
   
   let l: IDataList;
   l = null;
   ```
3. add this file to `ProjectB`:
   MainB.n4js (same content as MainA.n4js):
   ```javascript
   import {IDataList} from "IDataList";
   
   let l: IDataList;
   l = null;
   ```
4. clean everything, open VSCode in root folder of yarn workspace project, and wait for initial build to complete.
5. open file `IDataCollection.n4js` and comment out the interface. Save and close this file.
   --> 2 errors show up in non-opened file `IDataList.n4js` (but no other errors)
6. this state is stable, i.e. you can open the two main files, edit them, save them and no additional errors will appear.
7. now execute command "N4JS: Rebuild"
   --> an additional error will show up in `MainB.n4js`
8. now open, touch, and save file `MainA.n4js`. Then close this file.
   --> an additional error will show up in this file (only after closing).

Potential cause: maybe the initial build does not update the index for resources with errors. Maybe this applies to both builds but in case of incremental build it does not show up because the old index state is retained for the resource with errors?

mor-n4:
For comparison:

1) when following the above steps in Eclipse, we never get the errors in the "Main*" files (even after a clean and full build).
2) when compiling the projects (after commenting out the interface) with the headless builder, we get the additional error in `MainB.n4js`.

mor-n4:
A test is available on branch `GH-1846`.

mor-n4:
The additional errors showing up in the full build seem related to this rule in `UserDataMapper#createUserData()`:
[...]

PR #1927, opened 2020-10-07
GH-1846: Initial build in LSP server shows more errors than the incremental build

mor-n4:
See #1846.


mor-n4:
Builds green, merging ...

Issue #1926, opened 2020-10-07
Provide more defaults in the IDE test infrastructure


mor-n4:
To make writing IDE tests easier and more convenient, more defaults should be added to `TestWorkspaceManager`. In particular:
1. projects requiring a dependency to `n4js-runtime` should have such a dependency added automatically to their package.json.
2. the runtime `n4js-runtime` should be provided by default in the node_modules folder (but only if it is required!).

In addition, the following convenience improvements should be implemented:
1. asserting issues in package.json files should be easier (esp. when issues in N4JS need to be asserted at the same time),
2. access to some frequently used constants should be possible easily

PR #1928, opened 2020-10-07
GH-1926: Provide more defaults in the IDE test infrastructure

mor-n4:
See #1926.


mor-n4:
Builds green, merging ...

Issue #1932, opened 2020-10-14
As a developer I want that running 'yarn install' will solve all issues in a running LSP server


mor-n4:
The following should work:
1. clone a large N4JS code base
2. start VScode and the LSP server & wait for initial build to complete
   -> many errors will show up, because NPM dependencies not installed yet
3. run `yarn install` (while LSP server is running) & wait for the incremental build(s) to complete
   -> all errors should be resolved and everything should be in a valid state
4. close & reopen VSCode
   -> initial build should be fast (i.e. all information should be loaded from the project state files)

Currently the above is not supported yet for large projects (some errors remain / invalid state).

### ACs

1. test the above for several large N4JS code bases.
2. investigate what causes the remaining problems.
3. solve the bugs / problems or create follow-up tickets (if too much effort).

mor-n4:
Note: since VSCode does not send `didChangeWatchedFiles` events for node_module folders, the above scenario cannot be fully supported. Instead, the new "N4JS: Refresh" command must be executed by the user after step 3 and only then all errors will be gone.

PR #1933, opened 2020-10-14
GH-1932: As a developer I want that running 'yarn install' will solve all issues in a running LSP server

mor-n4:
See #1932.


mor-n4:
Builds green, merging ...

Issue #1931, opened 2020-10-13
EMF resources created in wrong resource set in LSP builder


mor-n4:
In the LSP builder, each project has its own EMF resource set and all source files in that project will be represented as EMF resources in that resource set.

In some cases, resources are created in the wrong resource set. In other words, sometimes the resource set of a project builder for ProjectA contains resources with URIs for which the path of ProjectA is not a prefix.

Probably this is caused by the fact that a `ResourceLocator` registered on a resource (i.e. `WorkspaceAwareResourceLocator`) will only handle resource creation if it happens during a `ResourceSet#getResource(uri, true)` call, not if client code invokes `ResourceSet#createResource(uri)`.

### ACs

1. [x] add consistency checks and logging to spot such problems earlier, in the future.
2. [x] fix the bug.
3. [x] check whether after the above fix class `WorkspaceAwareCanLoadFromDescriptionHelper` is now no longer required.

PR #1934, opened 2020-10-14
GH-1931: EMF resources created in wrong resource set in LSP builder

mor-n4:
See #1931.


mor-n4:
Builds green, merging ...

Issue #1943, opened 2020-10-20
NullPointerException in case of syntax errors in Semver version strings


mor-n4:
When editing a Semver version in a `package.json` file (e.g. the version constraint of a dependency in the "dependencies" section), `NullPointerException`s are thrown and logged in some cases.

For example, the string `"n4js-runtime": "^0..0"` leads to
```
java.lang.NullPointerException
	at org.eclipse.n4js.semver.SemverMatcher.relation(SemverMatcher.java:376)
	at org.eclipse.n4js.semver.SemverMatcher.matches(SemverMatcher.java:305)
	at org.eclipse.n4js.semver.SemverMatcher.matches(SemverMatcher.java:267)
	at org.eclipse.n4js.semver.SemverMatcher.matches(SemverMatcher.java:207)
	at org.eclipse.n4js.semver.SemverMatcher.matches(SemverMatcher.java:104)
	at org.eclipse.n4js.validation.validators.packagejson.N4JSProjectSetupJsonValidatorExtension.checkVersions(N4JSProjectSetupJsonValidatorExtension.java:1494)
	at org.eclipse.n4js.validation.validators.packagejson.N4JSProjectSetupJsonValidatorExtension.checkReference(N4JSProjectSetupJsonValidatorExtension.java:1411)
	at org.eclipse.n4js.validation.validators.packagejson.N4JSProjectSetupJsonValidatorExtension.checkReferencedProjects(N4JSProjectSetupJsonValidatorExtension.java:1322)
	at org.eclipse.n4js.validation.validators.packagejson.N4JSProjectSetupJsonValidatorExtension.checkDependenciesAndDevDependencies(N4JSProjectSetupJsonValidatorExtension.java:788)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
[...]
```

The string `"n4js-runtime": "^.0.0"` leads to
```
java.lang.NullPointerException
	at org.eclipse.n4js.semver.SemverUtils$VersionNumberDescriptor.getCaretUpperBound(SemverUtils.java:81)
	at org.eclipse.n4js.semver.SemverConverter.simplifyCaret(SemverConverter.java:93)
	at org.eclipse.n4js.semver.SemverConverter.simplify(SemverConverter.java:52)
	at org.eclipse.n4js.semver.SemverConverter.simplify(SemverConverter.java:37)
	at org.eclipse.n4js.semver.SemverMatcher.matches(SemverMatcher.java:206)
[...]

PR #1944, opened 2020-10-21
GH-1943: NullPointerException in case of syntax errors in Semver version strings

mor-n4:
See #1943.


mor-n4:
Builds green, merging ...

Issue #1942, opened 2020-10-20
As a developer I want the LSP server to be resilient to unsupported URIs


mor-n4:
The LSP specification does not mandate that only URIs with a scheme of `file:` may be used. For example, VSCode uses the following special URIs:
* URIs of the form <code>untitled:Untitled-N</code> (with N = 1, 2, ...) are used for newly opened editors that have not yet been saved for the first time.
* compare editors use URIs with a scheme of "git", etc.
* ...

Other LSP clients might introduce other special URIs.

Currently, such URIs make their way into the builder and other parts of the server, including language-specific implementations, causing exceptions and risking invalid internal state and showing bogus errors to the user.

The goal of this user story is to
1. filter out URIs with an unknown/unsupported scheme as early as possible in the LSP server and
2. react in a way that ignores those URIs (i.e., do not show an error to the user, because those URIs are legal).

PR #1945, opened 2020-10-21
GH-1942: As a developer I want the LSP server to be resilient to unsupported URIs

mor-n4:
See #1942.


mor-n4:
Builds green, merging ...

Issue #1946, opened 2020-10-21
As a developer I want number-based enums


mor-n4:
N4JS supports so-called "string-based enums", declared like an ordinary enum but with the annotation `@StringBased`. They do not have any runtime representation and in the output code the literals of such enums are represented as ordinary string literals.

This user story is about adding "number-based enums", that work in much the same way, except that literals are represented as number literals in the output code.

### ACs

1. add support for number-based literals in the type system, compile-time expressions, transpiler, etc.
2. review existing tests for string-based enums (maybe add more, if necessary),
3. add corresponding tests for number-based enums,
4. adjust specification.

PR #1947, opened 2020-10-26
GH-1946: As a developer I want number-based enums

mor-n4:
See #1946.


mor-n4:
Reverted 35090c2 to not blur the diff of this PR with an unrelated fix.

mor-n4:
Builds green, merging ...

PR #1950, opened 2020-10-26
GH-1946: As a developer I want number-based enums

mor-n4:
See #1946.

Fix: literals of number-based enums should also be a subtype of `int`, not just `number`.

(note: so far, `int` in N4JS is just a synonym for `number`)


mor-n4:
Builds green, merging ...

Issue #1952, opened 2020-10-29
As a smith I want the test infrastructure to support integration tests using the LSP protocol (i.e. "IDE tests")


mor-n4:
Currently, all our integration tests are using the headless builder or Eclipse. The goal of this user story is to create an integration test that is using the LSP protocol (based on `AbstractIdeTest`).

The integration tests themselves will reside in a separate repository (because they use proprietary N4JS code as test data) but some adjustments of the IDE test infrastructure are required, which is the topic of this user story.

PR #1953, opened 2020-10-29
GH-1952: As a smith I want the test infrastructure to support integration tests using the LSP protocol (i.e. "IDE tests")

mor-n4:
See #1952.


mor-n4:
Builds green, merging ...

Issue #1951, opened 2020-10-27
Exception in open editors due to stale URIs from other opened editors


mor-n4:
Given three N4JS files A, B, and C, with A having a reference to B and B a reference to C. If file C is deleted while editors are open for files A and B, and the resource task context of A is refreshed before the one of B, then during proxy resolution in A the outdated proxy from B to C will be seen (after B's TModule is loaded from index) and is resolved, leading to a `FileNotFoundException`. As a result, a broken resource for C will exist in the resource set of the resource task context of A (with `isLoaded` flag of `true` but empty content).

This will lead to the following exception upon every future attempt to resolve a proxy to C:
```
ERROR - exception during proxy resolution in super class of N4JSResource:
java.lang.IllegalStateException: cannot install derived state in resource 'file:///Users/mark-oliver.reiser/Home/Prog/n4js/test-onlinepresence/git-repo/onlinepresence/packages/com.enfore.opr.app/src/n4js/deps/n4/ServiceDeploymentEnv.n4jsd' without AST
	at org.eclipse.n4js.resource.N4JSDerivedStateComputer.installDerivedState(N4JSDerivedStateComputer.java:50)
	at org.eclipse.n4js.resource.N4JSResource.installDerivedState(N4JSResource.java:982)
	at org.eclipse.xtext.resource.DerivedStateAwareResource.getContents(DerivedStateAwareResource.java:97)
	at org.eclipse.n4js.resource.N4JSResource.getContents(N4JSResource.java:360)
	at org.eclipse.emf.ecore.resource.impl.ResourceImpl.getEObjectForURIFragmentRootSegment(ResourceImpl.java:766)
	at org.eclipse.xtext.resource.DerivedStateAwareResource.getEObjectForURIFragmentRootSegment(DerivedStateAwareResource.java:212)
	at org.eclipse.n4js.resource.N4JSResource.getEObjectForURIFragmentRootSegment(N4JSResource.java:1198)
	at org.eclipse.emf.ecore.resource.impl.ResourceImpl.getEObject(ResourceImpl.java:808)
	at org.eclipse.emf.ecore.resource.impl.ResourceImpl.getEObject(ResourceImpl.java:787)
	at org.eclipse.xtext.resource.XtextResource.access$101(XtextResource.java:63)
	at org.eclipse.xtext.resource.XtextResource$1.getEObject(XtextResource.java:119)
[...]

PR #1954, opened 2020-10-29
GH-1951: Exception in open editors due to stale URIs from other opened editors

mor-n4:
See #1951.


mor-n4:
Builds green, merging ...

Issue #1956, opened 2020-10-29
Relinking TModule leaving resource in invalid state in case of hash mismatch


mor-n4:
In case an attempt to relink an existing `TModule` to a newly loaded AST fails due to a hash mismatch, the containing resource is left in an invalid state leading to thousands exceptions being thrown and logged if the resource is being used for anything else afterwards.

Example of initial exception:
```
1761 [pool-1-thread-1] ERROR .eclipse.n4js.resource.N4JSResource  - Error in demandLoadResource for file:///Users/mark-oliver.reiser/Home/Prog/Java/n4js-sec/git-repo/n4js/tests/org.eclipse.n4js.ide.tests/test-workspace/ProjectMain/src/A.n4js
java.lang.IllegalStateException: cannot link existing TModule to new AST due to hash mismatch: file:///Users/mark-oliver.reiser/Home/Prog/Java/n4js-sec/git-repo/n4js/tests/org.eclipse.n4js.ide.tests/test-workspace/ProjectMain/src/A.n4js
	at org.eclipse.n4js.typesbuilder.N4JSTypesBuilder.relinkTModuleToSource(N4JSTypesBuilder.java:180)
	at org.eclipse.n4js.resource.N4JSDerivedStateComputer.installDerivedState(N4JSDerivedStateComputer.java:60)
	at org.eclipse.n4js.resource.N4JSResource.installDerivedState(N4JSResource.java:999)
	at org.eclipse.n4js.resource.N4JSResource.demandLoadResource(N4JSResource.java:564)
	at org.eclipse.n4js.resource.N4JSResource$ModuleAwareContentsList.resolve(N4JSResource.java:142)
	at org.eclipse.n4js.resource.N4JSResource$ModuleAwareContentsList.resolve(N4JSResource.java:1)
	at org.eclipse.emf.common.util.BasicEList.get(BasicEList.java:348)
	at org.eclipse.emf.ecore.resource.impl.ResourceImpl.getEObjectForURIFragmentRootSegment(ResourceImpl.java:769)
	at org.eclipse.xtext.resource.DerivedStateAwareResource.getEObjectForURIFragmentRootSegment(DerivedStateAwareResource.java:212)
	at org.eclipse.n4js.resource.N4JSResource.getEObjectForURIFragmentRootSegment(N4JSResource.java:1224)
	at org.eclipse.emf.ecore.resource.impl.ResourceImpl.getEObject(ResourceImpl.java:808)
[...]

PR #1957, opened 2020-11-02
GH-1956: Relinking TModule leaving resource in invalid state in case of hash mismatch

mor-n4:
See #1956.


mor-n4:
Builds green, merging ...

Issue #1959, opened 2020-11-03
As a smith I want to collect and investigate errors logged by the LSP server during production use (no. 2)


mor-n4:
Ongoing effort, continues #1938.

mor-n4:
TODO: find better solution for logging stack traces in LSP case
Search code base for GH-1938.

PR #1961, opened 2020-11-04
GH-1959: As a smith I want to collect and investigate errors logged by the LSP server during production use (no. 2)

mor-n4:
See #1959.


mor-n4:
Builds green, merging ...

Issue #1960, opened 2020-11-03
As a developer I want to use shorthand property names on object literals


halitogunc:
We can omit property name whenever we have a variable which is the same name as a key on an object.

```
let a = 'foo', b = 42, c = {};
let o = {a, b, c};
```

PR #1962, opened 2020-11-04
GH-1960: As a developer I want to use shorthand property names on object literals

mor-n4:
See #1960.


mor-n4:
Builds green, merging ...

Issue #1970, opened 2020-11-18
As a smith I want an experimental dts-to-n4jsd generator supporting simple cases of module structuring


mor-n4:
The goal is to have a first skeleton of the generator that can handle simple cases of `.d.ts` files, in particular with respect to module structuring and references between files.

Not part of this story:
* detailed mapping of types and type references (e.g. just type all function/method parameters as `any+`
* generating an entire n4jsd project
* a prototype that is already usable for development teams

PR #1971, opened 2020-11-19
GH-1970: As a smith I want an experimental dts-to-n4jsd generator supporting simple cases of module structuring

mor-n4:
See #1970.


mmews-n4 (commented):
All in all, I think it is a very good step forward where we can add more functionality later on. Since the acceptance criteria are only along the lines "first skeleton of a generator", I went over the most part without looking too deep into it. However, there are some things that are still noteworthy, I think:
* Please change the project name (as discussed), since "-gen" is usually used as an abbreviation for "generated" which would be misleading here.
* Please use a source folder and an output folder, since this user story is about a skeleton and it could be easily done.
* Please use library methods (of node.js) for dealing with paths, since dealing with paths could be implemented better using a library despite being a prototype.

mmews-n4 (approved):
👍

mor-n4:
Thanks @mmews-n4 !

Merging ...

(no builds because those files are not yet used during the build)

PR #1975, opened 2020-11-23
GH-1974: As a smith I want to refactor the creation of snapshots of workspaces/projects/source folders

### LINKED ISSUE #1974: As a smith I want to refactor the creation of snapshots of workspaces/projects/source folders
Currently, we use `toSnapshot` methods to create snapshots from (workspace/project/source folder) configurations. Since the snapshot classes should remain pure data holders (and should not do much of computation), creation of them is moved to a new factory.

mmews-n4:
closes #1974


plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/server/FileBasedWorkspaceInitializer.java
```diff
@@ -57,6 +58,9 @@
 	@Inject
 	private FileBasedWorkspace workspace;
 
+	@Inject
+	private N4JSConfigSnapshotFactory configSnapshotFactory;
```
mor-n4, line 62: Probably this won't cause problems, but I think it would be better to use `ConfigSnapshotFactory` as type, here.

plugins/org.eclipse.n4js/src/org/eclipse/n4js/xtext/workspace/XWorkspaceConfigSnapshotProvider.java
```diff
@@ -32,12 +30,9 @@
 	 * Note that this implementation may be removed when the N4JS Eclipse product is removed.
 	 */
 	class NullXWorkspaceConfigSnapshotProvider implements XWorkspaceConfigSnapshotProvider {
-		@Inject
-		ProjectBuildOrderInfo.Provider provider;
-
 		@Override
 		public WorkspaceConfigSnapshot getWorkspaceConfigSnapshot() {
-			return new WorkspaceConfigSnapshot(null, Collections.emptyList(), provider);
+			return new ConfigSnapshotFactory().createWorkspaceConfigSnapshot((URI) null);
```
mor-n4, line 35: I think class `NullXWorkspaceConfigSnapshotProvider` can use injection. If that is true, it would be better to use an injected instance of `ConfigSnapshotFactory` here.

plugins/org.eclipse.n4js/src/org/eclipse/n4js/xtext/workspace/ConfigSnapshotFactory.java
```diff
@@ -0,0 +1,173 @@
...
+
+	/** Creates an instance of {@link WorkspaceConfigSnapshot} */
+	public WorkspaceConfigSnapshot createWorkspaceConfigSnapshot(WorkspaceConfigSnapshot snapshot) {
+
+		return createWorkspaceConfigSnapshot(snapshot.path,
+				snapshot.name2Project, snapshot.projectPath2Project, snapshot.sourceFolderPath2Project);
+	}
+
+	/** Creates an instance of {@link WorkspaceConfigSnapshot} */
+	public WorkspaceConfigSnapshot createWorkspaceConfigSnapshot(URI path) {
```
mor-n4, line 73: Some clients are invoking this method with a path argument of `null`. Please add a comment that `path` may be null to API doc of this method and/or other methods in this class.

mor-n4 (approved):
As said yesterday, personally I'd rather keep the lookup maps of `WorkspaceConfigSnapshot` inside that class, because they do not contain additional information but are merely an implementation detail of that data class (for performance reasons). But not critical, no need to change it back right now.

Other than that: very nice refactoring! :+1:

plugins/org.eclipse.n4js/src/org/eclipse/n4js/xtext/workspace/ConfigSnapshotFactory.java
mmews-n4: Actually misleading since a null argument results in invalid instances. The only use case is inside N4JS in a temporary solution until the Eclipse product is removed. See comment at class `NullXWorkspaceConfigSnapshotProvider`.

plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/server/FileBasedWorkspaceInitializer.java
mmews-n4: Same is done with `N4JSScopeProvider` at several locations, e.g. in `N4JSPostProcessor`.

PR #1979, opened 2020-11-25
GH-1978: As a smith I want to remove support for Java-like type annotations

### LINKED ISSUE #1978: As a smith I want to remove support for Java-like type annotations
The support for the old Java-like type annotation syntax is deprecated since the new syntax (using colon and type) was introduced several years ago. As a transition phase, both of these syntaxes were supported to give validations and quickfixes as a courtesy to the user. We define this transition phase to end now, since a simplification of the grammar is helpful with regard to the content assist parser and new features we would like to introduce during the epic to TS Support.

mmews-n4:
closes #1978


mor-n4 (approved):
:+1:

PR #1986, opened 2020-12-02
GH-1985: As a developer I want to use hashbang comments in N4JS (a.k.a. shebang)

### LINKED ISSUE #1985: As a developer I want to use hashbang comments in N4JS (a.k.a. shebang)
To easily implement tools that can be invoked via the command line, it should be possible to add a hashbang comment at the beginning of an N4JS file, specifying the path to a Javascript execution engine similar to a shebang in shell scripts.

For example:
```javascript
#!/usr/bin/env node
console.log("hello world!");
```

A TC39 proposal exists for this:
https://github.com/tc39/proposal-hashbang
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#Hashbang_comments

Probably the parser should store the hashbang text in a (newly added) string-field in class `Script` and the transpiler can then simply prepend that string unchanged to the output file.

mmews-n4:
#closes #1985


plugins/org.eclipse.n4js.transpiler/src/org/eclipse/n4js/transpiler/print/PrettyPrinterSwitch.java
```diff
@@ -1399,6 +1420,10 @@ private String quote(String txt) {
 		return '\'' + ValueConverterUtils.convertToEscapedString(txt, false) + '\'';
 	}
 
+	private String prependHashbang(String txt) {
+		return "#!" + txt;
+	}
```
mmews-n4, line 1425: I wonder why we do not use the ValueConverter (N4JSStringValueConverter for strings, HashbangValueConverter for this, and others) to revert elements to source code.

plugins/org.eclipse.n4js/src/org/eclipse/n4js/N4JS.xtext
```diff
@@ -1077,6 +1082,10 @@ terminal STRING:
 	| "'" SINGLE_STRING_CHAR* "'"?
 ;
 
+terminal HASH_BANG returns ecore::EString:
+	  '#!' (!LINE_TERMINATOR_FRAGMENT)*
+;
```
mor-n4, line 1087: It seems this slipped back in (also in `N4JS.xcore` in class `TypedElement`).

mor-n4 (approved):
Nice! :+1:

plugins/org.eclipse.n4js.transpiler/src/org/eclipse/n4js/transpiler/print/PrettyPrinterSwitch.java
mor-n4: Good point. I'm not sure if there was a reason for this. But at least we are already reusing a utility method from the value converters in `PrettyPrinterSwitch#quote()`.

I also just noticed that there are several `#getValueAsString()` methods defined in `N4JS.xcore` (i.e. a third way of getting back to source code) and they do not seem to be used in all cases by `PrettyPrinterSwitch`.

plugins/org.eclipse.n4js/src/org/eclipse/n4js/N4JS.xtext
```diff
@@ -339,11 +340,15 @@ FormalParameter <Yield>:
...
 	( => bindingPattern=BindingPattern<Yield>
-	| annotations+=Annotation* variadic?='...'? name=BindingIdentifier<Yield> ColonSepDeclaredTypeRef?
+	| annotations+=Annotation* BogusTypeRefFragment? variadic?='...'? name=BindingIdentifier<Yield> ColonSepDeclaredTypeRef?
 	)
 	(hasInitializerAssignment?='=' initializer=AssignmentExpression<In=true, Yield>?)?
 ;
 
+fragment BogusTypeRefFragment*:
+	bogusTypeRef=TypeRefWithModifiers
+;
```
mor-n4, line 350: It seems this slipped back in (also in `N4JS.xcore` in class `TypedElement`).

mmews-n4: Yes, Thanks!

PR #1991, opened 2020-12-04
GH-1968: || (logical or) and ?? (nullish coalescing) operators resolve different return types for the same objects

### LINKED ISSUE #1964: As a developer I want dependency cycles to be reported consistently
**ACs:**
* Re-enable validation for cyclic project dependencies.
* Fix/remove cycle issue propagation inside ProjectBuildOrderInfo
* Adjust tests

### LINKED ISSUE #1968: || (logical or) and ?? (nullish coalescing) operators resolve different return types for the same objects
```typescript
function getSomething(): number[] {
    return [1, 2, 3];
}

function foo () {
    const a = getSomething();
    const b = a || []; // Array<number>
    const c = a ?? []; // union{Array<any>,Array<number>}
}
```
They have a subtle difference which is explained in MDN's website.
> Contrary to the logical OR (||) operator, the left operand is returned if it is a falsy value which is not null or undefined.  

But it shouldn't have any effect on the returned value's type.

mmews-n4:
closes #1968

### DIFF
--- a/plugins/org.eclipse.n4js/src/org/eclipse/n4js/typesystem/TypeJudgment.java
+++ b/plugins/org.eclipse.n4js/src/org/eclipse/n4js/typesystem/TypeJudgment.java
@@ -1187,9 +1187,21 @@ import com.google.inject.Inject;
 		}
 
 		@Override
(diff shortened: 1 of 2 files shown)


mor-n4 (approved):
:+1:

PR #1993, opened 2020-12-04
GH-1992: Negated instanceof guards should strip their checked type from guarded IdentifierRefs

### LINKED ISSUE #1992: Negated instanceof guards should strip their checked type from guarded IdentifierRefs
Mind the following example:

```js
interface I1 {
    f1 = 1;
}
interface I2 {
    f2 = 2;
}

function fn(p: I1|I2) : void {
    if (p instanceof I1) {
        p;
        return;
    }
    // XPECT type of 'p' --> I2
    // XPECT noerrors -->
    p.f2;
}
```

mmews-n4:
closes #1992


mmews-n4:
@mor-n4 Please have a look. In case you have some additional use cases, please let me know and I add some more tests!

plugins/org.eclipse.n4js/src/org/eclipse/n4js/typesystem/TypeJudgment.java
```diff
@@ -607,12 +612,55 @@ public TypeRef caseIdentifierRef(IdentifierRef idref) {
...
+							}
+							if (newUnionTypes.size() < union.getTypeRefs().size()) {
+								remainingType = tsh.createUnionType(G,
+										newUnionTypes.toArray(new TypeRef[newUnionTypes.size()]));
+							}
+						} else {
+
+							boolean removeType = false;
+							for (TypeRef neverHoldingType : excludedTypes) {
+								if (ts.equaltypeSucceeded(G, origType, neverHoldingType)) {
```
mor-n4, line 647: I think `ts.subtypeSucceeded()` might be enough here (and also in line 634 above). This would then support cases like:
```javascript
class A {}
class BSuper {}
class B extends BSuper {}

function foo(p: A|B) {
	if (p instanceof BSuper) {
		return;
	}
	// XPECT type of 'p' --> A
	p;
}
```
But I'm not sure if this is valid in all cases.

mor-n4 (approved):
This is such a cool feature! And hardly any changes required in the flow infrastructure!

plugins/org.eclipse.n4js/src/org/eclipse/n4js/typesystem/TypeJudgment.java
```diff
@@ -607,12 +612,55 @@ public TypeRef caseIdentifierRef(IdentifierRef idref) {
...
+						TypeRef remainingType = null;
+						if (origType instanceof UnionTypeExpression) {
+							final UnionTypeExpression union = (UnionTypeExpression) origType;
+							final Collection<TypeRef> newUnionTypes = new LinkedList<>();
+							for (TypeRef unionType : union.getTypeRefs()) {
+								for (TypeRef neverHoldingType : excludedTypes) {
+									if (!ts.equaltypeSucceeded(G, unionType, neverHoldingType)) {
+										newUnionTypes.add(unionType);
+									}
+								}
```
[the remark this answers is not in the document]
mmews-n4: Only in case there are multiple instanceof type guards that have a right hand side which evaluates to the same type. Note also the simplification logic of the intersection creation.

plugins/org.eclipse.n4js/src/org/eclipse/n4js/typesystem/TypeJudgment.java
mmews-n4: nice!

Issue #1996, opened 2020-12-08
Reflexive reference in default value of object destructuring binding patterns causing havoc


mor-n4:
To reproduce:
```javascript
function foo(config: ~Object with { prop1: boolean, prop2: string[], prop3: number }) {

	const { prop1, prop2, prop3=prop2.length } = config;
}
```

Interestingly, without the unused and seemingly unnecessary declaration of "prop1" in line 3 the bug cannot be reproduced.

Sample exception:
```
753141 [pool-2-thread-7] ERROR       org.eclipse.n4js.utils.UtilN4  - cache miss: no actual type in cache for AST node: org.eclipse.n4js.n4JS.impl.ParameterizedPropertyAccessExpressionImpl@34a7942e (optionalChaining: false) (propertyAsText: length) in resource: file:///Users/mark-oliver.reiser/Home/Prog/n4js/test-n4js/main/packages/ProjectMain/src/Temp3.n4js
java.lang.IllegalStateException: cache miss: no actual type in cache for AST node: org.eclipse.n4js.n4JS.impl.ParameterizedPropertyAccessExpressionImpl@34a7942e (optionalChaining: false) (propertyAsText: length) in resource: file:///Users/mark-oliver.reiser/Home/Prog/n4js/test-n4js/main/packages/ProjectMain/src/Temp3.n4js
	at org.eclipse.n4js.postprocessing.ASTMetaInfoCache.getType(ASTMetaInfoCache.java:112)
	at org.eclipse.n4js.postprocessing.DestructureProcessor.handleForwardReferenceWhileTypingDestructuringPattern(DestructureProcessor.java:125)
	at org.eclipse.n4js.postprocessing.TypeProcessor.getTypeOfForwardReference(TypeProcessor.java:321)
	at org.eclipse.n4js.postprocessing.TypeProcessor.getTypeInN4JSResource(TypeProcessor.java:295)
	at org.eclipse.n4js.postprocessing.TypeProcessor.getType(TypeProcessor.java:264)
	at org.eclipse.n4js.typesystem.N4JSTypeSystem.type(N4JSTypeSystem.java:94)
	at org.eclipse.n4js.utils.DestructureHelper.mergeWithTypeOfDefaultExpression(DestructureHelper.java:307)
	at org.eclipse.n4js.utils.DestructureHelper.addTypeAndContinueWithChildren(DestructureHelper.java:289)
[...]

PR #1997, opened 2020-12-08
GH-1996: Reflexive reference in default value of object destructuring binding patterns causing havoc

mor-n4:
See #1996.


mmews-n4 (approved):
nice bugfix!

mor-n4:

Issue #2023, opened 2021-01-11
Check whether a function type is a subtype of a structural object type yields incorrect result


mor-n4:
To reproduce:
```javascript
let objStr: ~Object with { prop: string };
let fn: (any)=>boolean;

// XPECT FIXME error --> "(any)=>boolean is not a subtype of ~Object with { prop: string; }." at "fn"
objStr = fn;
```

mor-n4:
Before closing this ticket, re-enable the 4 PluginUI integration tests that were temporarily set on ignore!

mor-n4:
The 4 PluginUI tests mentioned above are gone, now. So there's nothing that needs to be re-activated, anymore.

Closing this ticket as DONE.

PR #2024, opened 2021-01-11
GH-2023: Check whether a function type is a subtype of a structural object type yields incorrect result

mor-n4:
See #2023.


mmews-n4 (approved):
Nice fix and clean-ups! 👍

mor-n4:

Issue #2036, opened 2021-01-18
As a smith I need a first proper hash code computation for types and type references


mor-n4:
So far, the computation of hash codes for types and type references in `SemanticEqualsWrapper#hashCode()` is only implemented prototypically (based on `#getTypeRefAsString()`). While the current implementation yields correct results, it has severe performance issues.

For example, due to the performance issues, the recursion guard for the reduction of structural type references could not be implemented as usual and required a work-around (see `Reducer#reduceStructuralTypeRef()`). Probably this would also apply to older recursion guards, if they used the `SemanticEqualsWrapper` (which they should use!).

### ACs

1. implement a proper hash code computation in `SemanticEqualsWrapper#hashCode()` (must not be optimal),
2. ensure that performance is ok,
3. remove work-arounds, esp. in `Reducer#reduceStructuralTypeRef()` (search code base for ``),
4. review all recursion guards and replace direct use of TypeRef/TypeArguments in keys by SemanticEqualsWrapper, if possible without reducing performance.

mor-n4:
This is the work-around used in `Reducer#reduceStructuralTypeRef()` to there avoid the performance issues mentioned above:
8a03660346c0238706b422529df1f6ca1176b956

However, note that `#hashCode()` is still invoked after this change in each of the if-blocks, so it is not 100% clear why this change solved the problem. To be clarified.

PR #2030, opened 2021-01-14
GH-2029: Two bugs caused by recursive structural types

mor-n4:
See #2029.

Issue #2029, opened 2021-01-13
Two bugs caused by recursive structural types


mor-n4:
1) `StackOverflowException` in `Reducer`. To reproduce:
```javascript
class C {}
interface ~I<T> {
	public other(): I<T>
}
function <S> foo(): I<S> {
	return undefined;
}

let i: I<C>;
// i = foo() // <-- comment in for exception!
```
2) Massive performance leak in `StructuralTypingComputer`. To reproduce:
```javascript
class A {}
class B {}
class C {}
class D {}
class E {}
class F {}
interface ~I<T> {
	public m01(): ~I<A>
	public m02(): ~I<B>
	public m03(): ~I<C>
	public m04(): ~I<D>
	public m05(): ~I<E>
	public m06(): ~I<F>
	public m11(): ~I<A>
	public m12(): ~I<B>
	public m13(): ~I<C>
	public m14(): ~I<D>
	public m15(): ~I<E>
	public m16(): ~I<F>
	public m21(): ~I<A>
	public m22(): ~I<B>
	public m23(): ~I<C>
	public m24(): ~I<D>
	public m25(): ~I<E>
	public m26(): ~I<F>
}

let i1: I<string>;
let i2: I<number>;
// i1 = i2; // <-- comment in for loooooong computation!
```

PR #2038, opened 2021-01-18
GH-2029: Two bugs caused by recursive structural types (2nd try)

mor-n4:
See #2029.

Only the commits starting with 8a03660346c0238706b422529df1f6ca1176b956 are new (earlier commits were already part of PR #2030).

This should fix the performance problem introduced with PR #2030. The changes in that PR made use of the hash code computation for type references, for which we've only ever had a prototypical implementation with known performance issues (see newly created user story: `GH-2036`). This is also what caused the new performance issues.

I added a work-around to the fixes introduced with PR #2030. But that's only a temporary solution, the hash code computation must be improved (added corresponding TODO's to `GH-2036`).


mmews-n4 (approved):
👍 
Thanks for fixing the performance issue so fast!
Also, as discussed, you could note in the follow-up ticket to also check for performance.

mor-n4:
Thanks!

I added a comment to `GH-2036`: https://github.com/eclipse/n4js/issues/2036#issuecomment-762248894

Builds green, merging ...

Issue #2042, opened 2021-01-21
Some legal values of yarn's "workspaces" property in package.json not supported


mor-n4:
In the root `package.json` of a yarn workspace, we currently support:
```
"workspaces": [
  "packages/*"
]
```
However, the following is also a legal way of declaring the member projects of a yarn workspace:
```
"workspaces": {
  "packages": ["packages/*"],
  "nohoist": ["**/react-native", "**/react-native/**"]
}
```

This should be supported by N4JS.

See https://classic.yarnpkg.com/blog/2018/02/15/nohoist/

PR #2046, opened 2021-01-22
GH-2042-fix: Some legal values of yarn's "workspaces" property in package.json not supported

mmews-n4:
fix for #2042


mmews-n4:
Builds green, merging

Issue #21, opened 2017-07-07
As a developer I want type aliases


bsmith-n4:
In TypeScript, type aliases can be defined with keyword type, e.g.,

```javascript
type Name = string;
```

This would be in particular useful for union and intersection types, and for function type definitions, e.g.

```javascript
type callback = (int)=>string;
```

This must also be allowed in n4jsd files.

Clarify open questions, such as 
* use in API vs. implementation
* structural typing
* could they be chained (type alias on right hand side)
* generic types
* impact on type equality
* when to resolve the alias, e.g., in error messages and hovers

In general, do some research. Try to specify  feature as compatible to TypeScript as possible (at least similar syntax).

### Acceptance Criteria

```javascript
// XPECT noerrors --> 
type Name = string;

// XPECT errors --> "Type dog is not a recognized type" at "dog"
type Name = dog;
```

* write spec
* write tests
* implement

(moved from internal bug tracker, original reporter: @jpilgrim)

arrowplum:
@kduske-n4 This was not moved over in its original form https://github.com/NumberFour/n4js/issues/319 when n4js was moved over from eclipse. Note is a simplified version, and has no real examples.

Maybe you will have more luck getting anyone to add this than I did.

arrowplum:
See rules here: https://www.typescriptlang.org/docs/handbook/advanced-types.html#type-aliases

They make sense and answer most of the above questions.

- use in API vs. implementation: _must be both_
- structural typing: _should support all types including structural_
- could they be chained (type alias on right hand side) _No. (typescript has some special cases here but I think they don't apply)_
- generic types _should be supported. see typescript_
- impact on type equality _should have no impact (an alias is just a name for a type signature)_
- when to resolve the alias, e.g., in error messages and hovers _should resolve alias_

PR #2084, opened 2021-03-10
GH-21: As a developer I want type aliases

mor-n4:
Continues PR #2049 (which got out of sync with master); discussions/reviews can be found there.

See also #21.


mor-n4:
Builds green, merging ...

Issue #2087, opened 2021-03-17
As a smith I want support for issue suppression on a per-file basis in Xt test files


mor-n4:
In the old Xpect tests we can turn on and off certain errors/warnings in the Xpect setup section of an Xpect test file, using the keyword `IssueConfiguration`.

Support for this feature should be added to the new Xt test framework.

PR #2088, opened 2021-03-17
GH-2087: As a smith I want support for issue suppression on a per-file basis in Xt test files

mor-n4:
See #2087.


mor-n4:
Builds green, merging ...

Issue #2093, opened 2021-03-29
Bogus error shown when exporting a type alias from an N4JSD file


mor-n4:
To reproduce, create a file `Bug.n4jsd` with this content:
```javascript
export external class Cls {}

// XPECT noerrors --> "Only classes, interfaces, enums and functions declared as external as well as structural typed interfaces are allowed in n4jsd files." at "MyAlias"
export external type MyAlias = Cls;
```

PR #2094, opened 2021-03-29
GH-2093: Bogus error shown when exporting a type alias from an N4JSD file

mor-n4:
See #2093.


mor-n4:
Builds green-ish (7 failures, fixed above), merging ...

Issue #2101, opened 2021-04-07
type alias leaves export statement in transpiled output


dbo:
```n4js
export public type Foo = string;
export public const bar = 5;
```
=>
```js
export 
;
export const bar = 5;
```

PR #2102, opened 2021-04-07
GH-2101: type alias leaves export statement in transpiled output

mor-n4:
See #2101.


mor-n4:
Builds green, merging ...

Issue #2103, opened 2021-04-08
As a smith I want to update node.js to version 14


mor-n4:
Bump our build to the latest 14.x version of node.
Update yarn accordingly.

PR #2104, opened 2021-04-08
GH-2103: As a smith I want to update node.js to version 14

mor-n4:
See #2103.


mor-n4:
Builds green, merging ...

Issue #2108, opened 2021-04-12
Semicolon after a type alias declaration producing bogus empty statement in AST


mor-n4:
When the declaration of a type alias is followed by a semicolon, then the AST will contain an unnecessary empty statement. While this should not do any harm, it does not look nice when showing up in the output code (after the type alias declaration was removed).

PR #2110, opened 2021-04-13
GH-2108: Semicolon after a type alias declaration producing bogus empty statement in AST

mor-n4:
See #2108.


mor-n4:
Builds green, merging ...

Issue #2111, opened 2021-04-13
As a developer I'd like to use type aliases as values (esp. in JSX tags)


mor-n4:
When implementing the first iteration of type aliases, it was a deliberate decision to disallow the use of type aliases as values, because they are types only and do not have a representation at runtime.

For example, all the following is currently disallowed by a dedicated validation:
```javascript
function foo(p: constructor{?}) {}

class Cls {}

type A = Cls;

A; // error: A type alias may not be used as a value.
new A(); // error: A type alias may not be used as a value.
foo(A); // error: A type alias may not be used as a value.
```

However, it would be possible to allow a type alias being used as a value iff it resolves to a type that is also a value (in the above cases, for example). In that case, the transpiler would have to rewrite the reference to the type alias to a reference to the aliased / actual type (in the example above, rewrite references to `A` to refs to `Cls`).

Such a use of a type alias as value might be particularly useful in a JSX tag.

At the moment, this ticket is mainly intended to keep track of this topic. It is not yet clear if such a feature is worth the effort/complexity, so this has to be clarified before starting with this user story.

WARNING: supporting this is not trivial, because
* the transpiler would have to insert additional imports (if the aliased type is declared in a different module than the type alias),
* some invalid cases have to be considered (e.g. what if only the type alias is exported but not the aliased type; what if the aliased type is declared in a different project and no direct project dependency exists to that project?).

mor-n4:
IMPORTANT: the above, merged PR does **not** resolve this ticket!
It only included some preparations, see PR description for details.

PR #2112, opened 2021-04-13
GH-2111: As a developer I'd like to use type aliases as values (esp. in JSX tags) -- preparations

mor-n4:
See #2111.

This PR only includes some minor improvements of the handling of `IdentifierRef`s that point to type aliases. Since this is disallowed by a validation anyway, it does not have much impact, but improves consistency a bit and avoids some duplicate error messages. Also, might prove helpful in the future.

The actual topic of #2111 is not addressed by this PR.


mor-n4:
Builds green, merging ...

Issue #2116, opened 2021-04-13
JSX validation showing bogus error when React.Element is changed from class to interface


mor-n4:
When react's type `Element` is changed from a class to an interface, a JSX-related validation is showing a bogus error.

To reproduce, set up a small sample N4JSX project with a dependency to react. Now, in the definition file of the project `@n4jsd/react`, replace the declaration of abstract class Element with:
```javascript
export external public interface ~Element<P extends ComponentProps> {
    public type: string;
    public props: P;
    public key: string;
}
```
Then, the following code will show a bogus error message:
```javascript
import * as React from "react";

function Foo(): React.Element<?> {
	return null;
}

<Foo/>; // error "Expect a function returning React.Element but the return type is Element<? extends ComponentProps>." at "Foo"
```

mor-n4:
As far as I can tell, the actual problem was not the change of `Element` from abstract class to interface but that it was made structural _and_ generic. This wasn't supported by the validation, because it used a short-cut for constructing the type reference used for the subtype checking.

mor-n4:
A similar problem in the JSX support showed up (also after changing `Element` from class to structural, generic interface):
```javascript
import * as React from "react";

function Foo(): React.Element<?> {
	return null;
}

let b: boolean;
let value = b ? [<Foo/>] : null;

// XPECT noerrors --> "Array<Element> is not a subtype of Array<Element<? extends ComponentProps>> due to incompatible type arguments: Element<? extends ComponentProps> is not a structural subtype of Element: props failed: ComponentProps is not a subtype of P." at "value"
let elem: Array<React.Element<?>> = value;
```

PR #2117, opened 2021-04-13
GH-2116: JSX validation showing bogus error when React.Element is changed from class to interface

mor-n4:
See #2116.


mor-n4:
Builds green, merging ...

Issue #1637, opened 2020-01-30
As a developer I want a dynamic variant of default imports and explicit default imports


mmews-n4:
**Dynamic variant of default import** 
```ts
// XPECT noerrors --> "Parser error" at "+"
import lib+ from "some-lib";
```

**Explicit default import** 
```ts
// XPECT noerrors --> "Couldn't resolve reference to TExportableElement" at "default"
import {default as lib} from "some-lib";
```

PR #2127, opened 2021-04-22
GH-1637: As a developer I want a dynamic variant of default imports and explicit default imports

mor-n4:
See #1637.


mor-n4:
Builds green, merging ...

Issue #2128, opened 2021-04-22
As a smith I want a performance test suite for content assist of n4js/d/x files


mor-n4:
To tackle the issue of slow content assist we need to create a test suite to assess the (reproducible) current state of the time performance when invoking content assist. As a result we will be able to (1) monitor and hence recognise improvements/degrations in the time performance, and (2) speed-up our implementation based on that test suite and not only on some arbitrarily test examples.

PR #2129, opened 2021-04-22
GH-2128: As a smith I want a performance test suite for content assist of n4js/d/x files

mor-n4:
See #2128.

The actual integration test added in this user story is located in a separate repository because it uses a non-public N4JS code base as test data.


mor-n4:
Builds green, merging ...

Issue #2130, opened 2021-04-28
Editor opened in invalid state after following hyperlinks


mor-n4:
When following a hyperlink by moving the mouse pointer over an identifier and _immediately_ CMD-clicking, then the newly opened editor (containing the declaration the identifier refers to) is in an invalid state.

PR #2131, opened 2021-04-29
GH-2130: Editor opened in invalid state after following hyperlinks

mor-n4:
See #2130.


mor-n4:
Builds green, merging ...

Issue #2126, opened 2021-04-21
Value of this-keyword in functions is assignable to everything


mor-n4:
```javascript
class SomeClass {}

function foo() {
	// XPECT FIXME errors --> "..."
	let x: SomeClass = this;
	// XPECT FIXME errors --> "..."
	bar(this);
}

function bar(p: SomeClass) {}
```

mor-n4:
This is caused by the fact that `this` in a function without `@This` annotation is inferred to `undefined`, which is of course a subtype of everything, hence the assignments above are allowed.

While it is true that in strict mode 'this' in a function evaluates to true, when the function is invoked without setting the this target, we cannot ensure that this is the case (because the this target may be set and then the type of `this` is no longer `undefined`). Therefore, the type must be `any`.

PR #2132, opened 2021-04-29
GH-2126: Value of this-keyword in functions is assignable to everything

mor-n4:
See #2126.


mor-n4:
Builds green, merging ...

Issue #2140, opened 2021-05-25
As a smith I want to resolve two performance problems of content assist


mor-n4:
#### Problem 1

In large workspaces, the use of a flat `ResourceDescriptionsData` instead of a `ChunkedResourceDescriptions` leads to performance problems in content assist (`scope.getAllElements()` requires 600-800 ms). However, `ChunkedResourceDescriptions` seems to require the use of a `ProjectDescription` on the ResourceSet and thus the use of one resource set per project.

The goal of this user story is to tune Xtext to allow the use of `ChunkedResourceDescriptions` together with a single resource set, using an instance of `IAllContainersState` instead of `ProjectDescription`.

#### Problem 2

Secondly, in `ReferenceResolutionFinder` method `scope.getElements(shortName)` is invoked for each element in the iterable returned by `scope.getAllElements()`, which also leads to performance problems. This should also be fixed by this user story.

PR #2141, opened 2021-05-25
GH-2140: As a smith I want to resolve two performance problems of content assist

mor-n4:
See #2140.

The two main changes are:

https://github.com/eclipse/n4js/blob/7643dcfda0f224dec59da87b7cf109a3451f2c91/plugins/org.eclipse.n4js.xtext/src/org/eclipse/n4js/xtext/resource/containers/XStateBasedContainerManager.java#L43-L46
(allowing the use of `ChunkedResourceDescriptions` together with `IAllContainersState` instead of `ProjectDescription`)

... and:

https://github.com/eclipse/n4js/blob/7643dcfda0f224dec59da87b7cf109a3451f2c91/plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/imports/ReferenceResolutionFinder.java#L464-L466
(see problem 2 from task description)


mmews-n4 (approved):
I looked through all changed files. Actually, I don't know whether all of the changes are due to performance reasons or not. I guess that some also change the results of e.g. content assist. If so I wonder that no tests are modified to make sure the changes won't be broken by accident in the future.

mor-n4:
@mmews-n4 Aside from the main performance tweaks (see description of this PR), this only contains a few minor tweaks and simplifications and some purely cosmetic changes for readability (e.g. avoiding use of extension methods in `N4JSScopeProvider`). The actual result of content assist should not be changed (verified by temporary modifications to our `ContentAssistPerformanceTestInSuite`).

Most changes to content assist that I did introduce accidentally were discovered by existing CA tests. There was one exception, though, and I will try to add a test or two for that case.

mor-n4:
Thanks @mmews-n4!

Builds green, merging ...

Issue #2147, opened 2021-06-08
As a developer I want open workspace symbol to be responsive


mor-n4:
At the moment, open symbol functionality for the entire workspace (CMD+T in VS Code) takes extremely long to respond. The goal of this user story is to improve performance such that the response time is at least acceptable (need not be extremely fast, but should be usable).

PR #2148, opened 2021-06-08
GH-2147: As a developer I want open workspace symbol to be responsive

mor-n4:
See #2147.


mor-n4:
NOTE: performance could be improved further by sending symbols to the client in serveral steps, but that would require a newer version of LSP, which would require an update of lsp4j + Xtext.

See
https://microsoft.github.io/language-server-protocol/specifications//specification-3-17/#progress
https://microsoft.github.io/language-server-protocol/specifications//specification-3-17/#partialResults

Once this is done, the better filtering logic can & should be activated, here:
https://github.com/eclipse/n4js/blob/5b9654de939883764ec6ec5d59f1a612618a00ee/plugins/org.eclipse.n4js.ide/src/org/eclipse/n4js/ide/server/symbol/N4JSDocumentSymbolService.java#L55-L61

mor-n4:
PENDING: log a message if a partial result token is received from the client!

mmews-n4 (approved):
👍 Very nice!

mor-n4:
Builds green, merging ...

PR #2172, opened 2021-07-27
GH-2171: Small bugs in .d.ts export

### LINKED ISSUE #2171: Small bugs in .d.ts export
The N4JS code
```javascript
class C {
	public m() {
		return 42 !== 0;
	}
}
```
is producing this .d.ts output:
```typescript
declare class C {
	public m(): void;
}
```
but the return type should be `any`.

mmews-n4:
closes #2171

### DIFF
--- a/plugins/org.eclipse.n4js.transpiler.dts/src/org/eclipse/n4js/transpiler/dts/transform/ReturnTypeTransformation.xtend
+++ b/plugins/org.eclipse.n4js.transpiler.dts/src/org/eclipse/n4js/transpiler/dts/transform/ReturnTypeTransformation.xtend
@@ -10,9 +10,10 @@
  */
 package org.eclipse.n4js.transpiler.dts.transform
 
+import com.google.inject.Inject
 import org.eclipse.n4js.n4JS.FunctionDefinition
 import org.eclipse.n4js.transpiler.Transformation
-import org.eclipse.n4js.ts.types.TFunction
+import org.eclipse.n4js.transpiler.assistants.TypeAssistant
 
 import static org.eclipse.n4js.transpiler.TranspilerBuilderBlocks.*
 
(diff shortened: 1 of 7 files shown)


plugins/org.eclipse.n4js.transpiler.dts/src/org/eclipse/n4js/transpiler/dts/transform/ReturnTypeTransformation.xtend
```diff
@@ -42,7 +40,10 @@ class ReturnTypeTransformation extends Transformation {
 			// return type already provided explicitly, so nothing to do here
 			return;
 		}
-		val typeRefNode = _TypeReferenceNode(state, state.G.voidTypeRef);
+		
+		val elemInAST = state.tracer.getOriginalASTNode(funDef) as FunctionDefinition;
+		val tFunction = elemInAST.definedType as TFunction;
+		val typeRefNode = _TypeReferenceNode(state, tFunction.returnTypeRef);
```
mor-n4, line 46: If `funDef` was created by an earlier transformation, then `elemInAST` will be `null` and there won't be a TFunction for `funDef`. This means functions created by transformation must always provide the return type explicitly (which makes sense, because they might not have a body).

To remind us when running into that problem, could you throw an exception in this case? Best align the above three lines to
https://github.com/eclipse/n4js/blob/7109634cc31893844f140c03c19594833412dffb/plugins/org.eclipse.n4js.transpiler.dts/src/org/eclipse/n4js/transpiler/dts/transform/TypeReferenceTransformation.xtend#L129-L138

(or factor out the code from `TypeReferenceTransformation` into a utility method in `TypeAssistant`)

mor-n4 (approved):
Thanks!!

mmews-n4:
Thanks!

PR #2176, opened 2021-08-03
GH-2175: Regression due to GH-2164: compiler not stopping after printing out version with given option --version

### LINKED ISSUE #2175: Regression due to GH-2164: compiler not stopping after printing out version with given option --version
`n4jsc --version` is supposed to print out the version and then exiting successfully. Instead, it is continuing with the implicit compile goal.

mmews-n4:
closes #2175


mor-n4 (approved):
:+1:

That was quick! 😄  THanks!

mmews-n4:
Builds green, merging

PR #2183, opened 2021-08-04
GH-2182: Raw type leaking through in static methods of generic classes that return 'this' types

### LINKED ISSUE #2182: Raw type leaking through in static methods of generic classes that return 'this' types
```js
class C<T> {
  static m() : this { return null; }
}

const t = C.m();
// FIXME XPECT type of t --> C<?>
t;
```

mmews-n4:
closes #2182


mor-n4 (approved):
Very good fixes!

Issue #2200, opened 2021-08-23
Constraint solver confused by certain special cases of union types


mor-n4:
To reproduce:
```typescript
class B {}
class C extends B {}

class G1<T> {}
class G2<T> {}

function <TF> foo(p1: TF, p2: G1<TF> | G2<TF>): Array<TF> {
	return null;
}
let arg: G1<B> | G2<B>;
let result = foo(new C(), arg);
// XPECT FIXME type of 'result' --> Array<B>
result;
```

mor-n4:
For reference: as part of this ticket I fixed a second problem related to the use of wildcards in type aliases:
```typescript
class C {}

type Alias<T> = (param: T)=>void;

let ac: Alias<C>;
let aw: Alias<?>;


// simple subtype check
// XPECT FIXME noerrors --> "Alias<B> is not a subtype of Alias<?>." at "ab"
aw = ac;


function <TF> foo(p: Alias<TF>): Alias<TF> { return p; }
// inference with constraints
// XPECT FIXME noerrors --> "Alias<B> is not a subtype of Alias<TF>." at "ab"
aw = foo(ac);
```

mor-n4:
For reference: as part of this ticket I fixed a second problem related to the use of wildcards in type aliases:
```typescript
class C {}

type Alias<T> = (param: T)=>void;

let ac: Alias<C>;
let aw: Alias<?>;


// simple subtype check
// XPECT FIXME noerrors --> "Alias<B> is not a subtype of Alias<?>." at "ab"
aw = ac;


function <TF> foo(p: Alias<TF>): Alias<TF> { return p; }
// inference with constraints
// XPECT FIXME noerrors --> "Alias<B> is not a subtype of Alias<TF>." at "ab"
aw = foo(ac);
```

PR #2201, opened 2021-08-24
GH-2200: Constraint solver confused by certain special cases of union types

mor-n4:
See #2200.

NOTE: this actually includes fixes for two separate problems, see issue for details.


mor-n4:
Builds are green, merging ...

Issue #2197, opened 2021-08-19
As a developer I want to use literal types


mor-n4:
As a developer I want to use boolean, numeric, and string literal types in N4JS.

PR #2203, opened 2021-09-02
GH-2197: As a developer I want to use literal types

mor-n4:
See #2197.

Note: in addition to what was mentioned in the task description, I also added support for enum literal types.


mor-n4:
Pending:
- [ ] add more tests for standard use cases of literal types of all kinds
- [ ] check if validations for literals are still working
- [ ] reconsider approach of method `InferenceContext#isPromisingPartialSolution()`

tests/org.eclipse.n4js.xpect.tests/xt-tests/typesystemPT/variableFieldTypeInference/Importee.n4js
```diff
@@ -18,8 +18,8 @@ export public class Remote {
 	public fieldFN_decl : {function(string):number} = function(p : string) : number { return 0; };
 	public                           fieldFN_inf  = function(p : string) : number { return 0; };
 	
-	public fieldOL_decl : ~Object with { s: string; n: number; } = { s: "hello", n: 42.0 };
-	public                                      fieldOL_inf  = { s: "hello", n: 42.0 };
+	public fieldOL_decl : ~Object with { s: string; n: number; } = { s: "hello", n: 42.3 };
+	public                                      fieldOL_inf  = { s: "hello", n: 42.3 };
```
mmews-n4, line 22: I don't remember but did you need to change these digits because otherwise this literal would be interpreted as a whole number? If so, at least in Java, there are some situations where it is handy to define floats adding `.0`, right?

mmews-n4 (approved):
Wow, really great feature!

tests/org.eclipse.n4js.xpect.tests/xt-tests/typesystemPT/variableFieldTypeInference/Importee.n4js
mor-n4: Good point. At some point I thought it's ok to infer `1.0` to "int" instead of "number" and therefore changed these tests. A few days later I thought better of it (now `1.0` is inferred to "number") but forgot to revert the changes in this file (and 2 others).

Done!

mor-n4:
Thanks @mmews-n4!

Builds green, merging ...

PR #2206, opened 2021-09-09
GH-1314: As a smith I want to fully align scoping to npm/yarn

### LINKED ISSUE #1314: As a smith I want to fully align scoping to npm/yarn
BACKGROUND: _Most of the alignment of N4JS to npm/yarn has been completed in 2018 / early 2019 (`package.json` instead of `Manifest.MF`, redesign of library manager, simplification of runners, etc.). However, one final step remains in order to achieve full alignment:_

Currently, the N4JS IDE has the notion of a single set of external dependencies for the entire Eclipse workspace. This is in conflict with node/npm semantics in the following cases:

1. Given there are two projects P1 and P2 where each of them has a `node_modules` folder NM1 and NM2. Further given that both of the `node_modules` folders an npm is installed that has the same name, e.g. 'lodash'. At the moment, the type index of N4JS can contain only one package of the same name at once. Hence, 'lodash' of NM2 will shadow 'lodash' of NM1 (shadowing is resolved alphabetically). According to npm/node, however, each project can have its own dependencies without being affected by the other project's dependencies in any way.
2. The "side-by-side use case" (i.e. when having two projects P1, P2 in the Eclipse workspace with dependencies between each other) is no longer supported, but some remnants of the old behavior are still in the scoping and some other areas. However, P1 should be entirely invisible to P2, and vice versa.

In this user story, scoping and shadowing in the library manager (and other areas of the implementation such as validation, if need be) should be adjusted to ensure everything is handled in N4JS exactly as in node/npm.

The above applies similarly to the headless builder, so that should be adjusted accordingly.

### ACs

1. adjust scoping, library manager, etc. to
   1. support separate `node_modules` folders of unrelated projects,
   2. entirely disable the "side-by-side use case", i.e. disallow dependencies from one project to another when *not* having a yarn workspace setup.
2. ensure all validation (and their error messages) still make sense + consider adding new validations (if required)
3. ensure proper tool support is available for all related use cases (e.g. adjust wizards or quick fixes, if necessary).
4. remove all remaining aspects of the support for the side-by-side case from the N4JS implementation and the tests.
5. adjust and/or write new tests.
6. update design document, specification. etc.

mmews-n4:
closes #1314


plugins/org.eclipse.n4js/src/org/eclipse/n4js/packagejson/projectDescription/ProjectDescription.java
```diff
@@ -33,7 +34,11 @@
 @SuppressWarnings("javadoc")
 public class ProjectDescription extends ImmutableDataClass {
 
-	private final String name;
+	private final FileURI location;
+	private final FileURI relatedRootlocation;
+	private final String id;
+
+	private final String packageName;
```
mor-n4, line 41: Good idea to use "packageName" for this!

plugins/org.eclipse.n4js/src/org/eclipse/n4js/packagejson/projectDescription/ProjectDescriptionBuilder.java
```diff
@@ -54,14 +58,33 @@ public ProjectDescriptionBuilder() {
...
-				workspaces);
+		id = id == null ? computeQualifiedName() : id;
+		return new ProjectDescription(location, relatedRootLocation, id,
+				packageName, vendorId, vendorName, version, type, mainModule, extendedRuntimeEnvironment,
+				providedRuntimeLibraries, requiredRuntimeLibraries, dependencies, implementationId, implementedProjects,
+				outputPath, sourceContainers, moduleFilters, testedProjects, definesPackage, nestedNodeModulesFolder,
+				n4jsNature, yarnWorkspaceRoot, isGeneratorEnabledDts, workspaces);
+	}
+
+	public String computeQualifiedName() {
```
mor-n4, line 69: I guess you forgot to rename this to `computeId()`?

plugins/org.eclipse.n4js/src/org/eclipse/n4js/utils/NodeModulesDiscoveryHelper.java
```diff
@@ -69,10 +73,7 @@ public NodeModulesFolder(boolean isYarnWorkspaceRoot, boolean isYarnWorkspaceMem
 		 * @return {@link #localNodeModulesFolder} and {@link #workspaceNodeModulesFolder} iff not null.
 		 */
 		public List<File> getNodeModulesFoldersInOrderOfPriority() {
-			// GH-1314: according to node's module look-up semantics it would be correct to give the local folder
-			// priority over the workspace folder; however, because N4JS does not yet support several projects with the
-			// same name, we have to give the workspace folder the higher priority
```
mor-n4, line 74: Nice to see this comment go! :smile:

plugins/org.eclipse.n4js/src/org/eclipse/n4js/workspace/N4JSProjectConfigSnapshot.java
```diff
@@ -34,25 +35,28 @@
...
-			Iterable<? extends SourceFolderSnapshot> sourceFolders) {
+			Iterable<? extends SourceFolderSnapshot> sourceFolders, Map<String, String> packageNameToProjectIds) {
 
-		super(projectDescription.getName(), path,
+		super(projectDescription.getId(), path,
 				Collections.singleton(URIUtils.trimTrailingPathSeparator(path).appendSegment(N4JSGlobals.PACKAGE_JSON)),
 				indexOnly, generatorEnabled, dependencies, sourceFolders);
 
 		this.projectDescription = Objects.requireNonNull(projectDescription);
+		this.packageNameToProjectIds = packageNameToProjectIds;
```
mor-n4, line 59: To guarantee immutability of the snapshot and for consistency with the super class, we should create a defensive copy here. I.e.:
```java
this.packageNameToProjectIds  = ImmutableMap.copyOf(packageNameToProjectIds);
```
(also the type of field `packageNameToProjectIds` should be ImmutableMap<..> (compare with super class))

plugins/org.eclipse.n4js/src/org/eclipse/n4js/workspace/N4JSWorkspaceConfig.java
```diff
@@ -247,27 +258,43 @@ private WorkspaceChanges detectAddedRemovedProjects(WorkspaceConfigSnapshot oldW
...
+						changes = changes.merge(N4JSProjectConfig.computeChanges(oldPC, newPC));
+					}
+				} else {
+					// names changed
 					ProjectConfigSnapshot newPC = configSnapshotFactory
 							.createProjectConfigSnapshot(newProjectsMap.get(uri));
-
-					changes = changes.merge(N4JSProjectConfig.computeChanges(oldPC, newPC));
+					changes = changes.merge(WorkspaceChanges.createProjectAdded(newPC));
+					changes = changes.merge(WorkspaceChanges.createProjectRemoved(oldProjectsMap.get(uri)));
```
mor-n4, line 288: Ah! So behind this point, the old rule that project (now: package) names never change and instead projects are removed/added still holds? Great!

plugins/org.eclipse.n4js/src/org/eclipse/n4js/packagejson/projectDescription/ProjectDescription.java
```diff
@@ -279,7 +309,8 @@ public boolean isGeneratorEnabledDts() {
 	@Override
 	protected int computeHashCode() {
 		return Objects.hash(
-				name,
+				id,
```
mor-n4, line 312: Would be helpful to have a comment above this line, saying that field "location" is deliberately not included (same in method `#computeEquals()`).

mor-n4 (approved):
Very nice to have this long-standing issue solved, and with a very elegant approach!
Also great how you managed to avoid confusion with the terminology of "name" vs. "projectName" vs. ..., as far as possible.

Thanks!

plugins/org.eclipse.n4js/src/org/eclipse/n4js/utils/NodeModulesDiscoveryHelper.java
mmews-n4: True! 🥳

plugins/org.eclipse.n4js/src/org/eclipse/n4js/workspace/N4JSWorkspaceConfig.java
mmews-n4: Yes, true. The reason behind this is that when the package names changes, the dependency relations also might change and need to be recomputed.

plugins/org.eclipse.n4js/src/org/eclipse/n4js/packagejson/projectDescription/ProjectDescription.java
mmews-n4: Thanks for pointing that out. I added the missing fields to the `hashCode` and `equals` methods. Leaving them would be ok somehow but there is no specific reason to do so. Adding them would be at least consistent with the current state, although it could be argued that it is semantically not necessary.

mmews-n4:
Thank you! Builds are green. Merging.

PR #2218, opened 2021-09-30
GH-2212: As a smith I want to refactor scoping of identifier refs within the current file

### LINKED ISSUE #2212: As a smith I want to refactor scoping of identifier refs within the current file

mmews-n4:
closes #2212


plugins/org.eclipse.n4js/src/org/eclipse/n4js/scoping/utils/SourceElementExtensions.java
```diff
@@ -0,0 +1,318 @@
...
+		TreeIterator<EObject> allContents = element.eAllContents();
+		InternalCFEMapper veeSwitch = new InternalCFEMapper(start, includeBlockScopedElements, allContents, result);
+		while (allContents.hasNext()) {
+			EObject next = allContents.next();
+			veeSwitch.doSwitch(next);
+		}
+		return result;
+	}
+
+	private class InternalCFEMapper extends N4JSSwitch<Boolean> {
```
mor-n4, line 144: I guess CFE means "control flow element"? If so, the class name should be updated.

mor-n4, line 144: Do we need the return value of the case-methods? If not, we could use `java.lang.Void` as type argument for the super class.

plugins/org.eclipse.n4js/src/org/eclipse/n4js/scoping/utils/SourceElementExtensions.java
```diff
@@ -0,0 +1,318 @@
...
+			this.addHere = addHere;
+		}
+
+		@Override
+		public Boolean caseN4ClassDeclaration(N4ClassDeclaration feature) {
+			Type polyfilledOrOriginalType = getTypeOrPolyfilledType(feature);
+			N4ClassDeclaration nonNullClassDecl = feature;
+			if (polyfilledOrOriginalType instanceof TClass) {
+				TClass polyfilledOrOriginalTypeCasted = (TClass) polyfilledOrOriginalType;
+				N4ClassDeclaration n4ClassDecl = (N4ClassDeclaration) polyfilledOrOriginalTypeCasted.getAstElement();
```
mor-n4, line 165: Since going from types model to AST is problematic (most of the time), we should avoid `#getAstElement()` here, if possible. If I haven't overlooked anything, the AST node `n4ClassDecl` is only used in the following to invoke `#getDefinedType()` on it, which will return `polyfilledOrOriginalTypeCasted`. So going via the AST does not seem necessary.

tests/pom.xml
```diff
@@ -357,9 +357,10 @@ Contributors:
 				<module>org.eclipse.n4js.ide.tests</module>
 				<module>org.eclipse.n4js.ide.tests.helper.tests</module>
 				<module>org.eclipse.n4js.ui.tests</module> <!-- note the bad name: these are actually IDE tests -->
-
+<!-- 
 				<module>org.eclipse.n4js.n4idl.lang.tests</module>
 				<module>org.eclipse.n4js.n4idl.spec.tests</module>
+ -->
```
mor-n4, line 363: Don't forget to undo this! 😄

plugins/org.eclipse.n4js/src/org/eclipse/n4js/scoping/validation/IScopeValidator.java
```diff
@@ -0,0 +1,26 @@
...
+ *   NumberFour AG - Initial API and implementation
+ */
+package org.eclipse.n4js.scoping.validation;
+
+import org.eclipse.xtext.resource.IEObjectDescription;
+
+/**
+ * Interface for validators of scope elements.
+ */
+public interface IScopeValidator {
```
mor-n4, line 18: Nice design!

mor-n4 (approved):
Nice improvement of readability!

plugins/org.eclipse.n4js/src/org/eclipse/n4js/scoping/utils/SourceElementExtensions.java
mmews-n4: I like that `Void` return type but returning non-null values will cause the N4JSSwitch to stop checking parent classes. That's why I used Boolean here. Let's discuss this later.

plugins/org.eclipse.n4js/src/org/eclipse/n4js/scoping/utils/SourceElementExtensions.java
mmews-n4: I'll try to get rid of it. But actually this code is only a merge of the xtend file with the same name.

tests/pom.xml
mmews-n4: Or get rid of n4idl 😄

plugins/org.eclipse.n4js/src/org/eclipse/n4js/scoping/validation/IScopeValidator.java
mmews-n4: Thanks!

mmews-n4:
Thanks for the review! Builds green. Merging.

PR #2214, opened 2021-09-27
GH-2213: As a smith I want to improve the internal handling of built-in types in the N4JS compiler

mor-n4:
See #2213.

Issue #2213, opened 2021-09-27
As a smith I want to improve the internal handling of built-in types in the N4JS compiler


mor-n4:
Currently we have an internal language "n4ts" for representing built-in types. We would like to use "n4jsd" files instead, as far as possible. Ideally, we would like to get rid of language "n4ts" entirely (maybe in a later step).

PR #2220, opened 2021-09-30
GH-2213: As a smith I want to improve the internal handling of built-in types in the N4JS compiler

mor-n4:
See #2213.

This was already reviewed in PR #2214. However, we do not want to touch that PR at the moment, because it can be used to reproduce a GitHub bug that was reported to GitHub. Therefore, this new PR is used to merge the changes from GH-2213.


mor-n4:
Builds green, merging ...

PR #2225, opened 2021-10-07
GH-2223: As a smith I want to remove N4IDL

### LINKED ISSUE #2223: As a smith I want to remove N4IDL
Unfortunately, the feature N4IDL that supports versions of types and migrations between different versions, is used and there is no hint that it will be used anytime soon. However, this feature complicates especially the grammar and scoping of the N4JS language. With the current epic of supporting more idioms of Typescript in order to import/export d.ts files, we need to work on the N4JS grammar and scoping a lot. Hence, to simplify these tasks, we need to remove N4IDL.

mmews-n4:
closes #2223


mor-n4 (approved):
Thanks!

Issue #2224, opened 2021-10-06
As a developer I need support for polyfilling interfaces in runtime libraries


mor-n4:
Polyfills in runtime libraries are only supported for classes, not interfaces. The goal of this user story is to also support this for interfaces.

NOTE: this user story in only about ordinary polyfills (defined with annotation `@Polyfill`), not about static polyfills (i.e. `@StaticPolyfill`)!

PR #2226, opened 2021-10-07
GH-2224: As a developer I need support for polyfilling interfaces in runtime libraries

mor-n4:
See #2224.


mor-n4:
Thanks!

Build is green, merging ...

Issue #2228, opened 2021-10-13
(Non-static) polyfills throwing exception in case of several polyfills for same type per project


mor-n4:
When a single runtime library project contains several (non-static) polyfills _**for the same filled type**_, then exceptions are thrown in some cases.

(Note that this scenario always involves several source files, because several polyfills for the same filled type are not possible within a single source file.)

To reproduce:
* create a runtime library project.
* add two source files:
  ```
  // someModuleA.n4jsd
  @@Global @@ProvidedByRuntime
  
  @Polyfill export external public class Array<T> extends Array<T> {
  }
  ```
  and
  ```
  // someModuleB.n4jsd
  @@Global @@ProvidedByRuntime

  @Polyfill export external public class Array<T> extends Array<T> {
  	public [Symbol.hasInstance](): any+;
  }
  ```
* close everything and make a clean full build. This may or may not throw the below exception (depending on build order of the two modules).
* open `someModuleA.n4jsd` in VS Code and the below exception should appear.

Exception:
```
Caused by: java.lang.IllegalArgumentException: name may not be null
	at org.eclipse.n4js.ts.types.util.NameStaticPair.<init>(NameStaticPair.java:48)
	at org.eclipse.n4js.ts.types.util.NameStaticPair.<init>(NameStaticPair.java:40)
	at org.eclipse.n4js.ts.types.util.NameStaticPair.of(NameStaticPair.java:33)
	at org.eclipse.n4js.validation.utils.MemberCube.addMembers(MemberCube.java:52)
	at org.eclipse.n4js.validation.utils.MemberCube.<init>(MemberCube.java:43)
	at org.eclipse.n4js.validation.validators.N4JSMemberRedefinitionValidator.createMemberValidationList(N4JSMemberRedefinitionValidator.java:1278)
	at org.eclipse.n4js.validation.validators.N4JSMemberRedefinitionValidator.checkMemberRedefinitions(N4JSMemberRedefinitionValidator.java:189)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
[...]

PR #2229, opened 2021-10-13
GH-2228: (Non-static) polyfills throwing exception in case of several polyfills for same type per project

mor-n4:
See #2228.


mmews-n4 (approved):
Thanks! 👍

mor-n4:
Thanks!
Builds are green, merging ...

PR #2231, opened 2021-10-14
GH-2230: Incorrect scoping in modules containing (non-static) polyfills

### LINKED ISSUE #2230: Incorrect scoping in modules containing (non-static) polyfills
In modules containing a non-static polyfill, references to the _filled_ type outside the filling type are incorrectly resolved to the filling type instead of the filled type.

To reproduce:
* create a runtime library project.
* add a source file:
  ```
  // someModule.n4jsd
  @@Global @@ProvidedByRuntime
  
  @Polyfill export external public class Array<T> extends Array<T> {
  	addedMethod(): Array<string>; // correctly scoped to Array in builtin_js.n4jsd
  }
  
  export external public class SomeClass {
  	someMethod(): Array<string>; // incorrectly scoped to Array in this module
  }
  ```
* `Array` in the return type declaration of `someMethod` is incorrectly scoped (see comment above).

This leads to the problem that client code using method `SomeClass#someMethod()` cannot use the ordinary members from Array (such as `length`).

mmews-n4:
closes #2230


mor-n4 (approved):
Thanks!

PR #2247, opened 2021-11-03
GH-2246: Bump versions of dependencies in n4js-libs due to vulnerabilities

mor-n4:
See #2246.


mmews-n4 (approved):
Usually we make these changes explicitly, if possible in the dependency sections or otherwise in the resolution sections of some package.json files. Note that e.g. `tar` has already a resolution for version `4.4.15`. It might be confusing to rely on the `yarn.lock` file regarding this fix at a later point in time.

mor-n4:
Builds green, merging ...

PR #2249, opened 2021-11-03
GH-2245: As a developer I want type guards to support disjunctions via logical or expressions resulting in union types

### LINKED ISSUE #2245: As a developer I want type guards to support disjunctions via logical or expressions resulting in union types
Support this:
```js
class C1 {}
class C2 {}
class C3 {}
class C4 {}

let a: C1 | C2 | C3 | C4;
if (a instanceof C1 || a instanceof C2) {
    // XPECT FIXME type of 'a' --> union{C1,C2}
    a;
}
```

mmews-n4:
closes #2245


mor-n4 (approved):
Very nice!
This opens up cool 😎 opportunities for supporting more things in the future.

Just one remark: given the following code
```javascript
function foo(p: 1|2|3|4): 0|3|4 {
	if (p === 1 || p === 2) {
		return 0;
	}
	return p;
}
```
method `GuardStructure#getGuardList()` is invoked 4 times for expression `p === 1` and 4 times for expression `p === 2`. Seems that once each should be sufficient.

mmews-n4:
Thanks!

PR #2256, opened 2021-11-08
GH-2255: As a developer I want the default type of an optional type arguments to be optional

### LINKED ISSUE #2255: As a developer I want the default type of an optional type arguments to be optional
At the moment a default type has to be defined even if the optional type argument can be `any`. See:
```js
class C<T=any> {}
```

The goal of this user story is to make the type `any` optional, as in:
```js
class C<T= > {}
```

mmews-n4:
closes #2255


plugins/org.eclipse.n4js/src/org/eclipse/n4js/parser/LazyTokenStream.java
```diff
@@ -187,15 +189,123 @@ public void fillBuffer(int lookAhead) {
...
+	}
+
+	boolean forbidHiddenTokens = false;
+
+	@Override
+	protected void doSetHiddenTokens(String... lexerRules) {
+		super.doSetHiddenTokens(lexerRules);
+		if (lexerRules == null || lexerRules.length == 0) {
+			Preconditions.checkArgument(lexerRules == null || lexerRules.length == 0,
+					"Setting the hidden channel dynamically is only allowed for empty list of arguments");
```
mor-n4, line 219: Something seems amiss, here: given the condition of the if-statement, this `#checkArgument()` call never fails.

plugins/org.eclipse.n4js.model/model/N4JS.xcore
```diff
@@ -1990,17 +1990,18 @@ class N4TypeVariable extends IdentifiableElement, NamedElement {
 
 	boolean declaredCovariant
 	boolean declaredContravariant
+	boolean declaredOptional
 
 	contains TypeReferenceNode<TypeRef> declaredUpperBoundNode
 
-	/* The default type argument or {@code null} iff this type parameter is non-optional. */
+	/* The default type argument or {@code null}. */
 	contains TypeReferenceNode<TypeRef> defaultArgumentNode
```
mor-n4, line 1998: I'd propose to rename this property to `declaredDefaultArgumentNode`. This would make it easier to remember the subtle difference between this property and `TypeVariable#defaultArgument` in `Types.xcore`: this property can be `null` and should only be interpreted together with `declaredOptional`/`#isOptional()`, whereas the property in `Types.xcore` cannot be `null` and is "self-contained", etc.

mor-n4 (approved):
:+1: Better approach than what I had in mind, at first!

PR #2263, opened 2021-11-11
GH-2257: As a developer I want namespaces in N4JSD files

### LINKED ISSUE #2257: As a developer I want namespaces in N4JSD files

mmews-n4:
closes #2257


plugins/org.eclipse.n4js/src/org/eclipse/n4js/typesbuilder/N4JSClassDeclarationTypesBuilder.xtend
```diff
@@ -94,7 +93,7 @@ public class N4JSClassDeclarationTypesBuilder extends N4JSClassifierDeclarationT
 
 		n4Class.definedType = tclass;
 
-		target.internalTypes += tclass;
+		target.containingModule.internalTypes += tclass;
```
mor-n4, line 96: I find this confusing. It seems `target` is sometimes the target (i.e. the `AbstractNamespace` that will contain the new class) and sometimes the sibling of the new class.

plugins/org.eclipse.n4js/src/org/eclipse/n4js/utils/N4JSLanguageUtils.xtend
```diff
@@ -844,6 +845,11 @@ public class N4JSLanguageUtils {
 			&& value.compareTo(N4JSGlobals.INT32_MAX_VALUE_BD) <= 0;
 	}
 
+	/** Checks presence of {@link AnnotationDefinition#POLYFILL} annotation. See also {@link N4JSLanguageUtils#isStaticPolyfill(AnnotableElement) }*/
+	def static boolean isNamespaceElement(AnnotableElement astElement) {
+		return AnnotationDefinition.POLYFILL.hasAnnotation( astElement );
```
mor-n4, line 850: Name of this method does not seem to match its behavior. (probably a copy/paste error?)

mor-n4 (approved):
This resolves some long-standing limitations in representing plain-JS libraries. Great!

NOTE: since it is still in progress, some of my comments probably refer to things you were going to change before merging, anyway.

plugins/org.eclipse.n4js/src/org/eclipse/n4js/typesbuilder/N4JSClassDeclarationTypesBuilder.xtend
mmews-n4: What would be an example of `target` being a sibling of a new class here?

plugins/org.eclipse.n4js.ts.model/model/TypeRefs.xcore
```diff
@@ -571,7 +571,19 @@ class ParameterizedTypeRef extends BaseTypeRef {
 	 * <li>the declared type is an enum literal, then this is non-null and points to the TEnum containing the literal.
 	 * </ol>
 	 */
-	refers transient Type astDeclaredTypeQualifier
+	contains NamespaceLikeRef[] namespaceLikeRefs
```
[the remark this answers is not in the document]
mmews-n4: What is the meaning of `if ... is used as an AST node` ?

plugins/org.eclipse.n4js/src/org/eclipse/n4js/typesbuilder/N4JSClassDeclarationTypesBuilder.xtend
mor-n4: Sorry, I should've explained better what I meant: `target.containingModule` sometimes returns `target` itself, and sometimes the parent of `target`. Depending on these two cases, the role of an argument passed in for parameter `target` changes: in the former case the newly created `tClass` will be added to the argument passed in, in the latter case it will be added to the parent of the argument passed in, which seems confusing IMHO. Why not always adding the new `tClass` to `target`, i.e. always passing in the element that will be the container of the newly created `tClass`?

plugins/org.eclipse.n4js.ts.model/model/TypeRefs.xcore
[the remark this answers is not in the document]
mor-n4: "TypeRef is used as an AST node" means "TypeRef was created by the parser", i.e. "TypeRef represents an AST node".

plugins/org.eclipse.n4js/src/org/eclipse/n4js/typesbuilder/N4JSClassDeclarationTypesBuilder.xtend
mmews-n4: What you describe is true for three instances (Class, Function, ObjectLiteral). What you propose means either introduce another parameter or change the target parameter. Of course your proposal still can be done, but this would break the systematically structured calls in `N4JSTypesBuilder` in the bodies of the _dispatch method_ named `createType`. In case you prefer also to pass the new argument through right from `buildTypes` it should be noted that also all methods of `relinkType` were affected since they currently are implemented similarly.

mor-n4: Ok, sorry. My misunderstanding. I confused `#internalTypes` with `#topLevelTypes`.

mmews-n4:
Thanks for your review! The build is green... I'll do the merge.

Issue #2262, opened 2021-11-11
As a smith I want to change when/how tests of n4js-libs are executed and restructure test bundles


mor-n4:
The goal of this user story is to change when and how the tests of the n4js-libs are executed in the N4JS builds and also slightly restructure our test bundles.

In particular:
1. testing of n4js-libs should be moved from script `publish-n4js-libs.sh` to a proper test located in a test bundle.
2. merge bundle `org.eclipse.n4js.cli.tests` into `org.eclipse.n4js.ide.tests.jar`.
3. rename `org.eclipse.n4js.ide.tests.jar` to `org.eclipse.n4js.tests.integration`.
4. rename `com.enfore.n4js.ide.tests.integration` to `com.enfore.n4js.tests.integration`.
5. move test classes inheriting from `AbstractCliTest` from `org.eclipse.n4js.ide.tests` to `org.eclipse.n4js.tests.integration`.
6. ...

PR #2264, opened 2021-11-11
GH-2262: As a smith I want to change when/how tests of n4js-libs are executed and restructure test bundles

mor-n4:
See #2262.


mor-n4:
Note to self: before merging, compare number of tests on this branch with that on `master`!

mmews-n4 (approved):
really nice clean-ups!

mor-n4:
Thanks @mmews-n4!

Builds are green and the exact same number of tests are executed on this branch as on master (57150), which is expected because I have deleted 1 test and added 1 test.

Merging ...

PR #2270, opened 2021-11-24
GH-2269: Scoping for a type and value of same name is incorrect if they are global / provided by runtime

### LINKED ISSUE #2269: Scoping for a type and value of same name is incorrect if they are global / provided by runtime
If a type and a value have the same name, a reference to that name is incorrectly bound in some cases if the type and value are defined as `@@Global` / `@@ProvidedByRuntime`.

To reproduce:
1. add this file to a runtime library project:
   ```javascript
   @@Global @@ProvidedByRuntime
   
   export external public interface ~URL {
       toJSON(): string;
   }
   
   export public var URL: ~Object with {
       new(p: boolean): URL;
       (p: string): number;
   };
   ```
2. add this file to an ordinary project with a dependency to the runtime library from step 1:
   ```javascript
   // XPECT noerrors -->
   URL; // works (binds to variable)
   
   // XPECT noerrors -->
   URL(""); // works (binds to variable)
   
   // XPECT FIXME noerrors --> "Cannot instantiate interface URL." at "URL"
   new URL(true); // incorrect (binds to interface)
   ```

Note: when removing the line `@@Global @@ProvidedByRuntime` and adding an appropriate import to the second file, everything seems to be working as expected, even though `URL` is still defined in a runtime library.

mmews-n4:
closes #2269


mor-n4 (approved):
That was quick! Nice! :+1:

mmews-n4:
Thank you!

PR #2274, opened 2021-11-25
GH-2272: Wrong binding in rare situations when an identifier can bind to both value and type

### LINKED ISSUE #2272: Wrong binding in rare situations when an identifier can bind to both value and type
In case a module exports both a value and a type that have the same name, and that module lies in another project, it can happen that at the importing location the binding is incorrectly determined.

mmews-n4:
closes #2272


mor-n4 (approved):
Nice! :+1:

Issue #2277, opened 2021-12-02
As a developer I want type definitions for the DOM API imported from standard documents


mor-n4:
Similarly to #2233, this user story is about importing the `.n4jsd` files for the DOM API from TypeScript `.d.ts` files. Those files are currently located in `n4js-runtime-html5` and this user story won't change that.

PR #2278, opened 2021-12-02
GH-2277: As a developer I want type definitions for the DOM API imported from standard documents

mor-n4:
See #2277.


mmews-n4 (approved):
👍 Wow, despite we will keep on working and improving on that topic, I think this is a big step forward for N4JS!

mor-n4:
Thanks @mmews-n4!

Builds are green now, merging ...

Issue #2292, opened 2022-01-06
As a smith I need to find a proper way of installing n4js-libs in tests with "npm install"


mor-n4:
If a test is using `npm install` to install the dependencies of a test project that include one or more n4js-libs packages, then the latest published version of n4js-libs will be installed, not the version of the current build / current branch.

To avoid this problem (and to speed up tests), we do not install n4js-libs in tests using "npm/yarn install" but instead use the utility method `N4jsLibsAccess#installN4jsLibs()`.

However, some (few) tests do actually require the use of "npm/yarn install", because they test npm/yarn behavior that cannot easily re-implemented in `N4jsLibsAccess#installN4jsLibs()`.

The goal of this user story is to find an approach for implementing those tests without running into the above problem.

ACs:
1. find an approach.
2. review all existing tests that use "npm/yarn install" and either
   * adjust them to use this approach, or
   * delete them, if they are not actually required.

mor-n4:
Note: also during this issue I made our tests compatible with nvm.

PR #2297, opened 2022-01-17
GH-2292: As a smith I need to find a proper way of installing n4js-libs in tests with "npm install" (bonus)

mor-n4:
Adds a more helpful error message in case tests are run locally and developer forgot to start verdaccio first.
(without this, npm/yarn would hang extremely long before failing with a timeout)


mmews-n4 (approved):
👍

Issue #2291, opened 2022-01-05
As a developer I want to use node's native ES module support and get rid of esm


mor-n4:
Use of node's native ES module support can already be used in N4JS projects on the developer side, by simply adding node's `"type":"module"` to the package.json and removing `-r esm` from the command line when invoking node.

However, we want to get rid of the dependency from N4JS to library `esm` altogether, in particular removing the dependency to `esm` from `n4js-runtime` and removing it from all our node invocations during tests in the N4JS build.

Also:
* ensure interoperability with npm packages not using native esm (e.g. CJS),
* the n4js-libs must be updated to use `"type":"module"`,
* the `n4jsc init` command has to be adjusted, etc.,
* migrate existing N4JS code bases (or ensure backward compatibility),
* ...

Details to be discussed with @dbo

PR #2313, opened 2022-02-17
GH-2291: As a developer I want to use node's native ES module support and get rid of esm (part 2)

mor-n4:
See #2291.

This is part 2 of PR #2307, adding the following:
1. tests,
2. spec update,
3. minor improvements in test infrastructure (e.g. `N4jsLibsAccess` now uses JSON-library instead of the Xtext-based JSON implementation and thus no longer depends on an Xtext setup / injector).


mmews-n4 (approved):
👍  Nice!

mor-n4:
(builds were green)

PR #2316, opened 2022-02-17
GH-2314: As a smith I want the new N4JS compiler to not use more memory than the old one

### LINKED ISSUE #2314: As a smith I want the new N4JS compiler to not use more memory than the old one

mmews-n4:
see #2314


mor-n4 (approved):
Good you found something! :+1:

Issue #2318, opened 2022-02-23
As a developer I want support for module declarations in d.ts (first iteration)


mor-n4:
Module declarations should be supported in .d.ts files. This will require the concept of "declared modules" to be introduced internally in the N4JS language implementation, but they should not be made available in N4JS / N4JSD files.

Also legacy modules (i.e. module declarations with an identifier instead of a string literal as name) should be supported.

This user story is only about a first iteration, including:
* module declarations in .d.ts files should be parsed and processed correctly,
* those modules should be available to import statements in N4JS files,
* touching a .d.ts file should update N4JS files importing from a module declared in that .d.ts file.

Not covered by this first iteration:
* import statements inside declared modules in .d.ts files,
* build order issues (i.e. ensuring a .d.ts file with a declared module M is built before an N4JS file importing from M).

PR #2319, opened 2022-02-24
GH-2318: As a developer I want support for module declarations in d.ts (first iteration)

mor-n4:
See #2318.


plugins/org.eclipse.n4js.model/model/N4JS.xcore
```diff
@@ -2158,25 +2160,47 @@ abstract class N4ClassDefinition extends N4ClassifierDefinition, ThisTarget {
...
+abstract class N4AbstractNamespaceDeclaration extends VariableEnvironmentElement, NamedElement {
 
-class N4NamespaceDeclaration extends N4TypeDefinition, VariableEnvironmentElement, ModifiableElement, ExportableElement, NamedElement {
 	String name
 	contains NamespaceElement[] ownedElementsRaw
-	
+}
+
+/**
+ * This type of declaration is created only from .d.ts files and represents a module declaration such as
```
mmews-n4, line 2170: Did you consider to interpret a module declaration as a new "virtual" resource in the sense that a new resource is created while parsing that contains the ast of that module declaration? I could imagine that this would be possible and an N4JS module declaration would be obsolete. The only thing I wonder is how the URI of this virutual resource would look like.

plugins/org.eclipse.n4js.ts.model/model/Types.xcore
```diff
@@ -198,25 +219,52 @@ class TModule extends AbstractNamespace, SyntaxRelatedTElement, TAnnotableElemen
...
+ * </pre>
+ * Note that module declarations of the form
+ * <pre>
+ * // .d.ts
+ * declare module myLegacyModule { ... }
+ * </pre>
+ * are actually an old syntax for namespaces and will thus be represented as a {@link TNamespace},
+ * not as a {@code TDeclaredModule}.
+ */
+class TDeclaredModule extends AbstractModule {
```
mmews-n4, line 252: Adding to my comment above: I think it is a drawback that the grammar of N4JS is modified due to a pure d.ts feature. My assumption is, that his feature is not part of our roadmap for N4JS - what do you think? In other cases like the namespaces, it seems more likely that we actually want that as a N4JS feature.

plugins/org.eclipse.n4js/src/org/eclipse/n4js/resource/N4JSLinker.java
```diff
@@ -410,6 +411,10 @@ protected boolean hasLeafNodes(INode node) {
 	 */
 	@Override
 	protected EClass findInstantiableCompatible(EClass eType) {
+		// FIXME this is bad; #createProxy() above should be able to create proxies for abstract classes
```
mmews-n4, line 414: mind the fixme

mmews-n4 (commented):
Please mind my thoughts below. You'll notice that I am a little worried that this implementation is more intrusive than an alternative that would introduce "virtual" resources, meaning resources that have virutual uris, something like `[original URI]/moduleDeclaration`. I would like to hear your thoughts about that!

mmews-n4 (approved):

plugins/org.eclipse.n4js.model/model/N4JS.xcore
mor-n4: Yes, actually I started with something in that direction (not an entirely separate resource, but using "ordinary" `TModule`s instead of introducing `TDeclaredModule`). But at some point I stopped following this approach because I felt that it leads to too many special cases / complexity in other areas (and in areas that are more tricky to understand). Also, my gut feeling is that since we have to deal with declared modules anyway in later stages, it is better to have them represented explicitly as a concept.

However, the approach of a "virtual" resource is definitely a good idea, but it is difficult to tell what leads to the simplest solution without investigating this further. Therefore, we should definitely keep it in mind and reconsider it.

As discussed/decided in meeting: let's merge as is for now and come back to it later.

plugins/org.eclipse.n4js.dts/src/org/eclipse/n4js/dts/astbuilders/DtsNamespaceBuilder.java
```diff
@@ -69,14 +107,84 @@ public DtsNamespaceBuilder(DtsTokenStream tokenStream, LazyLinkingResource resou
...
+	 */
+	public N4NamespaceDeclaration consume(NamespaceDeclarationContext ctx) {
+		return (N4NamespaceDeclaration) super.consume(ctx);
+	}
+
+	/**
+	 * Overloads {@link #consume(ParserRuleContext)} to codify the rule that when a {@link ModuleDeclarationContext} is
+	 * passed in, you might get back an {@link N4NamespaceDeclaration} or an {@link N4ModuleDeclaration}.
+	 */
+	public N4AbstractNamespaceDeclaration consume(ModuleDeclarationContext ctx) {
```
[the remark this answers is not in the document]
mor-n4: I had separate sub-classes and merged them, because it isn't possible to get rid of the checks, anyway. However, I just double-checked this and found a solution that allows to get rid of at least some of the instanceof/checks.

plugins/org.eclipse.n4js.ts.model/model/Types.xcore
mor-n4: Yes, declared modules will most likely not be made available in N4JS[D].
(other than that, see response above)

plugins/org.eclipse.n4js/src/org/eclipse/n4js/resource/N4JSLinker.java
mor-n4: Good find! It is obsolete, will remove it.

mor-n4:
Thanks for the valuable feedback @mmews!

Builds green, merging ...

PR #2332, opened 2022-04-11
GH-2330: Bump versions of dependencies in n4js-libs due to vulnerabilities

mor-n4:
See #2330.


mmews-n4 (approved):
Thanks for bumping! :-)

Issue #2347, opened 2022-05-04
Functions with optional return value handled incorrectly by .d.ts export


mor-n4:
The N4JS code
```javascript
let fn: ()=>string?;
```
is exported to .d.ts as
```typescript
declare let fn: ()=>string;
```
but should produce
```typescript
declare let fn: ()=>string|void;
```

Thanks for reporting this @dbo!

PR #2348, opened 2022-05-04
GH-2347: Functions with optional return value handled incorrectly by .d.ts export

mor-n4:
See #2347.


mmews-n4 (approved):
👍  That was quick!

mor-n4:
Thanks @mmews-n4!

Builds are green, merging ...

Issue #2352, opened 2022-05-09
As a developer I want support for declaration merging in .d.ts


mor-n4:
Our .d.ts import should support declaration merging.

PR #2354, opened 2022-05-11
GH-2352: As a developer I want support for declaration merging in .d.ts (preparations, part 2)

mor-n4:
See #2352.

* more clean up in `N4JSQualifiedNameProvider`,
* removal of qualified name computation for AST nodes (was dead code),
* elements contained in namespaces are also added to global Xtext index (see `N4JSResourceDescriptionStrategy`),
* unique qualified names for all elements in the global Xtext index (hopefully),
* in particular: added new special segment to separate the module specifier from the element name(s) within a qualified name.


mmews-n4 (approved):
👍  Nice improvements!

PR #2372, opened 2022-06-09
GH-2266: Some special handling for @NumberBased enums missing

### LINKED ISSUE #2266: Some special handling for @NumberBased enums missing
```javascript
@NumberBased
enum E {
    Lit: 42
}

// XPECT FIXME noerrors --> "100 is not a subtype of E." at "100"
(E.Lit < 100);
// XPECT FIXME type of 'x' --> number
let x = E.Lit + 5; // x has type string

// TODO: look for more cases as the two above (probably there are more)

// working as intended:
let n: number = E.Lit;

// working as intended, but could be allowed since we now have numeric literal types available:
let e: E = 42; // error: "42 is not a subtype of E." at "42"
```

mmews-n4:
closes #2266


mmews-n4:
@mor-n4 Thanks!

Issue #2373, opened 2022-06-13
As a developer I want to replace n4js-runtime-node by @types/node


mor-n4:
Replace `n4js-runtime-node` by `@types/node` in ...
* n4js-libs:
  * entirely delete package `n4js-runtime-node` (also from publishing script, etc.)
  * adjust code of other n4js-libs packages
* Java source code of N4JS compiler and tooling
* VSCode extension
* stdlib
* OPR
* ...

PR #2374, opened 2022-06-13
GH-2373: As a developer I want to replace n4js-runtime-node by @types/node

mor-n4:
See #2373.


plugins/org.eclipse.n4js.model/src/org/eclipse/n4js/AnnotationDefinition.java
```diff
@@ -258,6 +259,29 @@ public final class AnnotationDefinition {
...
+			.targets(SCRIPT)
+			.args(STRING_LITERAL)
+			.transitive()
+			.retention(RetentionPolicy.TYPE)
+			.end();
+
+	/**
+	 * States that the annotated {@link TClassifier classifier} owns an index signature on the .d.ts side.
+	 * <p>
+	 * Only to be used in .d.ts.
```
mmews-n4, line 277: Do we need a validation for that constraint? I.e. unsupported feature validation?

mmews-n4 (approved):
👍  Really great improvements!

plugins/org.eclipse.n4js.model/src/org/eclipse/n4js/AnnotationDefinition.java
mor-n4: Good idea, will do!

mor-n4:
Thanks for the review @mmews-n4!

Build is green, merging ...

Issue #2381, opened 2022-06-15
Transpiler emits bogus output code for property access on element re-exported from .d.ts


mor-n4:
To reproduce:
```javascript
// def1.d.ts
export class Cls {}
```
and
```javascript
// def2.d.ts
export { Cls as Cls2 } from "./def1"
```
and
```javascript
// main.n4js
import * as d from "def2"
new d.Cls2();
```
will produce the output code:
```javascript
import * as d from './def2.js'
new d.Cls();  // <--- wrong! should be d.Cls2
```

mor-n4:
Ticket remains OPEN, for now.

Pending: double-check if maybe the old behavior can be removed from the transpiler entirely (instead of keeping it as fall-back behavior).

PR #2382, opened 2022-06-16
GH-2381: Transpiler emits bogus output code for property access on element re-exported from .d.ts

mor-n4:
See #2381.


mmews-n4 (approved):

mor-n4:
Thanks @mmews-n4!

Builds green, merging ...

PR #2385, opened 2022-06-20
GH-2347: Functions with optional return value handled incorrectly by .d.ts export

mor-n4:
The initial solution of PR #2348 leads to inconveniences in certain cases. This is an experiment to improve that situation.


mmews-n4 (approved):

mor-n4:
Thanks!

Build is green, merging ...

PR #2392, opened 2022-06-22
GH-2380: As a developer I want to use d.ts definitions shipped with immutable npm instead of @n4jsd/immutable (merge 2)

mmews-n4:
This contains the second part of all changes, i.e. bug fixes that occur when using d.ts computed property names. Since a remaining problem related to scoping remains open and will take some time (due to complexity of the scoping implementation), this PR is factored out to get on master early.


mor-n4 (approved):
:+1: Important improvements!

Issue #2394, opened 2022-06-22
As a developer I want support for combining "export =" in .d.ts with a named import (in both N4JS and .d.ts)


mor-n4:
Given a .d.ts type definition such as
```typescript
// lib.d.ts
declare interface Ifc {
	field: number;
	method(p: string): number;
}
declare const c: Ifc;
export = c;
```
support named imports such as:
```javascript
import {field} from "lib";
import {method} from "lib";
```

PR #2397, opened 2022-06-27
GH-2394: As a developer I want support for combining "export =" in .d.ts with a named import (in both N4JS and .d.ts)

mor-n4:
See #2394.


mmews-n4 (approved):
👍 Nice to have that feature!

mor-n4:
Thanks @mmews-n4!!

Builds green, merging ...

PR #2408, opened 2022-07-08
GH-2403: As a developer I want to improve handling of dependencies of plain-js projects

### LINKED ISSUE #2403: As a developer I want to improve handling of dependencies of plain-js projects
since moment-timezone makes use of re-exports of moment and also of declaration merging with elements from moment, we need to adjust the current why of (a) handling dependencies of Plain-Js projects and (b) declaration merging between separate projects. This userstory will work on (a) only.
Currently, dependencies of plain-js projects will be ignored, with the exception that all n4js-libs projects will be built first. The reason for ignoring dependencies is performance, the reason for handling n4js-libs projects first are d.ts files inside plain-js projects that may need type information from n4js-libs projects.
However, due to re-exports in d.ts files in plain-js projects we also need to improve the handling of dependencies.

mmews-n4:
closes #2403


mmews-n4:
Builds are as green as can be.

PR #2417, opened 2022-08-11
GH-2414: As a smith I want to investigate why the initial build of an already-built project takes >1 min

### LINKED ISSUE #2414: As a smith I want to investigate why the initial build of an already-built project takes >1 min

mmews-n4:
closes #2414


mmews-n4:
build is green, merging.

PR #2418, opened 2022-08-15
GH-1328: As a developer I want to use import.meta

### LINKED ISSUE #1328: As a developer I want to use import.meta
The import.meta object is a meta-property exposing context-specific meta data to a JavaScript module. It contains information about the module, like the module's URL.

A new built-in type will be required for the value the expression `import.meta` evaluates to. This type should be minimal as it can be polyfilles via runtime libraries.

See:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import.meta
https://github.com/tc39/proposal-import-meta

### ACs

1. adjust parser,
2. add new built-in type, add required polyfills to runtime libraries in n4js-libs,
3. adjust type inference,
4. add required validations,
5. add tests,
6. update specification.

mmews-n4:
closes #1328


mmews-n4:
build is green, merging

PR #2425, opened 2022-08-25
GH-2423: As a developer I want support/handling for declaration merging with existing global type definitions

### LINKED ISSUE #2423: As a developer I want support/handling for declaration merging with existing global type definitions
At the moment, elements in global modules defined in d.ts files collide with those from runtime libraries, i.e. merging does not work here. This userstory is about to either support merging of d.ts and n4jsd elements, or to mitigate that issue otherwise.

mmews-n4:
closes #2423


mmews-n4:
builds green, merging

PR #2433, opened 2022-09-08
GH-2432: Missing import in output js code when importing from polyfills with re-exports

### LINKED ISSUE #2432: Missing import in output js code when importing from polyfills with re-exports
Importing elements from a module that were previously re-exported from another module results in generated output code that lacks those imports. This is due to some merge logic that deals with polyfills but does not consider that re-exported elements point to other modules than those that are defined in import statements.

mmews-n4:
closes #2432


mmews-n4:
build green, merging

PR #2446, opened 2022-11-03
GH-2445: Support package.json property 'types' and consequently module separation of implementation and type definitions

### LINKED ISSUE #2445: Support package.json property 'types' and consequently module separation of implementation and type definitions
At the moment, N4JS assumes that modules share the same name in case their implementation and type definitions are separated into two modules that differ only in their file extensions (e.g. index.js and index.d.ts). However, package.json files do not enforce this convention and some npm packages come with a file for type definitions that has a completely different name than that for the corresponding implementation (e.g. `react-saga` uses `./index.d.ts` and `./dist/redux-saga-core-npm-proxy.esm.js`).

mmews-n4:
closes #2445


mmews-n4:
build is green, merging.

PR #2447, opened 2022-11-04
GH-2445-B: Support package.json property 'types' and consequently module separation of implementation and type definitions

mmews-n4:
second part of #2445


mmews-n4:
build is green, merging

PR #2452, opened 2022-11-28
GH-2438 B: As a developer I want to use d.ts react definitions (instead of n4jsd react definitions)

mmews-n4:


mmews-n4:
build is green (except for 1 error due to a generated file from another repo). merging.

PR #2463, opened 2022-12-19
GH-2462: Polyprocessor fails to confirm subtype relation in case of properties typed as union types that contain type ~Object with {}

### LINKED ISSUE #2462: Polyprocessor fails to confirm subtype relation in case of properties typed as union types that contain type ~Object with {}
Actually, this is an addendum to #2459.

**TEST**
```ts
export interface ~A {
	public sn?: string | number;
}
export interface ~B {
	public sb?: string | boolean;
}

let utf9 : string|number = "utf9";
// XPECT noerrors -->
let y1 : A & ~Object with {} = {sn : utf9};
// XPECT noerrors -->
let y3 : A & ~Object with {} = {sn : "utf9"};
// XPECT noerrors -->
let y4 : A & ~Object with { so?: string } = {sn : "utf8"};
```

mmews-n4:
closes #2462


mmews-n4:
build is green, merging

PR #2467, opened 2022-12-21
GH-2466: Cannot access members of unions when any+ is one of the composed types

### LINKED ISSUE #2466: Cannot access members of unions when any+ is one of the composed types
**TEST**
```ts
export class C {
	static sMethod(): void { }
}

export var V : C  | any+;

// XPECT errors --> "The static member sMethod cannot be accessed from a non-static context." at "sMethod"
V.sMethod();
```

mmews-n4:
closes #2466

The error message turned out to be unclear.


mmews-n4:
build is green, merging

PR #2509, opened 2023-04-27
GH-2508: Fix compiler exception due to d.ts term Promise<void>

### LINKED ISSUE #2508: Fix compiler exception due to d.ts term Promise<void>
Reproduce with the following two files:

**tslib.d.ts**
```ts
export function cp(source: string): Promise<void>;
```

**Module.n4js**
```js
import {cp} from "DTS";

async function F() : void {
    // XPECT type of 'cp' --> {function(string):Promise<undefined>}
    await cp(null);
}
```

mmews-n4:
closes #2508


mmews-n4:
build is green, merging.

PR #2526, opened 2023-07-14
GH-2525: VSCode extension fails to start due to wrong static value for variable import.meta.url

### LINKED ISSUE #2525: VSCode extension fails to start due to wrong static value for variable import.meta.url
Webpack bundles the extension and replaces `import.meta.url` by a static name since `import.meta.url` is not available in commonjs. This fails at runtime.

mmews-n4:
closes #2525


mmews-n4:
build is green, merging.

PR #2531, opened 2023-07-26
GH-2530: Default project type must be 'library' iff project nature is N4JS

### LINKED ISSUE #2530: Default project type must be 'library' iff project nature is N4JS
A very small `package.json` file that does not define the project type but still defines an n4js section shall default its project type to `library` instead of `plain js` as it is at the moment. See example below:

**package.json**
```
{
    "name" : "Test",
    "n4js" : {
        "sources": {
            "source": [
                "src"
            ]
        },
        "output": "src-gen"
    }
} 
```

mmews-n4:
closes #2530


mmews-n4:
build is green, merging

PR #2532, opened 2023-07-27
GH-2529: As a developer I want structural interfaces to become shapes

### LINKED ISSUE #2529: As a developer I want structural interfaces to become shapes
At the moment, N4JS provides classes, interfaces both as nominal and structural versions. This results in four type versions: nominal/structural classes or interfaces. All of these have in common that they provide reflection.

However, when classes or interfaces (structural or nominal) are used in n4jsd files to describe types from existing JavaScript, those do not provide reflection. (Side note: You can use the annotation \@N4JS to let the compiler know that reflection is provided by the described JavaScript code.) As a result, we have six different type version.

This userstory is about to remove one of those six versions: structural interfaces that have reflection. What will remain are nominal interfaces and so-called shapes known from TypeScript. Shapes only exist at compile time and have no manifestation at runtime. The annotation \@N4JS used in n4jsd files will become obsolete when used on structural interfaces.

mmews-n4:
closes #2529


mmews-n4:
tests green except for OPR test

PR #2563, opened 2023-11-01
GH-2561: As a smith I want to migrate the test framework from Xpect to Xt in bundle n4js.xpect.tests

### LINKED ISSUE #2561: As a smith I want to migrate the test framework from Xpect to Xt in bundle n4js.xpect.tests
In order to reduce the number of dependencies we decided to use an own, stripped-down implementation of Xpect as a test framework. That framework was introduced a while ago. This userstory is about to migrate all xt files that use Xpect to use Xt instead scoped to the bundle `n4js.xpect.tests`.

mmews-n4:
closes #2561


mmews-n4:
build is green, merging

PR #2578, opened 2023-11-21
GH-2576: TypeScript's meta type language has issues with circular type structures

### LINKED ISSUE #2576: TypeScript's meta type language has issues with circular type structures
In case a type T has a field also of type T (circular reference), TypeScripts meta type language features cause issues:

**Example of problematic TypeScript type**
```ts
export interface Clash {
    someMethod():void;
    fieldWithError: ExcludeMethods<Clash>[]; // 2 errors here, second is of interest
    // Error 0: 'fieldWithError' is referenced directly or indirectly in its own type annotation.
    // Error 1: Type of property 'fieldWithError' circularly references itself in mapped type '{ [K in keyof Clash]: Clash[K] extends Function ? never : K; }'
}

type ExcludeMethods<T> = Pick<T, { [K in keyof T]: T[K] extends Function ? never : K }[keyof T]>;
```

Since we use this definition when transpiling from n4js to d.ts, there can be errors in the resulting d.ts output. This needs to be addressed.

mmews-n4:
closes #2576


mmews-n4:
build is green, merging

PR #2580, opened 2023-11-24
GH-2579: As a smith I want to replace annotation NLS

### LINKED ISSUE #2579: As a smith I want to replace annotation NLS
This annotation is used to generate issue messages. Reasons:
- Its implementation is in xtend
- It is error-prone
- It has implications on the IDE (not usable outside Eclipse)
- It seems to be a case of over-engineering

mmews-n4:
closes #2579


mmews-n4:
build is green, merging.

PR #2591, opened 2024-01-10
GH-2588: Exported d.ts files contain non-public elements

### LINKED ISSUE #2588: Exported d.ts files contain non-public elements
**Test.n4js**
```js
function foo() {}
const bar = 5;
let baz = 3;
var qux = 6;
```

Results in
**Test.d.ts**
```ts
declare function foo(): void;
declare const bar: 5;
declare let baz: number;
declare var qux: number;
```

mmews-n4:
closes #2588


mmews-n4:
build is green, merging.

PR #2598, opened 2024-01-22
GH-2595: The d.ts export should append .js file extensions in import statements

### LINKED ISSUE #2595: The d.ts export should append .js file extensions in import statements
Double check if d.ts file extensions are valid as an alternative.

mmews-n4:
closes #2595


mmews-n4:
build is green, merging.

PR #2602, opened 2024-02-08
GH-2596: As a developer I need React jsx files using the new JSX transform

### LINKED ISSUE #2596: As a developer I need React jsx files using the new JSX transform
The n4jsx compiler compiles a redundant Object.assign() , e.g.
```jsx
<div foo="ff" />
```
into
```jsx
React.createElement('div', Object.assign({
	foo: "ff"
}));
```
Notes:
* Consider to either optimise Object.assign (remove it) iff it is the last argument
* Or consider to direct output destructuring patterns during transpilation (double check if that is consistent with other transpilations)

also see:
1) [here](https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html)
2) [and here](https://github.com/reactjs/rfcs/blob/createlement-rfc/text/0000-create-element-changes.md#motivation)

More notes:
* "existing namespace imports (i.e. `import * as React from "react"`) [...] is also a valid style" from 1)

mmews-n4:
closes #2596


mmews-n4:
build is green, merging

PR #2613, opened 2024-03-14
GH-2612: As a developer I want array literals to be typed as ArrayN<...> instead of just Array<union{...}>

### LINKED ISSUE #2612: As a developer I want array literals to be typed as ArrayN<...> instead of just Array<union{...}>
The following example shows a type cast to `ArrayN` which should be obsolete:
```js
let arr1 :Array2<number, string> = [42, 'hello'];
let [a, b] = arr1; // a typed as number, b typed as string
```

mmews-n4:
closes #2612


mmews-n4:
build is green, merging

PR #2621, opened 2024-04-22
GH-2614: N4JSX should accept functions that return ReactNodes

### LINKED ISSUE #2614: N4JSX should accept functions that return ReactNodes
```js

function Comp(): React.ReactNode {
    return null;
}
// XPECT noerrors -->
<Comp />
```

mmews-n4:
closes #2614


mmews-n4:
build is green, merging.

